PR #5, opened 2018-01-24
 Dev env using hab packages

elliott-davis:
Closes: #4442
tenor-4442921

Things missing in this PR: 
* Docker integration testing - changes need to be made to the worker
* Ability to run `cargo check` - this will require a dev mode in the studio
* Unstable packages by default - currently unstable is in a bad way with glibc

When this PR merges I plan on getting cards created for all of these.


BUILDER_DEV.md
```diff
@@ -1,225 +1,113 @@
...
+1. 80 - Web app
+1. 9631 - supervisor api port
+1. 5432 - posgres port
 
-* Install git
-```
-$ sudo apt-get install git
-```
+### Checkout
+1. `git clone https://github.com/habitat-sh/habitat.git`
```
baumanj: This should clone the builder repo now, right?

support/linux/provision.sh
```diff
@@ -1,43 +1,11 @@
...
-export GITHUB_WORKER_TEAM=2555389
-export GITHUB_WEBHOOK_SECRET=58d4afaf5e5617ab0f8c39e505605e78a054d003
-export GITHUB_APP_ID=5629
-pushd /src
-cp components/hab/install.sh /tmp/
-sh support/linux/install_dev_0_ubuntu_latest.sh
-sh support/linux/install_dev_9_linux.sh
+set -eux
+
+curl https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh | sudo bash
```
baumanj: Either `curl` needs to be a prerequisite (and can be removed from the `hab install` below, or the instructions need to run the installer script directly (either adding it to the builder repo, or referencing it from the habitat one).

support/linux/provision.sh
```diff
@@ -1,43 +1,11 @@
...
-export GITHUB_ADMIN_TEAM=1995301
-export GITHUB_WORKER_TEAM=2555389
-export GITHUB_WEBHOOK_SECRET=58d4afaf5e5617ab0f8c39e505605e78a054d003
-export GITHUB_APP_ID=5629
-pushd /src
-cp components/hab/install.sh /tmp/
-sh support/linux/install_dev_0_ubuntu_latest.sh
-sh support/linux/install_dev_9_linux.sh
+set -eux
+
```
baumanj: We can add this to make the `/src` link:
```
SRC_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd )"
sudo ln -s $(realpath $SRC_DIR/../..) /src 
```

support/linux/provision.sh
```diff
@@ -1,43 +1,11 @@
...
-sudo mkdir -p /hab/svc/builder-jobsrv/data
-sudo mkdir -p /hab/svc/builder-datastore
-sudo cp -f support/builder/datastore.toml /hab/svc/builder-datastore/user.toml
-support/builder/init-datastore.sh
-sudo -E support/builder/dev-config.sh
-support/linux/setup_keys.sh
-# Make sure setup_keys is the last thing so that if it has errors the user
-# can just rerun it without needing to redo the rest of this script
-popd
+echo 'eval "$(direnv hook bash)"' | sudo tee --append ~/.bashrc > /dev/null
```
baumanj: I think this should be
```
echo 'eval "$(direnv hook bash)"' >> ~/.bashrc
```
To support non-root users.

baumanj: I think it would also be good to add
```bash
echo "Please run `source ~/.bashrc; direnv allow` to complete direnv installation"
```

elliott-davis: I think both of those implementations support non-root users? They both point to the users home directory that is running the provision script

DEVELOPING.md
```diff
@@ -0,0 +1,137 @@
...
+### Ports required
+1. 9636 - Intra-supervisor communication
+1. 80 - Web app
+1. 9631 - supervisor api port
+1. 5432 - posgres port
+
+### Checkout
+1. If you are developing on Linux
+1. 1. Ensure you have curl
+1. 1. `git clone https://github.com/habitat-sh/builder.git /src/builder`
```
baumanj: I'm guessing you want:
```
* If you are developing on Linux
  * Ensure you have curl
  * `git clone https://github.com/habitat-sh/builder.git /src/builder`
* If you are developing on a Mac, `git clone https://github.com/habitat-sh/builder.git`
```
which renders as:
* If you are developing on Linux
  * Ensure you have curl
  * `git clone https://github.com/habitat-sh/builder.git /src/builder`
* If you are developing on a Mac, `git clone https://github.com/habitat-sh/builder.git`

Since I don't seem to be able to get numbered sublists to work right

DEVELOPING.md
```diff
@@ -0,0 +1,137 @@
...
+1. If you are developing on a Mac, `git clone https://github.com/habitat-sh/builder.git`
+
+If you are using Linux environment you can run `/src/support/linux/provision.sh` to configure your host
+If you are on a Mac, you will need brew, direnv, habitat, and Docker for Mac
+
+### GitHub OAuth Application
+`APP_HOSTNAME` mentioned below, will typically be `localhost`
```
baumanj: Do we support this being anything _other_ than `localhost`? Based on our scripts, we don't so, maybe we shouldn't imply there's any flexibility here and say it _must_ be `localhost`. Or, we could figure out the right way to support other values.

elliott-davis: This can also be an IP if you happen to be running this on a local network or EC2 VM. Which scripts are you referring to that are hard-coded to something?

DEVELOPING.md
```diff
@@ -0,0 +1,137 @@
...
+1. Create `${HABITAT_SRC_ROOT}/.secrets/habitat-env`
+1. Add
+```
+  APP_HOSTNAME=localhost
+  GITHUB_API_URL=https://api.github.com
+  GITHUB_WEB_URL=https://github.com
+  GITHUB_CLIENT_ID=${GITHUB_CLIENT_ID}
+  GITHUB_CLIENT_SECRET=${GITHUB_CLIENT_SECRET}
+  GITHUB_APP_ID=${GITHUP_APP_ID}
+  GITHUB_APP_NAME="my-dev-app"
```
baumanj: This needs to specifically match the name of the app downcased and with spaces replaced by hyphens. This makes it look like the literal value `"my-dev-app"` will work.

DEVELOPING.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Studio requirements
+
+1. Ensure you have run `hab setup` in your environment that will be executing the studio
+
+### Starting the services
+
+From either your VM or Mac:
+
+* `cd <habitat source path>`
```
baumanj: `builder source path` or maybe just `source path`?

DEVELOPING.md
cnunciato: Actually, this is the GitHub app "slug", which you can (and probably should) get from the Public Link field of an app's settings page. 

If we want to be super-clear, we should probably add a comment explaining what this is and where the value comes from, and consider changing the variable to use the full URL of that link (in a follow-up PR of course), since I believe that's all this is used for.

DEVELOPING.md
```diff
@@ -0,0 +1,138 @@
...
+  GITHUB_CLIENT_ID=${GITHUB_CLIENT_ID}
+  GITHUB_CLIENT_SECRET=${GITHUB_CLIENT_SECRET}
+  GITHUB_APP_ID=${GITHUP_APP_ID}
+  GITHUB_APP_NAME="${GITHUB_PUBLIC_LINK}"
+```
+1. Save and close
+
+### Studio requirements
+
+1. Ensure you have run `hab setup` in your environment that will be executing the studio
```
baumanj: I wonder if it would make sense to move this command to `provision.sh`. It has to run before `.envrc` gets sourced or else `HAB_AUTH_TOKEN` won't be set

elliott-davis: `hab setup` is interactive so I'd say probably not. Also, `hab setup` only needs to be run once. Adding it to the provision script would get pretty annoying for users (like myself) that provision several times a day.

DEVELOPING.md
```diff
@@ -0,0 +1,138 @@
...
+* `build-<component>` - Ex: `build-router` will build the router component for development and load it
+* `dev_docs` - Print the full set of command docs
+
+### Create a project using the web UI
+
+1. Go the web UI that you used in the last step
+2. Go to the origins page, and select your origin
+3. Click on the 'Connect a plan file' button
+4. Click on 'Install Github App' button to install the Builder Dev app on your github account
+5. Go back to the Packages page (from Step 3), and follow the instructions to link the plan you want to build
```
baumanj: Note: your GH app must have access to the repo containing the plan file. Forking `habitat-sh/core-plans` is an easy way to test.

DEVELOPING.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Run a build
+`build-builder`
+
+### Install dependencies in your local Builder env
+
+You may use the `load_packages` helper to specify a pacakge to upload. Ex:
```
baumanj: pacakge

DEVELOPING.md
baumanj: [`origin`](https://github.com/habitat-sh/builder/pull/5/files#diff-240ab41eb291d125c5844f2ba10aa78fR294), [`project`](https://github.com/habitat-sh/builder/pull/5/files#diff-240ab41eb291d125c5844f2ba10aa78fR312), possibly others.

support/linux/provision.sh
baumanj: Hmm, that does seem correct, though it didn't seem to happen that way when I ran it before. In any case:
```
echo 'eval "$(direnv hook bash)"' >> ~/.bashrc
```
Seems simpler  to me and should work for whatever user is running the script, right? I think the `sudo tee` construct is only needed when you want the output of a `sudo` command to write to a file (possibly appending) with `sudo` privileges.

DEVELOPING.md
elliott-davis: 👍

DEVELOPING.md
elliott-davis: Done. See `.envrc`

DEVELOPING.md
elliott-davis: ah, those will always be running locally but they are all localized to the studio. The times that we would need to care about external urls are for mostly the web stuff, which should be all encapsulated in `APP_HOSTNAME`

baumanj:
I think this can go ahead and merge. At this point any issues can be dealt with as follow-on work, I'd say

PR #7, opened 2018-01-24
Add version bump to git rev count since it's off after the builder migration

elliott-davis:
tenor-77877388


christophermaier:
tenor-214985538

PR #104, opened 2018-01-25
Update link for github app button in dev env

elliott-davis:

### DIFF
--- a/support/builder/config.sh
+++ b/support/builder/config.sh
@@ -24,7 +24,7 @@ mkdir -p /hab/svc/builder-api-proxy
 cat <<EOT > /hab/svc/builder-api-proxy/user.toml
 log_level = "debug"
 
-app_url = "http://localhost:9636"
+app_url = "http://${APP_HOSTNAME}:9636"
 
 [github]
 url = "$GITHUB_API_URL"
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

support/builder/config.sh
```diff
@@ -42,7 +32,7 @@ web_url = "$GITHUB_WEB_URL"
 client_id = "$GITHUB_CLIENT_ID"
 client_secret = "$GITHUB_CLIENT_SECRET"
 app_id = $GITHUB_APP_ID
-app_url = "${GITHUB_WEB_URL}/apps/$(echo ${GITHUB_APP_NAME} | tr "[:upper:]" "[:lower:]" | tr " " "-")"
+app_url = "${GITHUB_APP_NAME}"
```
cnunciato: I'm not sure this'll work, since the UI wants the full URL and not just the name. If I'm reading this right, we should should probably rename this to `GITHUB_APP_URL` and then update the sample config to use the URL to the dev-studio app as well. Does that sound right?

elliott-davis: Yeah that makes sense. I updated the value of the variable logically but not semantically. Got it patched up.

baumanj:
Looks good to me

PR #138, opened 2018-02-02
Hab service logging to Sumologic

chefsalim:
This change adds logging of the service output to Sumologic. It sets up a new systemd service that forwards journalctl logs to a Sumologic syslog collector, which then forwards it off to the Sumologic service.

tenor-1773504

### DIFF
--- a/terraform/files/sumocollector.service
+++ b/terraform/files/sumocollector.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=Send Habitat service logs to Sumologic syslog collector
+
+[Service]
+TimeoutStartSec=0
+ExecStart=/bin/sh -c '/bin/journalctl -fu hab-sup | /bin/ncat --udp localhost 1514'
+Restart=always
+RestartSec=5s
+[Install]
+WantedBy=multi-user.target
--- a/terraform/templates/sumo_sources_syslog.json
+++ b/terraform/templates/sumo_sources_syslog.json
@@ -1,0 +1,15 @@
+{
+  "api.version":"v1",
+  "source":{
+    "name":"${name}",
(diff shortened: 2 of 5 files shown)


thesentinels:

Thank you for contributing!

terraform/files/sumocollector.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Send Habitat service logs to Sumologic syslog collector
+
+[Service]
+TimeoutStartSec=0
+ExecStart=/bin/sh -c '/bin/journalctl -fu hab-sup | /bin/ncat --udp localhost 1514'
```
fnichol: It might be worth making this unit dependent on the `hab-sup.service` unit--that way this won't get started earlier. Then again, `journalctl` will display nothing until the service shows up (if ever).

terraform/templates/sumo_sources_syslog.json
```diff
@@ -0,0 +1,15 @@
+{
+  "api.version":"v1",
+  "source":{
+    "name":"${name}",
+    "category":"${category}",
+    "automaticDateParsing":true,
+    "forceTimeZone":false,
+    "timeZone":"Etc/UTC",
```
fnichol: The one-and-only Server Time :)

terraform/files/sumocollector.service
chefsalim: Good point - unfortunately the dependency options are not _quite_ what we need. The closest one would be adding a 'Before=' flag, however that would have to be added to the hab-sup.service instead of the sumocollector.service.  If we run into startup issues, I'll look at doing that.

PR #143, opened 2018-02-05
Introduce config flag to disable jobsrv communication

### LINKED ISSUE #137: [on-prem depot] Disable jobsrv
There are a number of places inside `builder-depot`, `builder-api` and `builder-originsrv` where we send messages to `builder-jobsrv`.  There needs to be a global flag that can be set in each service's configuration that will disable communication with `builder-jobsrv` so that the on-prem depot can be shipped without `builder-jobsrv` and still function.

raskchanky:
This PR introduces a new config flag for both `builder-depot` and `builder-api` called `jobsrv_enabled`.  It's `true` by default, but when it's set to `false`, then no communication with `builder-jobsrv` will occur.

I switched our routes from the macro syntax to the normal function based syntax because it turns out it's difficult to have programmatic control over your routes inside a macro.  I should also note that some of our routes are ambiguous enough (and iron's router isn't capable enough) that some of our routes will get called incorrectly.  One concrete example of this is:

`/pkgs/origins/:origin/stats`

This retrieves package stats and it uses `builder-jobsrv` to do it.  When `jobsrv_enabled` is set to `false`, that route isn't present in the routing table.  However, if you go to, e.g.:

`/pkgs/origins/core/stats`

instead of getting a 404, like you should, you'll get an empty package list.  That's because this route still matches:

`/pkgs/:origin/:pkg/:version`

which calls `list_packages`.  This kind of nonsense will go away once we have the opportunity to write v2 of the API, but for now, this is how things are.

tenor-107814571

Closes #137 

### DIFF
--- a/components/builder-api/src/config.rs
+++ b/components/builder-api/src/config.rs
@@ -43,6 +43,8 @@ pub struct Config {
     pub non_core_builds_enabled: bool,
     /// Where to record log events for funnel metrics
     pub log_dir: String,
+    /// Whether jobsrv is present or not
+    pub jobsrv_enabled: bool,
 }
 
 impl Default for Config {
@@ -57,6 +59,7 @@ impl Default for Config {
             events_enabled: false,
             non_core_builds_enabled: true,
             log_dir: env::temp_dir().to_string_lossy().into_owned(),
+            jobsrv_enabled: true,
         }
     }
 }
@@ -133,6 +136,7 @@ mod tests {
         let content = r#"
         events_enabled = true
         non_core_builds_enabled = true
+        jobsrv_enabled = false
 
         [http]
         listen = "0:0:0:0:0:0:0:1"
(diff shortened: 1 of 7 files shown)


thesentinels:

Thank you for contributing!

chefsalim:
@raskchanky unfortunately I think the scope of this change is actually larger than it needs to be, and we should avoid changing the routing from macros to implicit routing - it's more cumbersome to read and then there is the problem you mention above with incorrect routing currently.  I think we can satisfy the need requirements by only modifying the mainline paths that we know will get called in the on-prem scenario - things like upload package.  Other paths like schedule will not get called (or if they do it's fine to error out with a time out like what will happen when jobsrv is not present).

raskchanky:
@chefsalim My comment about the incorrect routing was merely to highlight something that already exists.  My change to the routing syntax didn't create that issue, and it will continue to exist even if we go back to the macro syntax.

Perhaps we should discuss this more offline, but I feel pretty strongly that leaving broken routes in our on-prem product is a bad idea, as is relying on the macro routing syntax.  I don't think the function syntax is any harder to read, just different.  More importantly, it's flexible when we need it to be, like in this scenario.  The macro syntax isn't flexible at all.

Issue #130, opened 2018-01-30
[on-prem depot] Bootstrapping packages script


raskchanky:
There are two options here.

The easiest thing to do is just to write a simple bash script that can pull a specific set of packages from the public Habitat depot and upload those artifacts to the on-prem depot.  This should be considered a must-have.  The specific set of packages that are needed is still TBD, but this script can be written and tested using any random set of packages for now, and refined after we have the final list.

Time permitting, a better long term solution is to have the on-prem depot automatically try to download packages from the public Habitat depot if it can't find them in its own local storage.  So the flow from the server's perspective looks like:
* Client requests to download a package
* Try to find said package in our local archive.
  * Success?  Send it to the client.  Done.
  * Not found?  Try to download said package from the public depot.
    * Success?  Send it to the client.  Also store it in our local archive as though the user had just uploaded it.  Done.
    * Not found?  404.  Done.

chefsalim:
Updated title, and will open a separate issue for the longer-term cache-thru approach.

raskchanky:
FWIW, I don't think the pull-through approach would be that much work.  That was the main reason I picked up this issue 😁

chefsalim:
We might need both - a script to populate the core packages initially, and then something to pull from the public depot, as initially there will be no package metadata at all...?

raskchanky:
Which core packages are we wanting to populate an on-prem depot with initially?

/cc @elliott-davis

chefsalim:
I'd think we just take them all. We could probably use a combination of the bootstrap script + the bulk promote tool to retrofit a simple script that takes all the latest core packages and puts them up on S3, and then they can easily be downloaded and installed on-prem from that bootstrap.

raskchanky:
Ah, so to clarify the kind of script(s) we're writing here:

There will be one script that is intended to be run on the builder-api production server, to find every hart file and tar them all up into a giant tarball (and it will be giant - the acceptance tarball @elliott-davis made a few months ago was on the order of 50GB), and from there, upload that tarball to S3.

Then there will be a separate script, which will connect to S3 and download the giant tarball, un-tar it, and manually upload all of the resulting hart files to the newly installed on-prem depot.

Do I have that correct?

raskchanky:
[...]

Issue #150, opened 2018-02-06
[on-prem depot] Install packages from an upstream depot


chefsalim:
This card is to add functionality to support the capability to use the public depot as an optional cache for retrieving package metadata and artifacts.  This can be implemented as a read-through (and optionally refresh-ahead) cache, allowing artifacts that are not present to be automatically retrieved when requested, and updated on a periodic time interval. Cache read-through may (should?) include channel information as well.  The calls to query 'latest' versions of a package should also be accounted for.

PR #180, opened 2018-02-09
Add a script to bootstrap an on-prem depot

raskchanky:
There are two scripts in this PR:

`on-prem-archive.sh` was my initial attempt at making this work by iterating over all of the plans in the `core-plans` repo, downloading the latest stable release of each one, tarring all of those up and uploading it to S3. It also supported a mode of operation to download the tarball from S3, crack it open, and upload all the resulting hart files to a fresh depot.

That strategy sounded fine at first glance but turned out to be fraught with peril. It turns out that many packages had deps that were older than the latest stable release, so when uploading the transitive deps of some packages, things failed because they couldn't find the right files.

I've left this script in this PR because eventually we're going to want to support air-gapped environments, and this script is a good way to do that, but it's missing the equivalent of [hab pkg download](https://github.com/habitat-sh/habitat/issues/4616).

There's also `on-prem-install.sh`, which is my second attempt at making this work.  This _does not_ support air-gapped installs, but instead just downloads all of the core-plans from stable, and immediately re-uploads them all to the depot specified.  Note that this is an error prone process and what we have here is not a perfect solution, but a best-effort one.  In practice, there are a number of failures that can happen in this automated process that a user wouldn't run into if they were installing packages by hand.  In my opinion, the long term solution to this is captured [in this issue](https://github.com/habitat-sh/builder/issues/150).

tenor-167905666

Closes https://github.com/habitat-sh/builder/issues/130

### DIFF
--- a/support/on-prem-archive.sh
+++ b/support/on-prem-archive.sh
@@ -1,0 +1,180 @@
+#!/bin/bash
+
+# The purpose of this script is to download the latest stable version of every
+# package in the core-plans repo, tar them up, and upload to S3. It also supports
+# downloading the archive from S3, extracting it, and uploading to a new depot.
+#
+# There are some environment variables you can set to control the behavior of this
+# script:
+#
+# HAB_ON_PREM_BOOTSTRAP_BUCKET_NAME: This controls the name of the S3 bucket where
+# the archive is placed. The default is habitat-on-prem-builder-bootstrap
+#
+# HAB_ON_PREM_BOOTSTRAP_S3_ROOT_URL: This controls the domain name for S3 where the
+# files will be downloaded from. The default is https://s3-us-west-2.amazonaws.com
+#
+# HAB_ON_PREM_BOOTSTRAP_DONT_CLEAN_UP: This controls whether the script cleans up
+# after itself by deleting the intermediate files that were created during its run.
+# Setting this variable to any value will cause the cleanup to be skipped. By
+# default, the script will clean up after itself.
+#
+# IT SHOULD BE NOTED THAT THIS SCRIPT DOES NOT WORK IN ITS CURRENT FORM. The key
+# problem is that some packages that are the latest stable are built against libs
+# that are older than the current stable version of those libs, so when you go to
+# upload the resulting packages from the S3 archive, uploads fail because when
+# trying to upload the transitive deps, the wrong versions are present (it's looking
+# for older versions than what have been downloaded).
+
+set -eo pipefail
+
+help() {
+    echo "Usage: on-prem-archive.sh {create-archive|populate-depot <DEPOT_URL>}"
+}
+
+exists() {
+  if command -v $1 >/dev/null 2>&1
+  then
+    return 0
+  else
+    return 1
+  fi
+}
+
+s3_cp() {
+  aws s3 cp --acl=public-read ${1} ${2} >&2
+}
+
+check_tools() {
+  local ref=$1[@]
+
+  for tool in ${!ref}
+  do
+    if ! exists "$tool"; then
+      echo "Please install $tool and run this script again."
+      exit 1
+    fi
+  done
+}
+
+check_vars() {
+  local ref=$1[@]
+
+  for var in ${!ref}
+  do
+    if [ -z "${!var}" ]; then
+      echo "Please ensure that $var is exported in your environment and run this script again."
+      exit 1
+    fi
+  done
+}
+
+bucket="${HAB_ON_PREM_BOOTSTRAP_BUCKET_NAME:-habitat-on-prem-builder-bootstrap}"
+marker="LATEST.tar.gz"
+
+case "$1" in
+  create-archive)
+    required_tools=( aws git curl jq )
+    required_vars=( AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY )
+
+    check_tools required_tools
+    check_vars required_vars
+
+    core_tmp=$(mktemp -d)
+    core="$core_tmp/core-plans"
+    upstream_depot="https://bldr.habitat.sh"
+    bootstrap_file="on-prem-bootstrap-$(date +%Y%m%d%H%M%S).tar.gz"
+    tar_file="/tmp/$bootstrap_file"
+    tmp_dir=$(mktemp -d)
+
+    git clone https://github.com/habitat-sh/core-plans.git "$core"
+    pushd "$core"
+    dir_list=$(find . -type f -name "plan.sh" -printf "%h\n" | sed -r "s|^\.\/||" | sort -u)
+    total=$(echo "$dir_list" | wc -l)
+    count="0"
+
+    for p in $dir_list
+    do
+      count=$((count+1))
+      echo ""
+      echo "[$count/$total] Resolving latest version of core/$p"
+      latest=$(curl -s -H "Accept: application/json" "$upstream_depot/v1/depot/channels/core/stable/pkgs/$p/latest")
+      raw_ident=$(echo "$latest" | jq ".ident")
+
+      if [ "$raw_ident" = "null" ]; then
+        echo "Failed to find a latest version. Skipping."
+        continue
+      fi
+
+      slash_ident=$(echo "$raw_ident" | jq "\"\(.origin)/\(.name)/\(.version)/\(.release)\"" | tr -d '"')
+      dash_ident=$(echo "$raw_ident" | jq "\"\(.origin)-\(.name)-\(.version)-\(.release)\"" | tr -d '"')
+      target=$(echo "$latest" | jq ".target" | tr -d '"')
+
+      echo "[$count/$total] Downloading $slash_ident"
+      curl -s -H "Accept: application/json" -o $tmp_dir/$dash_ident-$target.hart "$upstream_depot/v1/depot/pkgs/$slash_ident/download"
+    done
+    popd
+
+    cd /tmp
+    tar zcvf $tar_file -C $tmp_dir .
+    s3_cp $tar_file s3://$bucket/
+    s3_cp s3://$bucket/$bootstrap_file s3://$bucket/$marker
+
+    if [ -z "$HAB_ON_PREM_BOOTSTRAP_DONT_CLEAN_UP" ]; then
+      echo "Cleaning up."
+      rm -fr "$tmp_dir"
+      rm -fr "$core_tmp"
+      rm "$tar_file"
+    else
+      echo "Cleanup skipped."
+    fi
+
+    echo "Done."
+    ;;
+  populate-depot)
+    if [ -z "$2" ]; then
+      help
+      exit 1
+    fi
+
+    required_tools=( curl )
+    required_vars=( HAB_AUTH_TOKEN )
+
+    check_tools required_tools
+    check_vars required_vars
+
+    s3_root_url="${HAB_ON_PREM_BOOTSTRAP_S3_ROOT_URL:-https://s3-us-west-2.amazonaws.com}/$bucket"
+    tmp_dir=$(mktemp -d)
+
+    cd "$tmp_dir"
+    echo "Fetching latest package bootstrap file."
+    curl -O "$s3_root_url/$marker"
+    tar zxvf $marker
+
+    harts=$(find . -type f -name "*.hart")
+    total=$(echo "$harts" | wc -l)
+    count="0"
+
+    for hart in $harts
+    do
+      count=$((count+1))
+      echo ""
+      echo "[$count/$total] Uploading $hart to the depot at $2"
+      hab pkg upload --url $2 --channel stable $hart
+    done
+
+    echo "Package uploads finished."
+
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

elliott-davis:
<bikeshed> The name of this is a bit confusing to me. I assumed it was a script to start start the depot. I think something like `depot_archive_tools.sh` might make more sense </bikeshed>

support/on-prem-bootstrap.sh
```diff
@@ -0,0 +1,103 @@
+#!/bin/bash
+
+# The purpose of this script is to download the latest stable version of every
+# package in the core-plans repo, tar them up, and upload to S3. It also supports
+# downloading the archive from S3, extracting it, and uploading to a new depot.
+
+set -e
```
chefsalim: Use 'set -euo pipefail' instead?

support/on-prem-bootstrap.sh
```diff
@@ -0,0 +1,103 @@
...
+  else
+    return 1
+  fi
+}
+
+required_tools=( aws git curl jq )
+required_vars=( AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY HAB_AUTH_TOKEN )
+upstream_depot="https://bldr.habitat.sh"
+bucket="habitat-on-prem-builder-bootstrap"
+bootstrap_file="on-prem-bootstrap.tar.gz"
```
chefsalim: Might be good to append a timestamp so we can have different versions of the package bundles...

support/on-prem-bootstrap.sh
```diff
@@ -0,0 +1,103 @@
...
+    echo "Please ensure that $var is exported in your environment and run this script again."
+    exit 1
+  fi
+done
+
+case "$1" in
+  create-archive)
+    core="/tmp/core-plans"
+    tar_file="/tmp/$bootstrap_file"
+    name=$(cat /dev/urandom | tr -cd 'a-f0-9' | head -c 32)
```
chefsalim: mktemp might be a simpler alternative?

support/on-prem-bootstrap.sh
```diff
@@ -0,0 +1,103 @@
...
+do
+  if [ -z "${!var}" ]; then
+    echo "Please ensure that $var is exported in your environment and run this script again."
+    exit 1
+  fi
+done
+
+case "$1" in
+  create-archive)
+    core="/tmp/core-plans"
```
chefsalim: Really paranoid, but might be worth creating a random temp name instead of using 'core-plans' for the location to clone

support/on-prem-bootstrap.sh
raskchanky: I can add `-o pipefail` but I specifically don't want `-u` because that will cause the graceful error handling below on line 38 to fail.  I'd rather explicitly check for the env vars we expect to be set and gracefully fail, as that's a better UX.

support/on-prem-bootstrap.sh
raskchanky: Sure, I can do that, though it complicates the download process, as then we need to fetch everything from S3 and iterate through all of the files in order to find the correct one to download.  As the number of files in S3 grows larger, that's going to cause a bigger and bigger delay downloading files.

support/on-prem-bootstrap.sh
raskchanky: Sure.

support/on-prem-bootstrap.sh
raskchanky: Sure.

support/on-prem-bootstrap.sh
chefsalim: We could do it exactly the how the hab bootstrap bundle script does it, with a LATEST pointer.. copy/pastable code from the other script... a bit more work but would provide some more flexibility.

```
manifest_url=${s3_root_url}/LATEST
log "Downloading latest builder tarball manifest from ${manifest_url}"
${curl} --remote-name ${manifest_url} >&2
latest_archive=$(${awk} 'NR==1' LATEST)
```

support/on-prem-bootstrap.sh
chefsalim: Cool, sounds good

support/on-prem-bootstrap.sh
raskchanky: Ah, nice.  Will do.

support/on-prem-bootstrap.sh
```diff
@@ -0,0 +1,103 @@
...
+    if [ -z "$2" ]; then
+      help
+      exit 1
+    fi
+
+    cache_dir="/hab/cache/artifacts"
+
+    mkdir -p "$cache_dir"
+    cd "$cache_dir"
+    aws s3 cp s3://$bucket/$bootstrap_file .
```
chefsalim: We should not depend on aws tools for the download as that is one more tool for the customer to set up...

Instead can use what the other bootstrap is doing (also good to have a status message output as this is likely a long operation)

```
latest_package_url=${s3_root_url}/${latest_archive}
log "Downloading ${latest_archive} from ${latest_package_url}"
${curl} --remote-name ${s3_root_url}/${latest_archive} >&2
```

support/on-prem-bootstrap.sh
```diff
@@ -0,0 +1,103 @@
...
+  fi
+}
+
+required_tools=( aws git curl jq )
+required_vars=( AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY HAB_AUTH_TOKEN )
+upstream_depot="https://bldr.habitat.sh"
+bucket="habitat-on-prem-builder-bootstrap"
+bootstrap_file="on-prem-bootstrap.tar.gz"
+
+for tool in "${required_tools[@]}"
```
chefsalim: We should only need to depend on curl for populate_depot, so this could be two checks...

support/on-prem-bootstrap.sh
```diff
@@ -0,0 +1,103 @@
...
+    rm "$tar_file"
+    echo "Done."
+    ;;
+  populate-depot)
+    if [ -z "$2" ]; then
+      help
+      exit 1
+    fi
+
+    cache_dir="/hab/cache/artifacts"
```
chefsalim: Is this correct?  I could be wrong, but I _believe_ when uploading a hart, it looks for the deps in the same directory as the hart.

If we need to take over the user's artifact cache, we should probably warn and prompt them..

raskchanky: It didn't seem to work in my tests, but it's hard to do a "real" test of this because doing so would create a gigantic tarball.  Maybe I'll try it out on an EC2 instance.

raskchanky:
@elliott-davis renamed

raskchanky:
@chefsalim All feedback addressed

support/on-prem-archive.sh
```diff
@@ -0,0 +1,122 @@
...
+    ;;
+  populate-depot)
+    if [ -z "$2" ]; then
+      help
+      exit 1
+    fi
+
+    required_tools=( curl )
+    required_vars=( HAB_AUTH_TOKEN )
+    s3_root_url="https://s3-us-west-2.amazonaws.com/$bucket"
```
cnunciato: I happen to be running this with a personal us-east key, so making this URL configurable as well might be nice.

support/on-prem-archive.sh
```diff
@@ -0,0 +1,122 @@
...
+  for var in "$1"
+  do
+    if [ -z "${!var}" ]; then
+      echo "Please ensure that $var is exported in your environment and run this script again."
+      exit 1
+    fi
+  done
+}
+
+bucket="habitat-on-prem-builder-bootstrap"
```
cnunciato: I think top-level bucket names have to be unique, either globally or regionally. Should this be configurable?

support/on-prem-archive.sh
```diff
@@ -0,0 +1,122 @@
...
+
+      curl -H "Accept: application/json" -o $tmp_dir/$dash_ident-$target.hart "$upstream_depot/v1/depot/pkgs/$slash_ident/download"
+    done
+    popd
+
+    cd /tmp
+    tar zcvf $tar_file -C $tmp_dir .
+    s3_cp $tar_file s3://$bucket/
+    s3_cp s3://$bucket/$bootstrap_file s3://$bucket/$marker
+    rm -fr "$tmp_dir"
```
cnunciato: Since we'll have waited a long time for all of these to download, should we allow for keeping these around? (I think the thing I'm running right now is going to fail because of a name conflict, so if it does, it'd be good not to have to download everything again.)

support/on-prem-archive.sh
```diff
@@ -0,0 +1,122 @@
...
+
+      if [ "$raw_ident" = "null" ]; then
+        continue
+      fi
+
+      slash_ident=$(echo "$raw_ident" | jq "\"\(.origin)/\(.name)/\(.version)/\(.release)\"" | tr -d '"')
+      dash_ident=$(echo "$raw_ident" | jq "\"\(.origin)-\(.name)-\(.version)-\(.release)\"" | tr -d '"')
+      target=$(echo "$latest" | jq ".target" | tr -d '"')
+
+      curl -H "Accept: application/json" -o $tmp_dir/$dash_ident-$target.hart "$upstream_depot/v1/depot/pkgs/$slash_ident/download"
```
cnunciato: It might be helpful to echo the name and ordinal position in the list, just to give some sense of what's been done and what remains.

support/on-prem-archive.sh
cnunciato: Yeah, I just ran this and got ```upload failed: ../../tmp/on-prem-bootstrap-20180212121935.tar.gz to s3://habitat-on-prem-builder-bootstrap/on-prem-bootstrap-20180212121935.tar.gz An error occurred (AccessDenied) when calling the CreateMultipartUpload operation: Access Denied``` I had commented the removals, though, so luckily I still have the 4 GB tar file.

support/on-prem-archive.sh
cnunciato: Yeah, when I attempt to upload, I get an error, and my S3 client tells me the name's already in use:

image

support/on-prem-archive.sh
raskchanky: Sure.

support/on-prem-archive.sh
```diff
@@ -0,0 +1,122 @@
...
+    check_tools ${required_tools[@]}
+    check_vars ${required_vars[@]}
+
+    cd "$tmp_dir"
+    curl -O "$s3_root_url/$marker"
+    tar zxvf $marker
+
+    for hart in *.hart
+    do
+      hab pkg upload --url $2 --channel stable $hart
```
cnunciato: I do get some failures when I attempt to upload the set of packages created by the `create-archive` step.

Several complaints about these two specifically so far:

```
∵ Missing artifact for core-gmp-6.1.0-20170513202112-x86_64-linux.hart was not found
∵ Missing artifact for core-cacerts-2017.01.18-20170513215104-x86_64-linux.hart was not found
```

raskchanky: I had a suspicion that this might happen, and I bet the reason for it is the way we're preparing this tarball.  We get a list of everything in core-plans, and grab the latest stable version of all of those packages.  But, if one of those packages happens to have a dependency that's not the latest stable version for that particular package, then it won't get downloaded and put into the tarball.

cnunciato: Yep. I'm guessing this'll get worse when I try to run a build (which is probably outside the scope of this PR anyway), but fingers crossed, we'll see!

support/on-prem-archive.sh
raskchanky: Sure

support/on-prem-archive.sh
raskchanky: Sure

raskchanky: `HAB_ON_PREM_BOOTSTRAP_S3_ROOT_URL` should configure this now.

support/on-prem-archive.sh
raskchanky: `HAB_ON_PREM_BOOTSTRAP_BUCKET_NAME` will configure this now.

support/on-prem-archive.sh
raskchanky: `HAB_ON_PREM_BOOTSTRAP_DONT_CLEAN_UP` will skip the removal of things now.

raskchanky:
@elliott-davis @chefsalim @cnunciato I've made quite a few changes since y'all last looked at this, including an entirely new PR description.  Give it a fresh look if you get a chance.  My current tests in EC2 look good.

chefsalim:
The new script looks good to me... we should move/copy it over to the on-prem-depot repo, along with instructions on how to use it in the README.

raskchanky:
@chefsalim Done https://github.com/habitat-sh/on-prem-builder/pull/4

PR #184, opened 2018-02-09
added new proto defs for origin encryption keys

elliott-davis:

### DIFF
--- a/components/builder-protocol/protocols/originsrv.proto
+++ b/components/builder-protocol/protocols/originsrv.proto
@@ -413,6 +413,53 @@ message OriginProjectList {
   repeated string names = 1;
 }
 
+message  OriginPublicEncryptionKey {
+  optional uint64 id = 1;
+  optional uint64 origin_id = 2;
+  optional string name = 3;
+  optional string revision = 4;
+  optional bytes body = 5;
+  optional uint64 owner_id = 6;
+}
+
+message  OriginPublicEncryptionKeyCreate {
+  optional OriginPublicEncryptionKey public_encryption_key = 1;
+}
+
+message  OriginPublicEncryptionKeyGet {
+  optional uint64 owner_id = 1;
+  optional string origin = 2;
+  optional string revision = 3;
+}
+
+message  OriginPublicEncryptionKeyLatestGet {
+  optional uint64 owner_id = 1;
+  optional string origin = 2;
+}
+
+message OriginPublicEncryptionKeyListRequest {
+  optional uint64 owner_id = 1;
+  optional uint64 origin_id = 2;
+}
+
+message OriginPublicEncryptionKeyListResponse {
+  optional uint64 origin_id = 1;
+  repeated OriginPublicEncryptionKey keys = 2;
+}
+
+message OriginPrivateEncryptionKey {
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/builder-protocol/protocols/originsrv.proto
```diff
@@ -413,6 +413,61 @@ message OriginProjectList {
...
+  optional uint64 owner_id = 1;
+  optional uint64 origin_id = 2;
+}
+
+message OriginPublicEncryptionKeyListResponse {
+  optional uint64 origin_id = 1;
+  repeated OriginPublicEncryptionKey keys = 2;
+}
+
+message OriginEncryptionKey {
```
chefsalim: Name this OriginSecretEncryptionKey to be consistent with the OriginSecretKey terminology?

chefsalim:
Looks good, w/ just a comment on naming..

components/builder-protocol/protocols/originsrv.proto
elliott-davis: I left it this way to be consistent with the naming for `OriginSigningKey` but there is also an open PR for that so I'm happy to change both if you think it would be more clear

raskchanky: Since the other one is called `OriginPublicEncryptionKey`, it feels appropriate for this one to have a qualifier as well and remove any ambiguity around what kind of key it is.

components/builder-protocol/protocols/originsrv.proto
```diff
@@ -413,6 +413,61 @@ message OriginProjectList {
...
+  optional uint64 owner_id = 6;
+}
+
+message OriginEncryptionKeyCreate {
+  optional uint64 origin_id = 1;
+  optional string name = 2;
+  optional string revision = 3;
+  optional bytes body = 4;
+  optional uint64 owner_id = 5;
+}
```
raskchanky: This is not a huge deal, but I'm a fan of the pattern where message that create things hold a reference to the thing they're creating, rather than duplicating all the fields.  So in this instance, `OriginEncryptionKeyCreate` would have 1 field:  `optional OriginEncryptionKey encryption_key = 1;`.  The same applies to `OriginPublicEncryptionKey` above.  Like I said, not a deal breaker, but it feels better than having all the fields duplicated.

raskchanky:
A few minor comments, but on the whole, looks great.

components/builder-protocol/protocols/originsrv.proto
chefsalim: Yeah, it feels better to have Public/Secret as the secrecy qualifier, and Signing/Encryption as the functional qualifier for both these keys.  But it's more change.  If we're changing the origin key names, then we should probably do it.  Another option could be to just leave the origin key names alone, and only add the new encryption keys.  So many choices... :)

components/builder-protocol/protocols/originsrv.proto
chefsalim: 👍

elliott-davis:
@thesentinels approve

PR #193, opened 2018-02-12
Check for PEM file before starting worker

chefsalim:
Check for PEM file in the builder-worker run hook - if the PEM file does not exist, the worker can crash if a job is queued.

tenor-170557511

### DIFF
--- a/components/builder-worker/habitat/hooks/run
+++ b/components/builder-worker/habitat/hooks/run
@@ -6,4 +6,11 @@ export RUST_BACKTRACE=1
 export HAB_STUDIO_BACKLINE_PKG=core/hab-backline
 pkg_svc_run="bldr-worker start -c {{pkg.svc_config_path}}/config.toml"
 
+# Wait for pem file before starting the service
+while ! [ -f {{pkg.svc_files_path}}/builder-github-app.pem ];
+do
+    echo "Waiting for builder-github-app.pem"
+    sleep 30
+done
+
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/builder-worker/habitat/hooks/run
```diff
@@ -6,4 +6,11 @@ export RUST_BACKTRACE=1
 export HAB_STUDIO_BACKLINE_PKG=core/hab-backline
 pkg_svc_run="bldr-worker start -c {{pkg.svc_config_path}}/config.toml"
 
+# Wait for pem file before starting the service
+while ! [ -f {{pkg.svc_files_path}}/builder-github-app.pem ];
+do
+    echo "Waiting for builder-github-app.pem"
+    sleep 30
+done
```
raskchanky: Why is it preferable for the worker to sit in an infinite loop if the PEM file isn't present vs just aborting with an error message?

chefsalim: It's because the pem file will eventually gossip it's way through.. if we are spinning up a bunch of workers, we just need to wait for the file to show up rather than aborting the run.  In some ways, it's like waiting for a service bind... ideally we would not need to do this directly, but it seems like a sufficient workaround for now.

raskchanky: Gotcha

raskchanky:
tenor-170784832

PR #212, opened 2018-02-15
Use Builder error codes

cnunciato:
This change adds an `ErrorCode` enum that maps to [Builder’s error codes](https://github.com/habitat-sh/builder/blob/d6ea4585efa49f2f1a11e70d2869675cc7b9acbe/components/builder-protocol/protocols/net.proto#L12), and makes use of the `NotFound` error to handle responses for log files that don’t yet exist.

tenor-70723691

### DIFF
--- a/components/builder-web/app/client/builder-api.ts
+++ b/components/builder-web/app/client/builder-api.ts
@@ -20,6 +20,10 @@ import { AppStore } from '../app.store';
 import { addNotification, signOut } from '../actions/index';
 import { WARNING } from '../actions/notifications';
 
+
+export enum ErrorCode {
+  NotFound = 4
+}
 export class BuilderApiClient {
   private headers;
   private urlPrefix: string = `${config['habitat_api_url']}/v1`;
@@ -315,7 +319,7 @@ export class BuilderApiClient {
           if (response.ok) {
             resolve(response.json());
           } else {
-            reject(new Error(response.statusText));
+            response.json().then(reject);
           }
         })
         .catch(error => this.handleError(error, reject));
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/builder-web/app/client/builder-api.ts
```diff
@@ -20,6 +20,10 @@ import { AppStore } from '../app.store';
 import { addNotification, signOut } from '../actions/index';
 import { WARNING } from '../actions/notifications';
 
+
+export enum ErrorCode {
```
mgamini: Would it be worth having a separate error class for apis? Maybe not since there's only one.

cnunciato: Enum should probably be fine for now, yeah. I'll also make a card for generating these based on our actual [protocols](https://github.com/habitat-sh/builder/tree/master/components/builder-protocol/protocols) via [protobuf.js](https://github.com/dcodeIO/protobuf.js), which we did on another project recently.

components/builder-web/app/client/builder-api.ts
```diff
@@ -315,7 +319,9 @@ export class BuilderApiClient {
           if (response.ok) {
             resolve(response.json());
           } else {
-            reject(new Error(response.statusText));
+            response.json().then(error => {
```
mgamini: I think this can just be 
`response.json().then(reject);`

(haven't tested it though)

cnunciato: Yeah, that's a little cleaner, huh -- I'll try it.

PR #217, opened 2018-02-20
Fix non-existent-user invitation message

cnunciato:
Currently, when an invitation is extended to a user who hasn’t signed into Builder before, the error message we surface is simply `Unauthorized`, which is not helpful. This just changes that messages to make it clearer what needs to be done.

tenor-63173793

### DIFF
--- a/components/builder-web/app/client/builder-api.ts
+++ b/components/builder-web/app/client/builder-api.ts
@@ -15,11 +15,9 @@
 import 'whatwg-fetch';
 import config from '../config';
 import { parseKey } from '../util';
-import { GitHubApiClient } from './github-api';
 import { AppStore } from '../app.store';
 import { addNotification, signOut } from '../actions/index';
 import { WARNING } from '../actions/notifications';
-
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

raskchanky:
Is there a reason we don't want to check GH any more and verify that the person they entered is actually a valid GH user?  Is this a security measure?  I could see someone becoming fairly frustrated if they're trying to invite someone and unknowingly typoing the name repeatedly.

cnunciato:
I took out this check just because of the need to begin detaching from GitHub for auth, yeah.

raskchanky:
Ah, fair point.  Part of me is sad to lose that feature, as it feels like a really good UX.  I wonder if it's worth making a comment/note to revisit it in the future when we have a more solid strategy for generic auth and making it available where possible (e.g. GH and BB would likely have APIs to support this, but maybe not A2).  Either way, it shouldn't hold up merging this.

PR #218, opened 2018-02-21
Add ability to generate origin encryption keys via API endpoint

elliott-davis:


thesentinels:

Thank you for contributing!

components/builder-http-gateway/src/http/helpers.rs
```diff
@@ -429,6 +431,52 @@ pub fn get_optional_session_id(req: &mut Request) -> Option<u64> {
...
+    let mut private_key = OriginPrivateEncryptionKey::new();
+    public_key.set_owner_id(session.get_id());
+    private_key.set_owner_id(session.get_id());
+    public_key.set_name(origin.get_name().to_string());
+    public_key.set_origin_id(origin.get_id());
+    private_key.set_name(origin.get_name().to_string());
+    private_key.set_origin_id(origin.get_id());
+
+    let pair = BoxKeyPair::generate_pair_for_origin(origin.get_name())
+        .expect("failed to generate origin encryption key pair");
```
chefsalim: Handle failure instead of expect?

components/builder-http-gateway/src/http/helpers.rs
```diff
@@ -429,6 +431,52 @@ pub fn get_optional_session_id(req: &mut Request) -> Option<u64> {
...
+    public_key.set_revision(pair.rev.clone());
+    public_key.set_body(
+        pair.to_public_string()
+            .expect("no public key in generated encryption key pair")
```
chefsalim: Same q on the other expects..

components/builder-http-gateway/src/http/helpers.rs
```diff
@@ -429,6 +431,52 @@ pub fn get_optional_session_id(req: &mut Request) -> Option<u64> {
     }
 }
 
+pub fn generate_origin_encryption_keys(
```
chefsalim: Minor - the function might read better if it was split into two  smaller ones vs. essentially repeating every line twice?  Not a blocker tho.

components/builder-originsrv/src/data_store.rs
```diff
@@ -634,6 +634,156 @@ impl DataStore {
...
+            &[
+                &(opek.get_origin_id() as i64),
+                &(opek.get_owner_id() as i64),
+                &opek.get_name(),
+                &opek.get_revision(),
+                &format!("{}-{}", opek.get_name(), opek.get_revision()),
+                &opek.get_body(),
+            ],
+        ).map_err(SrvError::OriginPrivateEncryptionKeyCreate)?;
+        let row = rows.iter().nth(0).expect(
```
chefsalim: Check row len instead of expect?

components/builder-originsrv/src/data_store.rs
```diff
@@ -634,6 +634,156 @@ impl DataStore {
...
+            &[
+                &(opek.get_origin_id() as i64),
+                &(opek.get_owner_id() as i64),
+                &opek.get_name(),
+                &opek.get_revision(),
+                &format!("{}-{}", opek.get_name(), opek.get_revision()),
+                &opek.get_body(),
+            ],
+        ).map_err(SrvError::OriginPublicEncryptionKeyCreate)?;
+        let row = rows.iter().nth(0).expect(
```
chefsalim: check row len?

components/builder-depot/src/server.rs
```diff
@@ -457,6 +457,29 @@ fn write_archive(filename: &PathBuf, body: &mut Body) -> Result<PackageArchive>
     Ok(PackageArchive::new(filename))
 }
 
+fn generate_origin_encryption_keys(req: &mut Request) -> IronResult<Response> {
+    debug!("Generate Origin Encryption Keys {:?}", req);
+    let session = req.extensions.get::<Authenticated>().unwrap().clone();
```
raskchanky: If you're concerned about minimizing your allocations :trollface: you could eliminate this line completely and pull the session out of the `req` struct in your `generate_origin_encryption_keys` helper function.  Doing it there would work without cloning it.

components/builder-originsrv/src/migrations/2018-02-21-142803_add_origin_encryption_keys/up.sql
```diff
@@ -0,0 +1,111 @@
...
+
+CREATE OR REPLACE FUNCTION insert_origin_public_encryption_key_v1 (
+  opek_origin_id bigint,
+  opek_owner_id bigint,
+  opek_name text,
+  opek_revision text,
+  opek_full_name text,
+  opek_body bytea
+) RETURNS SETOF origin_public_encryption_keys AS $$
+    BEGIN
```
chefsalim: There was a discussion a while ago that we should try to use SQL instead of PSQL as the language wherever possible.. I don't think it's a blocker but just FYI...

chefsalim:
Looks good, with a few comments.

raskchanky:
One minor comment but otherwise looks good to me.

components/builder-http-gateway/src/http/helpers.rs
raskchanky: In fact, looking through the code that's already in this crate, there are `expect` and `unwrap` calls everywhere, so this at least seems consistent to me.  https://github.com/habitat-sh/builder/issues/219 has been created to track this.

components/builder-originsrv/src/data_store.rs
elliott-davis: I think the option here is leave it as is or do something like:
```
if rows.len() != 0 {
    let row = rows.iter().nth(0).unwrap();
     Ok(self.row_to_origin_private_encryption_key(row))
} else {
    Err("some error")
}```

Does that sound right?

elliott-davis:
@thesentinels approve

PR #220, opened 2018-02-21
Add api endpoint to fetch latest origin public encryption key

elliott-davis:

### DIFF
--- a/components/builder-depot/src/server.rs
+++ b/components/builder-depot/src/server.rs
@@ -88,8 +88,7 @@ pub fn origin_update(req: &mut Request) -> IronResult<Response> {
     match req.get::<bodyparser::Struct<OriginUpdateReq>>() {
         Ok(Some(body)) => {
             let dpv = match body.default_package_visibility
-                .parse::<OriginPackageVisibility>()
-            {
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/builder-depot/src/server.rs
```diff
@@ -455,6 +457,36 @@ fn write_archive(filename: &PathBuf, body: &mut Body) -> Result<PackageArchive>
...
+    }
+    let key = match route_message::<
+        OriginPublicEncryptionKeyLatestGet,
+        OriginPublicEncryptionKey,
+    >(req, &request) {
+        Ok(key) => key,
+        Err(err) => return Ok(render_net_error(&err)),
+    };
+
+    let xfilename = format!("{}-{}.pub", key.get_name(), key.get_revision());
```
mwrock: might not be a bad idea extracting a `public_key_file_name` function since we are using this format in a few places in this file.

elliott-davis: I tried putting this logic in `download_content_as_file` but there is `sig.key` and `pub` types which added more checking logic than it removed

PR #221, opened 2018-02-22
Personal Access Tokens API

chefsalim:
This change adds the Personal Access Token API and middleware authentication functionality. This means we can now start transitioning away from using the github personal access tokens, and use our own tokens instead. The new tokens are drop-in replacable for use with the hab cli, so they are usable with existing hab clients (no need to upgrade the client!). 

A couple of caveats:
1. This functionality is still behind a feature flag in the UI, and will be gradually phased in to prod starting with the acceptance environment.
2. There is still some work to be done to handle token revocation, that will follow soon.

tenor-252921655


components/builder-sessionsrv/src/migrations/2018-02-15-232946_personal_access_tokens/up.sql
```diff
@@ -0,0 +1,33 @@
+CREATE SEQUENCE IF NOT EXISTS account_tokens_id_seq;
+
+CREATE TABLE IF NOT EXISTS account_tokens (
+  id bigint PRIMARY KEY DEFAULT next_id_v1('account_tokens_id_seq'),
+  account_id bigint,
+  token text UNIQUE,
+  revoked boolean DEFAULT FALSE,
```
elliott-davis: Why would we keep revoked tokens in the db instead of removing the row?

components/builder-http-gateway/src/http/middleware.rs
```diff
@@ -138,30 +138,49 @@ impl Authenticated {
...
-                Ok(session)
+    fn authenticate(&self, req: &mut Request, token: &str) -> IronResult<Session> {
+        // Test hook - always create a valid session
+        if env::var_os("HAB_FUNC_TEST").is_some() {
+            debug!(
+                "HAB_FUNC_TEST: {:?}; calling session_create_short_circuit",
+                env::var_os("HAB_FUNC_TEST")
+            );
+            return session_create_short_circuit(req, token);
+        };
```
raskchanky: I love that we're checking for this first now.  Good call.

components/builder-http-gateway/src/http/middleware.rs
```diff
@@ -138,30 +138,49 @@ impl Authenticated {
...
-        let flags = FeatureFlags::from_bits(session.get_flags()).unwrap();
-        if !flags.contains(self.features) {
-            let err = NetError::new(ErrCode::ACCESS_DENIED, "net:auth:2");
-            return Err(IronError::new(err, Status::Forbidden));
+        if let Ok(session_token) = message::decode(&decoded_token) {
+            session_create(req, self.features, session_token)
+        } else {
+            // TBD: Deprecate github personal token support and remove this
+            // We are temporarily falling back to github personal access token
+            session_create_github(req, token)
```
raskchanky: Is it worth outputting a log message here, maybe on `info` level, that says something to the effect of "GH auth token used - this is deprecated and will be removed"?  Basically, just something that is reminding us to remove this at a later date, so we don't forget about it.

components/builder-http-gateway/src/http/middleware.rs
```diff
@@ -246,7 +227,32 @@ impl AfterMiddleware for Cors {
     }
 }
```
raskchanky: This entire function is SO MUCH BETTER now than it was before.  Wow.

components/builder-http-gateway/src/http/middleware.rs
```diff
@@ -246,7 +227,32 @@ impl AfterMiddleware for Cors {
...
+            }
+            Ok(session)
+        }
+        Err(err) => {
+            let status = net_err_to_http(err.get_code());
+            let body = itry!(serde_json::to_string(&err));
+            Err(IronError::new(err, (body, status)))
+        }
+    }
+}
```
raskchanky: I'm confused - this is the `session_create` function, but all we try to do is fetch a session and if one doesn't already exist, return an error?  Shouldn't we be routing a `SessionCreate` message?

components/builder-sessionsrv/src/migrations/2018-02-15-232946_personal_access_tokens/up.sql
raskchanky: Yeah maybe I'm missing something here as well, but I can't think why we'd want to keep revoked tokens in the db.

components/builder-sessionsrv/src/server/handlers.rs
```diff
@@ -116,6 +116,60 @@ pub fn account_find_or_create(
...
+
+pub fn account_tokens_get(
+    req: &mut Message,
+    conn: &mut RouteConn,
+    state: &mut ServerState,
+) -> SrvResult<()> {
+    let msg = req.parse::<proto::AccountTokensGet>()?;
+    match state.datastore.get_account_tokens(&msg) {
+        Ok(account_tokens) => {
+            println!("account_tokens_get: {:?}", account_tokens);
```
raskchanky: Was this meant to be `debug!`?

raskchanky:
Given that our tokens don't have fine grained permissions, I'm confused why people would need more than one of them. Is the capability to have more than one thinking ahead to a time when we'd be able to assign permissions to them, oauth-style?

components/builder-api/src/server/handlers.rs
```diff
@@ -223,6 +225,80 @@ pub fn get_profile(req: &mut Request) -> IronResult<Response> {
...
+
+    let mut request = AccountTokensGet::new();
+    request.set_account_id(session_id);
+    request.set_revoked(false);
+
+    match route_message::<AccountTokensGet, AccountTokens>(req, &request) {
+        Ok(account_tokens) => Ok(render_json(status::Ok, &account_tokens)),
+        Err(err) => return Ok(render_net_error(&err)),
+    }
+}
```
raskchanky: I'm probably missing something, but the presence of this function confuses me a little.  Don't you need to be authenticated to call this function?  And being authenticated would necessitate the use of an access token?  It feels like a chicken/egg problem.

chefsalim: Yeah its a bit confusing - we have multiple ways to auth.  The web auth flow is oauth -> session tokens.  That is not changing currently.  The personal access tokens are targeted to the hab cli (and build worker) flow.  In order for a user to generate/revoke their access token, they need to come to our web site.  Therefore, they will need to auth there first, and that is totally what we want.

components/builder-core/src/access_token.rs
```diff
@@ -0,0 +1,94 @@
...
+use std::path::PathBuf;
+use chrono::Utc;
+
+use hab_net::privilege::FeatureFlags;
+use error::{Error, Result};
+use integrations::{encrypt, decrypt};
+use protocol::{message, sessionsrv};
+
+pub const BUILDER_ACCOUNT_ID: u64 = 0;
+const ACCESS_TOKEN_PREFIX: &'static str = "TOKN";
```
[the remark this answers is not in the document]
chefsalim: tenor-141966054

components/builder-depot/src/handlers/integrations.rs
```diff
@@ -45,8 +45,11 @@ pub fn decrypt(req: &mut Request, content: &str) -> Result<String, Status> {
     );
     let depot = lock.read().expect("depot read lock is poisoned");
 
-    bldr_core::integrations::decrypt(&depot.config.key_dir, content)
-        .map_err(|_| status::InternalServerError)
+    let bytes = bldr_core::integrations::decrypt(&depot.config.key_dir, content)
+        .map_err(|_| status::InternalServerError)?;
+    Ok(String::from_utf8(bytes).map_err(
```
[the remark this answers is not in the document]
chefsalim: Yes..Status is returned as an Error component of the Result, so this map_err will return Err(status::InternalServerError)

components/builder-http-gateway/src/http/middleware.rs
chefsalim: Good point - I'd rather not log here, as we'll get a huge amount of logs, but I'm thinking we might want to put a deprecation warning in the hab cli itself if a non-access token is being used.

components/builder-core/src/access_token.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    Ok(format!("{}{}", ACCESS_TOKEN_PREFIX, ciphertext))
+}
+
+pub fn is_access_token(token: &str) -> bool {
+    token.starts_with(ACCESS_TOKEN_PREFIX)
+}
+
+pub fn validate_access_token(key_dir: &PathBuf, token: &str) -> Result<sessionsrv::Session> {
+    if !token.starts_with(ACCESS_TOKEN_PREFIX) {
```
christophermaier: Might just use `is_access_token(token)` here instead.

components/builder-http-gateway/src/http/middleware.rs
chefsalim: It took a lot of head beating... :)

components/builder-depot/src/handlers/integrations.rs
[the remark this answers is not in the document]
christophermaier: @elliott-davis I believe the `?` after the `map_err` call takes care of that.

components/builder-http-gateway/src/http/middleware.rs
chefsalim: Ah, the name isn't quite right.. the it's really session_validate - I'll fix it. The session is actually getting created elsewhere (either in the builder api oauth flow, or in the session_create_github for the cli auth flow).

components/builder-sessionsrv/src/migrations/2018-02-15-232946_personal_access_tokens/up.sql
chefsalim: This is a really really good question. The design is predicated currently on the notion that the universe of revoked tokens is smaller than the universe of actual tokens.  We want token validation to not require hitting the disk (or even network), so the notion is that we have a cache of revoked tokens to check against. However, an equally valid case can be made that over time the # of revoked tokens will grow to be larger than the set of valid tokens. In either case we're going to need to cache some set of tokens - either valid or revoked.  So the net of it is - we can go either way.  I'm happy to switch it over if folks think it will be cleaner/simpler.

components/builder-sessionsrv/src/server/handlers.rs
chefsalim: Good catch - it's a leftover from debugging I missed removing.

components/builder-core/src/access_token.rs
chefsalim: Yes!  this used to be assert using is_access_token, at one point I had removed that function and forgot to put it back as it was - good catch.

chefsalim:
@raskchanky correct, the idea is that we'll eventually support multiple tokens with more fine grained access, and this is early plumbing to support that future scenario.

components/builder-http-gateway/src/http/middleware.rs
raskchanky: Sounds good.  I don't know how important it is for that kind of information to be communicated back via straight HTTP, e.g. if someone was hitting the API via curl.  It feels like a nice-to-have.  I tend to use curl a lot when testing the API but I'm not sure how much actual use it gets by our users.

components/builder-http-gateway/src/http/middleware.rs
raskchanky: Perfect.

components/builder-api/src/server/handlers.rs
```diff
@@ -223,6 +225,73 @@ pub fn get_profile(req: &mut Request) -> IronResult<Response> {
...
+
+pub fn generate_access_token(req: &mut Request) -> IronResult<Response> {
+    let (session_id, flags) = {
+        let session = req.extensions.get::<Authenticated>().unwrap();
+        (session.get_id(), session.get_flags())
+    };
```
baumanj: I was momentarily confused as to why a tuple was used here. I expected `session_id` and `flags` to be used more since they got their own variables, but since they're each referenced only once, I wonder if it might be a bit simpler to eliminate the `let (session_id, flags)` like so:

```rust
    let session = req.extensions.get::<Authenticated>().unwrap();
…
    request.set_id(session.get_id());
…
    let token = bldr_core::access_token::generate_user_token(
        &cfg.depot.key_dir,
        account.get_id(),
        session.get_flags(),
    ).unwrap();
```

Not a major thing, but it makes the code a bit shorter and more evocative to my eye.

components/builder-core/src/access_token.rs
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) 2017 Chef Software Inc. and/or applicable contributors
```
baumanj: Should this be 2018?

components/builder-core/src/access_token.rs
```diff
@@ -0,0 +1,92 @@
...
+use integrations::{encrypt, decrypt};
+use protocol::{message, sessionsrv};
+
+pub const BUILDER_ACCOUNT_ID: u64 = 0;
+const ACCESS_TOKEN_PREFIX: &'static str = "TOKN";
+
+const BUILDER_TOKEN_EXPIRATION_SECONDS: i64 = 7200; // 2 hours
+const USER_TOKEN_EXPIRATION_SECONDS: i64 = 315360000; // 10 years
+
+pub fn generate_bldr_token(key_dir: &PathBuf) -> Result<String> {
```
baumanj: Considering this is used only one place, why not omit the constant, which allows the use of the more evocative `Duration` type? For example:

```rust
    let builder_token_lifetime = chrono::Duration::hours(2);

    generate_access_token(
        …
        builder_token_lifetime.num_seconds(),
    )
```

This may seem like a trivial thing, but I started thinking about it when I noticed `7200; // 2 hours`. I've dealt with code like this in the past where the value gets changed but the comment doesn't, leading to confusion.

components/builder-core/src/access_token.rs
```diff
@@ -0,0 +1,92 @@
...
+        BUILDER_TOKEN_EXPIRATION_SECONDS,
+    )
+}
+
+pub fn generate_user_token(key_dir: &PathBuf, account_id: u64, privileges: u32) -> Result<String> {
+    generate_access_token(
+        key_dir,
+        account_id,
+        privileges,
+        USER_TOKEN_EXPIRATION_SECONDS,
```
baumanj: Similar comment as above, but another nicety of using `Duration` is that it allows you to write `chrono::Duration:: max_value()`, assuming there's nothing particularly special about 10 years.

components/builder-core/src/access_token.rs
```diff
@@ -0,0 +1,92 @@
...
+    )
+}
+
+pub fn generate_access_token(
+    key_dir: &PathBuf,
+    account_id: u64,
+    flags: u32,
+    expiration: i64,
+) -> Result<String> {
+    let expires = Utc::now().timestamp() + expiration;
```
baumanj: If you use `chrono::Duration::max_value()` above, you'll have to handle this differently:
```rust
    let expires =Utc::now()
        .checked_add_signed(expiration)
        .unwrap_or(chrono::MAX_DATE.and_hms(0, 0, 0));
```

components/builder-api/src/server/handlers.rs
chefsalim: Yes, that's an unfortunate byproduct of the rust/iron/router combination not being ownership friendly.  If we don't wrap session in it's own scope, we'll get this error, because later on in the code the message routing we're passing the req mutably.

```
error[E0502]: cannot borrow `*req` as mutable because `req.extensions` is also borrowed as immutable
```

So we use this pattern all over the place in the code.

components/builder-core/src/access_token.rs
chefsalim: I seem to recall we concluded not worrying about the copyright date and not needing to change the standard header to new dates - although I think part of this code may be from 2017 anyways.

components/builder-core/src/access_token.rs
chefsalim: Cool, yeah I like that!

components/builder-api/src/server/handlers.rs
baumanj: Ah, that makes a lot of sense. I should've waited for my build to get this error before I made my suggestion 😬

components/builder-core/src/access_token.rs
chefsalim: Changing from 10 years to Duration::max_value() increases generated token size by 4 chars .. go figure.  Code is cleaner though, so I'll just go with it.

components/builder-core/src/access_token.rs
```diff
@@ -0,0 +1,93 @@
...
+pub fn is_access_token(token: &str) -> bool {
+    token.starts_with(ACCESS_TOKEN_PREFIX)
+}
+
+pub fn validate_access_token(key_dir: &PathBuf, token: &str) -> Result<sessionsrv::Session> {
+    assert!(is_access_token(token));
+
+    let encoded = token.to_owned().split_off(ACCESS_TOKEN_PREFIX.len());
+
+    let bytes = decrypt(key_dir, &encoded)?;
```
baumanj: For my rust edification, could this work equally well and obviate the creation of a `String`?
```rust
    let (_, encoded) = token.split_at(ACCESS_TOKEN_PREFIX.len());
    let bytes = decrypt(key_dir, encoded)?;
```

Also, do we want to panic if `validate_access_token` is called with a `token` shorter than `ACCESS_TOKEN_PREFIX`? Presumably that won't happen since the one consumer checks `is_access_token` first, but perhaps the panic should be called out in a doc comment.

chefsalim: Oh yes, that's a good catch.. it may still require a to_owned but if not, it's should be perfect - don't know how I missed that... so many string functions.  Re: the panic, yes, that should never happen since we're explicitly checking prefix size earlier..

components/builder-core/src/access_token.rs
```diff
@@ -0,0 +1,93 @@
...
+        Ok(p) => p,
+        Err(e) => {
+            warn!("Unable to deserialize access token, err={:?}", e);
+            return Err(Error::TokenInvalid);
+        }
+    };
+
+    let now = Utc::now().timestamp();
+
+    if payload.get_expires() < now {
```
baumanj: This might seem like overkill, but I like taking maximum advantage of type-safety by comparing the `DateTime`s rather than raw integers:
```rust
    let expiration = Utc.timestamp(payload.get_expires(), 0 /* nanoseconds */);

    if expiration < Utc::now() {
```

components/builder-http-gateway/src/http/middleware.rs
```diff
@@ -138,30 +138,49 @@ impl Authenticated {
...
+        if env::var_os("HAB_FUNC_TEST").is_some() {
+            debug!(
+                "HAB_FUNC_TEST: {:?}; calling session_create_short_circuit",
+                env::var_os("HAB_FUNC_TEST")
+            );
+            return session_create_short_circuit(req, token);
+        };
+
+        // Check for a valid personal access token
+        if bldr_core::access_token::is_access_token(token) {
```
baumanj: Is there any possibility that the base64-encoding of an internal sessionsrv token could start with `TOKN`? Perhaps we should make `ACCESS_TOKEN_PREFIX` include some non-base64 character like `!`.

baumanj:
I haven't reviewed all the code yet, but I'm about to stop work for the weekend, so I'll resume on Monday.

Generally things look really, really good. One general comment: have you considered writing any tests for the API changes? We have pretty good support for running them locally and in Travis now.

components/builder-core/src/access_token.rs
chefsalim: That seems reasonable.. the only caveat is that looks like Utc.timestamp(..) can panic.  In that case the direct compare would probably give undefined behavior.  However for the case of user tokens, the undefined behavior would be _more_ conservative, ie, we would not allow use of the token.  So on the whole, we're probably better off with the direct compare here I think.

components/builder-http-gateway/src/http/middleware.rs
chefsalim: No, there's no possibility of that happening. The possibility of that happening is the same as having an encrypted version of a plaintext reveal several consecutive characters of the plaintext in the encrypted output, which is nil. The bigger issue is needing the prefix at all - it's a bit of a ugly hack - ideally I'd like to eliminate it completely, but there are several tradeoffs to doing so at the moment.

chefsalim:
Thanks for the reminder on the tests @baumanj ... I'll look into getting some tests up, it will probably be a separate PR.

components/builder-http-gateway/src/http/middleware.rs
chefsalim: Update - this explanation isn't quite right.  However the issue is moot because there is a significant advantage of using a non-base-64 char in the prefix - ie, we can eliminate all the other prefix chars, since we only need to differentiate between access token and session token (and github token, for now).  So we'll save 3 extra characters from the token size!  Which brings us back to the size before the duration change was made.. pretty cool :)

components/builder-protocol/protocols/sessionsrv.proto
```diff
@@ -36,6 +36,36 @@ message AccountUpdate {
   optional string email = 2;
 }
 
+message AccessToken {
+    optional uint64 account_id = 1;
+    optional uint32 flags = 2;
+    optional int64 expires = 3;
```
baumanj: Would this make more sense as a `uint64`? We don't have negative timestamps, do we?

chefsalim: Yeah.. it should ideally be u64.. however chrono timestamp() returns i64, and postgres natively stores i64, so might as well leave it i64 and not do conversions all over the place.

baumanj: On second thought, `chrono` supports it for the sake of representing dates before the epoch, so maybe we should keep it.

components/builder-core/src/access_token.rs
baumanj: I don't understand why we can conclude that the undefined behavior would be more conservative. Let's consider the situations where `Utc.timestamp()` panics:
1. Out-of-range number of seconds
2. invalid nanosecond

#2 is a non-issue since we're using literal 0. #1 Should only be a problem if we end up outside the representable range that `chrono` supports (roughly 262145 BCE–262143 CE). If the expiration timestamp in the payload represented a time _before_ that, we would not allow use of the token, but if it represented a time _after_ we *would*. Now, we don't expect to ever be able to get those sorts of values short of a logic error or data corruption. In those cases, I think panicking may be appropriate. However, if we want to not panic, we can use `timestamp_opt()` to ensure we deny access when the unexpected occurs and allow us to make a note of the unexpected situation without crashing:
```rust
    match Utc.timestamp_opt(payload.get_expires(), 0 /* nanoseconds */) {
        Some(time) => {
            if time < Utc::now() {
                Err(Error::TokenExpired)
            } else {
                Ok(payload.into())
            }
        }
        _ => Err(Error::TokenInvalid),
    }
```

baumanj:
I've been through all the code now and Aside from the one open point of discussion (and tests to be added later), I think it looks good. The only other question that occurred to me was more of a design-level thing.

In GitHub's token implementation you can only see the value of a token at the time it's generated. There's nothing akin to the `GET access-tokens` interface we have here. Do we understand why GH did things that way, and if so, do we think our situation is different, or do we just disagree with GH's approach?

components/builder-core/src/access_token.rs
chefsalim: 👍 to timestamp_opt - that works.  Yeah, definitely want to avoid panics if at all possible.  Since user tokens don't expire, we don't have to worry about allowing access if the seconds are out of range - revocation won't depend on this.  Builder tokens do expire, so that's the most of the exposure.

chefsalim:
Thanks @baumanj .. the get access token result is actually not to original spec.  The original spec calls for only a couple of fields, but I've left additional fields including the token in.  We can strip out the access token from the GET.  Overall it's not clear that it actually adds anything to the security, since if you have access to make the GET call, you certainly have access to re-generate a new token.  When we've discussed this in the past, a valid reason for now showing things like secret keys during generation is if you decide to not store those secrets in your backend.  If you're have the keys, and allow regeneration, there's no real reason to hide them from the user.  We had the same discussion with our secret keys, and decided to allow users to download existing keys.  Perhaps with tokens there is more of a risk that someone takes a screen shot while you are browsing your profile.  I want to think about this more in course of the revocation work. We might be better off allowing user's to 'reveal' an existing token rather than only allowing regeneration.

chefsalim:
@baumanj FYI, PR is updated with some basic integration tests

baumanj:
Looks great! Merge at will!

PR #230, opened 2018-02-27
Add initial support for configurable OAuth providers

cnunciato:
Currently, our OAuth support is wound pretty tightly around GitHub, and this is something we’d like to change. This PR adds initial support for configurable OAuth providers, authorization URLs and variable URL parameters by assembling them from configuration based on the specified provider (which, it should be noted, is still limited to GitHub).


thesentinels:

Thank you for contributing!

components/builder-web/app/actions/oauth.ts
```diff
@@ -0,0 +1,170 @@
+// Copyright (c) 2016-2017 Chef Software Inc. and/or applicable contributors
```
elliott-davis: This should be 2016-2018

components/builder-web/app/reducers/oauth.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2016-2017 Chef Software Inc. and/or applicable contributors
```
elliott-davis: Same copyright info as comment above

components/builder-web/app/util.ts
```diff
@@ -32,6 +32,32 @@ export function createGitHubLoginUrl(state) {
   return `${urlPrefix}?${queryString}`;
 }
 
+export function createLoginUrl() {
+  const store = new AppStore();
+  let params: any = {};
+  let uri;
+
+  switch (config.oauth_provider) {
+    case 'github':
```
elliott-davis: Is there a default case we can add for generic oauth providers?

components/builder-web/app/actions/oauth.ts
raskchanky: Last I heard, it was sufficient to just put the year it was written, so 2018 should be fine.  That's what I've been putting in new files.

components/builder-web/app/util.ts
raskchanky: This sounds good, but I fear would be pretty difficult in practice.  Bitbucket's setup and URLs are different from GitHub's and I bet if we checked Gitlab, those would be slightly different as well.  I'm not sure there's a "standard" set of URLs for oauth.

raskchanky:
This looks good to me, certainly as a first pass.  I'm guessing we might need to tweak some things once the server side stuff is done and we can actually test it out with multiple providers, but let's get this merged and go from there.

components/builder-web/app/util.ts
cnunciato: That's right, there isn't really a default case from our perspective -- there are conventions for OAuth paths and param names and the like, but that's about it. Our list of supported providers will also be explicit, and as of now, it's still only GitHub.

components/builder-web/app/actions/oauth.ts
cnunciato: Thanks, I'll fix this up.

cnunciato:
@elliott-davis Good call on the dev-env tweak; I missed that.

@raskchanky Absolutely, it does feel like there's room for some nicer abstraction here for sure. I just figured I'd wait to see how the BitBucket stuff took shape, and A2 as well, before going too far down that road.

PR #232, opened 2018-02-28
Elliott/add secret api

elliott-davis:

### DIFF
--- a/components/builder-depot/src/server.rs
+++ b/components/builder-depot/src/server.rs
@@ -16,7 +16,7 @@ use std::fs::{self, File};
 use std::path::PathBuf;
 use std::io::{BufWriter, Read, Write};
 use std::result;
-use std::str::FromStr;
+use std::str::{from_utf8, FromStr};
 
 use base64;
 use bldr_core;
@@ -26,12 +26,13 @@ use bodyparser;
 use github_api_client::GitHubClient;
 use hab_core::package::{ident, FromArchive, Identifiable, PackageArchive, PackageIdent,
                         PackageTarget};
-use hab_core::crypto::keys::{PairType, parse_key_str};
+use hab_core::crypto::keys::{PairType, parse_key_str, parse_name_with_rev};
 use hab_core::crypto::BoxKeyPair;
 use hab_core::crypto::PUBLIC_BOX_KEY_VERSION;
 use http_gateway::http::controller::*;
 use http_gateway::http::helpers::{self, all_visibilities, check_origin_access, check_origin_owner,
-                                  dont_cache_response, get_param, visibility_for_optional_session};
+                                  dont_cache_response, get_param, validate_params,
+                                  visibility_for_optional_session};
 use http_gateway::http::middleware::{SegmentCli, XRouteClient};
 use hab_net::{ErrCode, NetOk, NetResult};
 use hab_net::privilege::FeatureFlags;
@@ -68,6 +69,12 @@ struct OriginCreateReq {
 #[derive(Clone, Serialize, Deserialize)]
 struct OriginUpdateReq {
     default_package_visibility: String,
+}
+
+#[derive(Clone, Serialize, Deserialize)]
+struct OriginSecretPayload {
+    name: String,
+    value: String,
 }
 
 const ONE_YEAR_IN_SECS: usize = 31536000;
@@ -439,55 +446,91 @@ fn write_archive(filename: &PathBuf, body: &mut Body) -> Result<PackageArchive> 
 }
 
 fn download_latest_origin_encryption_key(req: &mut Request) -> IronResult<Response> {
-    let origin = match get_param(req, "origin") {
-        Some(origin) => origin,
-        None => return Ok(Response::with(status::BadRequest)),
+    let params = match validate_params(req, &["origin"]) {
+        Ok(p) => p,
+        Err(st) => return Ok(Response::with(st)),
     };
 
-    if !check_origin_access(req, &origin).unwrap_or(false) {
-        return Ok(Response::with(status::Forbidden));
-    }
-
     let mut request = OriginPublicEncryptionKeyLatestGet::new();
-    match helpers::get_origin(req, origin) {
+    let origin = match helpers::get_origin(req, &params["origin"]) {
         Ok(mut origin) => {
             request.set_owner_id(origin.get_owner_id());
             request.set_origin(origin.take_name());
+            origin
         }
         Err(err) => return Ok(render_net_error(&err)),
-    }
+    };
+
     let key = match route_message::<
         OriginPublicEncryptionKeyLatestGet,
         OriginPublicEncryptionKey,
     >(req, &request) {
         Ok(key) => key,
-        Err(err) => return Ok(render_net_error(&err)),
+        Err(err) => {
+            if err.get_code() == ErrCode::ENTITY_NOT_FOUND {
+                match generate_origin_encryption_keys(&origin, req) {
+                    Ok(key) => key,
+                    Err(Error::NetError(e)) => return Ok(render_net_error(&e)),
+                    Err(_) => unreachable!(),
+                }
+            } else {
+                return Ok(render_net_error(&err));
+            }
+        }
     };
 
     let xfilename = format!("{}-{}.pub", key.get_name(), key.get_revision());
     download_content_as_file(key.get_body(), xfilename)
 }
 
-fn generate_origin_encryption_keys(req: &mut Request) -> IronResult<Response> {
-    debug!("Generate Origin Encryption Keys {:?}", req);
-    match get_param(req, "origin") {
-        Some(origin) => {
-            if !check_origin_access(req, &origin).unwrap_or(false) {
-                return Ok(Response::with(status::Forbidden));
-            }
+fn generate_origin_encryption_keys(
+    origin: &Origin,
+    req: &mut Request,
+) -> Result<OriginPublicEncryptionKey> {
+    debug!("Generate Origin Encryption Keys {:?} for {:?}", req, origin);
 
-            match helpers::get_origin(req, origin) {
-                Ok(origin) => {
-                    match helpers::generate_origin_encryption_keys(req, origin) {
-                        Ok(_) => Ok(Response::with(status::Created)),
-                        Err(err) => Ok(render_net_error(&err)),
-                    }
-                }
-                Err(err) => Ok(render_net_error(&err)),
-            }
-        }
-        None => Ok(Response::with(status::BadRequest)),
+    let mut public_request = OriginPublicEncryptionKeyCreate::new();
+    let mut private_request = OriginPrivateEncryptionKeyCreate::new();
+    let mut public_key = OriginPublicEncryptionKey::new();
+    let mut private_key = OriginPrivateEncryptionKey::new();
+    {
+        let session = req.extensions.get::<Authenticated>().unwrap();
+        public_key.set_owner_id(session.get_id());
+        private_key.set_owner_id(session.get_id());
     }
+    public_key.set_name(origin.get_name().to_string());
+    public_key.set_origin_id(origin.get_id());
+    private_key.set_name(origin.get_name().to_string());
+    private_key.set_origin_id(origin.get_id());
+
+    let pair = BoxKeyPair::generate_pair_for_origin(origin.get_name())
+        .map_err(Error::HabitatCore)?;
+    public_key.set_revision(pair.rev.clone());
+    public_key.set_body(
+        pair.to_public_string()
+            .map_err(Error::HabitatCore)?
+            .into_bytes(),
+    );
+    private_key.set_revision(pair.rev.clone());
+    private_key.set_body(
+        pair.to_secret_string()
+            .map_err(Error::HabitatCore)?
+            .into_bytes(),
+    );
+
+    public_request.set_public_encryption_key(public_key);
+    private_request.set_private_encryption_key(private_key);
+
+    let key = route_message::<OriginPublicEncryptionKeyCreate, OriginPublicEncryptionKey>(
+        req,
+        &public_request,
+    )?;
+    route_message::<OriginPrivateEncryptionKeyCreate, OriginPrivateEncryptionKey>(
+        req,
+        &private_request,
+    )?;
+
+    Ok(key)
 }
 
 fn generate_origin_keys(req: &mut Request) -> IronResult<Response> {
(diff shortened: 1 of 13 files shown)


thesentinels:

Thank you for contributing!

components/builder-depot/src/server.rs
```diff
@@ -130,7 +137,12 @@ pub fn origin_create(req: &mut Request) -> IronResult<Response> {
     }
 
     match route_message::<OriginCreate, Origin>(req, &request) {
-        Ok(origin) => Ok(render_json(status::Created, &origin)),
+        Ok(origin) => {
+            // TED: If this fails we need to delete the origin and report an origin create failure
+            // currently there is no mechanism for deleting an origin
+            generate_origin_encryption_keys(origin.get_name(), req)?;
```
chefsalim: Hmm, yeah this is a bit tricky if there's no way for a user to recover.  
A secondary question is, how will we seed the already existing origins?

components/builder-depot/src/server.rs
```diff
@@ -461,32 +473,34 @@ fn download_latest_origin_encryption_key(req: &mut Request) -> IronResult<Respon
         OriginPublicEncryptionKey,
     >(req, &request) {
         Ok(key) => key,
-        Err(err) => return Ok(render_net_error(&err)),
+        Err(err) => {
+            if err.get_code() == ErrCode::ENTITY_NOT_FOUND {
+                generate_origin_encryption_keys(&origin, req)?;
+                return download_latest_origin_encryption_key(req);
```
chefsalim: Not a fan of a recursive call here - would be better to make this linear - generate_origin_encryption_keys could return the keys, not an IronResponse.

components/builder-depot/src/server.rs
```diff
@@ -461,32 +473,34 @@ fn download_latest_origin_encryption_key(req: &mut Request) -> IronResult<Respon
...
 }
 
-fn generate_origin_encryption_keys(req: &mut Request) -> IronResult<Response> {
-    debug!("Generate Origin Encryption Keys {:?}", req);
-    match get_param(req, "origin") {
-        Some(origin) => {
-            if !check_origin_access(req, &origin).unwrap_or(false) {
-                return Ok(Response::with(status::Forbidden));
-            }
+fn generate_origin_encryption_keys(origin: &str, req: &mut Request) -> IronResult<Response> {
```
chefsalim: I'm not sure we even need a generation function at this level at all with the new strategy - this could all get collapsed inside of the get_origin_encryption_keys in originsrv (maybe renamed to get_or_create_origin_encrytpion_keys w/ a GetOrCreateOriginEncryptionKeys message).

components/builder-depot/src/server.rs
```diff
@@ -2064,6 +2078,195 @@ pub fn download_latest_builder_key(req: &mut Request) -> IronResult<Response> {
...
+                    status::UnprocessableEntity,
+                    "Missing value for field: `value`",
+                )));
+            }
+            body
+        }
+        _ => return Ok(Response::with(status::UnprocessableEntity)),
+    };
+
+    if !check_origin_access(req, origin).unwrap_or(false) {
```
chefsalim: You could move up the check and do it before even getting the body..

chefsalim: Actually, validate_params already does  an origin check..

components/builder-depot/src/server.rs
```diff
@@ -2064,6 +2078,195 @@ pub fn download_latest_builder_key(req: &mut Request) -> IronResult<Response> {
...
+        }
+        Err(err) => return Ok(render_net_error(&err)),
+    }
+    let priv_key = match route_message::<
+        OriginPrivateEncryptionKeyGet,
+        OriginPrivateEncryptionKey,
+    >(req, &db_priv_request) {
+        Ok(key) => {
+            let key_str = from_utf8(key.get_body()).unwrap();
+            debug!("KEY2: {:?}", key_str);
```
chefsalim: Maybe remove the debug, or change KEY2: to something more meaningful..

components/builder-depot/src/server.rs
```diff
@@ -2064,6 +2078,195 @@ pub fn download_latest_builder_key(req: &mut Request) -> IronResult<Response> {
...
+    request.set_secret(secret);
+
+    match route_message::<OriginSecretCreate, OriginSecret>(req, &request) {
+        Ok(_) => Ok(Response::with(status::Created)),
+        Err(err) => Ok(render_net_error(&err)),
+    }
+}
+
+pub fn list_origin_secrets(req: &mut Request) -> IronResult<Response> {
+    let origin_name = match get_param(req, "origin") {
```
chefsalim: You could use validate_params here as well..

components/builder-originsrv/src/migrations/2018-03-05-224023_origin_secrets/down.sql
```diff
@@ -0,0 +1 @@
+-- This file should undo anything in `up.sql`
```
chefsalim: Are we leaving these empty?  If so, would be good to add that prescriptive guidance to the migrations doc if we are not able to take advantage of the diesel rollbacks.

components/builder-originsrv/src/migrations/2018-03-05-224023_origin_secrets/up.sql
```diff
@@ -0,0 +1,53 @@
...
+  os_name varchar(256),
+  os_value text
+) RETURNS SETOF origin_secrets AS $$
+    BEGIN
+      RETURN QUERY INSERT INTO origin_secrets (origin_id, name, value)
+          VALUES (os_origin_id, os_name, os_value)
+          RETURNING *;
+      RETURN;
+    END
+$$ LANGUAGE plpgsql VOLATILE;
```
chefsalim: We wanted to standardize on SQL instead of plpgsql where possible - consider switching it over.

components/builder-originsrv/src/migrations/2018-03-05-224023_origin_secrets/up.sql
```diff
@@ -0,0 +1,53 @@
...
+$$ LANGUAGE SQL STABLE;
+
+CREATE OR REPLACE FUNCTION delete_origin_secret_v1 (
+  os_origin_id bigint,
+  os_name text
+) RETURNS SETOF origin_secrets AS $$
+  BEGIN
+      DELETE FROM origin_secrets WHERE name = os_name AND origin_id = os_origin_id;
+  END
+$$ LANGUAGE plpgsql VOLATILE;
```
chefsalim: Same - convert to SQL

components/builder-depot/src/server.rs
chefsalim: If we are also seeding during the first download, then perhaps this generate call should be removed from here..

elliott-davis: Yeah it's a bit of a pickle. I'm happy to write an internal only origin delete but I was afraid to do it as part of this PR. To answer the secondary question: If a user requests to upload a key and their origin doesn't have an existing encryption key, we generate one and the cli downloads it, then continues on as normal.

chefsalim: OK yeah.. is there any use case that would not be satisfied if we just removed this generate call from here then?

elliott-davis: It's probably safe to remove here

components/builder-originsrv/src/migrations/2018-03-05-224023_origin_secrets/down.sql
elliott-davis: Yeah the rollbacks are sort of pointless for our usecase. I'll add a card to write up a doc on it

components/builder-originsrv/src/migrations/2018-03-05-224023_origin_secrets/up.sql
```diff
@@ -0,0 +1,53 @@
+CREATE SEQUENCE IF NOT EXISTS origin_secrets_id_seq;
+
+CREATE TABLE IF NOT EXISTS origin_secrets (
+  id bigint PRIMARY KEY DEFAULT next_id_v1('origin_secrets_id_seq'),
+  origin_id bigint REFERENCES origins(id),
+  owner_id bigint,
+  name varchar(256),
```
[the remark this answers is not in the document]
elliott-davis: I think we should be using varchar in more places. Using text doesn't make sense for things like names.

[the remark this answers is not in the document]
chefsalim: Hmm, why is that? It seems we're getting more flexibility with text, with basically same cost under the covers. Another thing that might be weird is comparison between text and varchars, looks like there are subtle differences in implicit comparison.  It might be better to just be consistent with the rest of the code for now... unless there's some other strong reason?

components/builder-originsrv/src/migrations/2018-03-05-224023_origin_secrets/down.sql
elliott-davis: https://github.com/habitat-sh/builder/issues/275

Cargo.toml
```diff
@@ -18,6 +18,5 @@ members = [
   "components/builder-worker",
   "components/github-api-client",
   "components/net",
-  "components/op",
-  "components/segment-api-client"
```
raskchanky: `segment-api-client` should remain in this repo.  I think @christophermaier just moved it over here the other day?

christophermaier: Yes, please leave it!

components/builder-depot/src/server.rs
```diff
@@ -439,17 +446,15 @@ fn write_archive(filename: &PathBuf, body: &mut Body) -> Result<PackageArchive>
...
-        Some(origin) => origin,
-        None => return Ok(Response::with(status::BadRequest)),
+    let params = match validate_params(req, &["origin"]) {
+        Ok(p) => p,
+        Err(st) => return Ok(Response::with(st)),
     };
 
-    if !check_origin_access(req, &origin).unwrap_or(false) {
-        return Ok(Response::with(status::Forbidden));
-    }
```
raskchanky: Why was this check removed?

components/builder-depot/src/server.rs
```diff
@@ -461,32 +466,76 @@ fn download_latest_origin_encryption_key(req: &mut Request) -> IronResult<Respon
...
 
-            match helpers::get_origin(req, origin) {
-                Ok(origin) => {
-                    match helpers::generate_origin_encryption_keys(req, origin) {
-                        Ok(_) => Ok(Response::with(status::Created)),
-                        Err(err) => Ok(render_net_error(&err)),
-                    }
-                }
-                Err(err) => Ok(render_net_error(&err)),
+    match helpers::get_origin(req, origin) {
```
raskchanky: Is `generate_origin_encryption_keys` only called in one place?  I couldn't tell definitively from this diff, but if the `download_latest_origin_encryption_key` function above is the only place that calls it, you can probably skip this call to `helpers::get_origin` (which I think makes a network trip to the db server, no?) since `download_latest_origin_encryption_key` already does that.

components/builder-depot/src/server.rs
```diff
@@ -461,32 +466,76 @@ fn download_latest_origin_encryption_key(req: &mut Request) -> IronResult<Respon
...
+                public_key.set_owner_id(session.get_id());
+                private_key.set_owner_id(session.get_id());
             }
+            public_key.set_name(origin.get_name().to_string());
+            public_key.set_origin_id(origin.get_id());
+            private_key.set_name(origin.get_name().to_string());
+            private_key.set_origin_id(origin.get_id());
+
+            let pair = BoxKeyPair::generate_pair_for_origin(origin.get_name())
+                .expect("failed to generate origin encryption key pair");
```
raskchanky: Since the function we're in returns a `Result`, is it worth returning an `Err` here instead of calling `expect`?  I remember there was some conversation around the use of `expect` with the crypto functions, and I'm guessing that's how we ended up here.  It looks like the error module in this crate already has an enum for errors from the core crate, so it might work to just turn this into:

```rust
let pair = BoxKeyPair::generate_pair_for_origin(origin.get_name()).map_err(Error::HabitatCore)?;
```

If that doesn't work though, I wouldn't pursue it further.  It looked like an easy win, but maybe it's not.

components/builder-depot/src/server.rs
```diff
@@ -461,32 +466,76 @@ fn download_latest_origin_encryption_key(req: &mut Request) -> IronResult<Respon
...
+
+            let key =
+                route_message::<OriginPublicEncryptionKeyCreate, OriginPublicEncryptionKey>(
+                    req,
+                    &public_request,
+                )?;
+            route_message::<OriginPrivateEncryptionKeyCreate, OriginPrivateEncryptionKey>(
+                req,
+                &private_request,
+            )?;
```
raskchanky: What happens if the `OriginPublicEncryptionKeyCreate` succeeds but the `OriginPrivateEncryptionKeyCreate` fails?  Is the user left in a bad state?  If they try to create a new key again, will it succeed if this was aborted in a partially finished state?

components/builder-depot/src/server.rs
```diff
@@ -2064,6 +2113,186 @@ pub fn download_latest_builder_key(req: &mut Request) -> IronResult<Response> {
...
+                    status::UnprocessableEntity,
+                    "Missing value for field: `name`",
+                )));
+            }
+            if body.value.len() <= 0 {
+                return Ok(Response::with((
+                    status::UnprocessableEntity,
+                    "Missing value for field: `value`",
+                )));
+            }
```
raskchanky: I'm so stoked you're validating this kind of stuff.  I wish we had a better mechanism for DRYing this kind of thing up in our codebase.  (And just to be clear, I realize that adding one is outside of the scope of this PR 😁 )

components/builder-depot/src/server.rs
```diff
@@ -2064,6 +2113,186 @@ pub fn download_latest_builder_key(req: &mut Request) -> IronResult<Response> {
...
+        Ok(res) => res,
+        Err(e) => {
+            return Ok(Response::with((
+                status::UnprocessableEntity,
+                format!("Failed to get metadata from payload: {}", e),
+            )))
+        }
+    };
+
+    debug!("Secret Metadata: {:?}", secret_metadata);
```
raskchanky: Will this make private keys show up in our logs?

Cargo.toml
elliott-davis: oops - rebase fail

components/builder-depot/src/server.rs
```diff
@@ -2064,6 +2113,186 @@ pub fn download_latest_builder_key(req: &mut Request) -> IronResult<Response> {
...
+    };
+
+    debug!("Using key {:?}-{:?}", name, &rev);
+
+    // fetch the public origin encryption key from the database
+    let mut db_pub_request = OriginPublicEncryptionKeyGet::new();
+    match helpers::get_origin(req, origin) {
+        Ok(mut origin) => {
+            db_pub_request.set_owner_id(origin.get_owner_id());
+            db_pub_request.set_origin(origin.take_name());
```
raskchanky: Is it worth caching these values in local variables from the call to `helpers::get_origin` above to avoid another network round trip?

components/builder-depot/src/server.rs
elliott-davis: validate_params does the chef for us

components/builder-depot/src/server.rs
```diff
@@ -2064,6 +2113,186 @@ pub fn download_latest_builder_key(req: &mut Request) -> IronResult<Response> {
...
+                (status::UnprocessableEntity, format!("{}", e)),
+            ))
+        }
+    };
+
+    let mut secret = OriginSecret::new();
+    secret.set_name(body.name);
+    secret.set_value(body.value.clone());
+    match helpers::get_origin(req, &origin) {
+        Ok(origin) => secret.set_origin_id(origin.get_id()),
```
raskchanky: Some comment here re: caching values from the first call to `helpers::get_origin`.

components/builder-depot/src/server.rs
```diff
@@ -2311,13 +2540,8 @@ where
...
         "origin_key",
     );
-    r.post(
-        "/origins/:origin/encryption_keys",
-        XHandler::new(generate_origin_encryption_keys).before(basic.clone()),
-        "origin_encryption_key_generate",
-    );
     r.get(
-        "/origins/:origin/encryption_keys/latest",
+        "/origins/:origin/encryption_key",
```
raskchanky: It might be worth adding the `breaking-changes` label (do we have one yet?) because of this part - just for the sake of release notes.  I think the odds of anyone using our HTTP API directly (independent of the `hab` CLI) are low, but just in case, it's worth pointing out that this is a breaking change.

components/builder-originsrv/src/data_store.rs
```diff
@@ -1813,6 +1825,83 @@ impl DataStore {
         oin.set_names(names);
         oin
     }
+    pub fn create_origin_secret(
+        &self,
+        os: &originsrv::OriginSecretCreate,
```
raskchanky: Shouldn't this be `osc`?  :trollface: KIDDING KIDDING

components/builder-originsrv/src/data_store.rs
```diff
@@ -1813,6 +1825,83 @@ impl DataStore {
...
+            "SELECT * FROM insert_origin_secret_v1($1, $2, $3)",
+            &[
+                &(os.get_secret().get_origin_id() as i64),
+                &os.get_secret().get_name(),
+                &os.get_secret().get_value(),
+            ],
+        ).map_err(SrvError::OriginSecretCreate)?;
+        match rows.iter().nth(0) {
+            Some(row) => Ok(self.row_to_origin_secret(row)),
+            None => Err(SrvError::NoRowsReturnedAfterInsert()),
```
raskchanky: Wow, this is _fantastic_.

components/builder-originsrv/src/data_store.rs
```diff
@@ -1813,6 +1825,83 @@ impl DataStore {
...
+        &self,
+        os_get: &originsrv::OriginSecretGet,
+    ) -> SrvResult<Option<originsrv::OriginSecret>> {
+        let conn = self.pool.get(os_get)?;
+        let rows = &conn.query(
+            "SELECT * FROM get_origin_secret_v1($1, $2)",
+            &[&(os_get.get_origin_id() as i64), &os_get.get_name()],
+        ).map_err(SrvError::OriginSecretGet)?;
+        if rows.len() != 0 {
+            // We just checked - we know there is a value here
```
raskchanky: All the comments justifying `unwrap()`s are one of my favorite parts of our code.  😆

components/builder-protocol/protocols/originsrv.proto
```diff
@@ -413,6 +413,36 @@ message OriginProjectList {
...
+// Origin Secret
+message OriginSecret {
+  optional uint64 id = 1;
+  optional uint64 origin_id = 2;
+  optional string name = 3;
+  optional string value = 4;
+}
+
+message OriginSecretCreate {
+  optional OriginSecret secret = 1;
```
raskchanky: I ❤️ this pattern.

components/builder-protocol/src/originsrv.rs
```diff
@@ -1158,6 +1158,66 @@ impl Serialize for OriginProjectList {
     }
 }
 
+impl Serialize for OriginSecret {
+    fn serialize<S>(&self, serializer: S) -> result::Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut state = serializer.serialize_struct("secret", 2)?;
```
raskchanky: I think the number in this call refers to the number of fields being serialized in the struct, so I _think_ this should be 4.  I'm assuming everything serializes normally though, so honestly, I'm a little baffled.

components/builder-depot/src/server.rs
raskchanky: Oh, nice.  Good call.

components/builder-depot/src/server.rs
elliott-davis: No, it's just metadata about the encrypted payload. It will return the values here https://github.com/habitat-sh/core/blob/master/components/core/src/crypto/keys/box_key_pair.rs#L32-L37

raskchanky: Ah, perfect.

raskchanky:
This is great stuff.  I left some comments that might be worth thinking about but nothing that looked like a huge blocker.

components/builder-originsrv/src/data_store.rs
elliott-davis: :shaved_ice:

components/builder-depot/src/server.rs
elliott-davis: good point. I thought it was safe since we hadn't shipped to prod yet, but we did that yesterday

elliott-davis:
@raskchanky @chefsalim I added the breaking change label to this since I changed an api endpoint. I doubt anyone is using the encryption keys feature since it's in the api only. Better safe than sorry.

components/builder-depot/src/server.rs
```diff
@@ -439,55 +446,91 @@ fn write_archive(filename: &PathBuf, body: &mut Body) -> Result<PackageArchive>
...
+    private_request.set_private_encryption_key(private_key);
+
+    let key = route_message::<OriginPublicEncryptionKeyCreate, OriginPublicEncryptionKey>(
+        req,
+        &public_request,
+    )?;
+    route_message::<OriginPrivateEncryptionKeyCreate, OriginPrivateEncryptionKey>(
+        req,
+        &private_request,
+    )?;
```
raskchanky: So if I'm understanding the existing code correctly, if someone hits this function and the call to create the public key succeeds, and the call to create the private key fails for some transient reason, it's not a big deal.  We'll return an error to the user, but if the user tries again (assuming the transient issue has resolved), it will work and they'll be all set.  From what I could tell, the only unique constraint we have is on the full name of the key, which includes the revision, and the revision will change every time they generate a new key.  Do I have all that correct?

If so, then this seems fine as is.

elliott-davis: yep that is all correct. I am thinking about revisiting the logic there to do an update on the public keys but that's a future iteration

elliott-davis:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/builder/builds/352462892).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/builder/builds/352462892).

PR #235, opened 2018-03-01
Add revoked token check

chefsalim:
This change adds the ability for revoked access tokens to get properly handled. We keep a cache of access tokens in memory in the sessionsrv, and allow for a validation check without hitting the DB (if the token is in cache). For simplicity, the current cache design supports only a single valid token / account id, and cached items do not age out - we can visit those tradeoffs when needed at a later date.

tenor-109821243


thesentinels:

Thank you for contributing!

nellshamrell:
👍

PR #262, opened 2018-03-07
Create the tmuxinator directory if it doesn't already exist

baumanj:
Also, clean up shellcheck lints,make the script run correctly regardless of the
current directory, and provide some instructive output if run without args.

I removed the "TODO: constrain the list of possible environments?" comment since
I think that is more the domain of the generate-habitat-env-tmuxinator.sh
script. If we added new environments, we don't want to have to modify this.

### DIFF
--- a/tools/ssh_helpers/hab-env
+++ b/tools/ssh_helpers/hab-env
@@ -2,15 +2,17 @@
 
 set -euo pipefail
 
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

PR #263, opened 2018-03-07
Refactor metrics infrastructure

christophermaier:
This commit involves several refactorings of our existing metrics code
that should make it more flexible moving forward. It evolved through
several iterations, so I'll attempt to lay out the story here in
prose, rather than individual commits; sorry :/

Since our existing metrics are mainly around Github API interactions,
that's what most of this discussion will focus on.

First, rather than instrumenting each call site where we make a Github
API call, our metrics calls are now embedded within our Github API
client library; anything that calls it will automatically be measured.

Second, a new set of Github token types are introduced: `AppToken` and
`UserToken`. The former bundles together an API token with the Github
App installation ID that was used to generate it. Everything outside
the `github-api-client` library will just see this as an opaque type,
but bundling the `installation_id` will ensure that we can trace each
call back to the installation that made it. This functionality is
unused in the current commit, but is in preparation for future work
where we can use it to take advantage of DataDog's metric tags.

The `UserToken` is simply a type alias, but is useful for making it
explicit where App tokens are used, and where they aren't. The sole
user of the `UserToken` type will be phased out over time, as we roll
out Builder tokens more widely.

Third, the metrics code itself has been refactored to be more
modular. Previously, all the metrics we tracked, across all of
Builder, were centralized in the `builder-core` crate. Now, only the
core metric functionality is located there, and each crate that sends
metrics has its own `metrics.rs` module that defines the metrics
relevant to that crate.

Additionally, new traits have been introduced, such as `Metric` and
`CounterMetric` that can be used by whatever types you choose to
implement your metrics with. Currently, the metrics have been
implemented with enums, but nothing would stop you from building
metrics on other types as desired.

The actual sending of the metrics has been decoupled from the types as
well; now you simply pass a metric to a function from the
`builder-core::metrics` module. Right now, there's only `incr` for
incrementing counters, which is the only thing we use, but other
functions for decrementing, gauges, and other StatsD types could
easily be made. This way, it should be relatively easy to swap out the
underlying StatsD client, which I believe we will want to do shortly,
in order to take advantage of DataDog's tagging capabilities.

Finally, what used to be increments of the `github.api` metric, are
now broken out by endpoint. Now we have things like
`github.api.contents`, `github.api.check_team_membership`, etc. This
will provide a more granular view of our API activity. It does _not_
yet incorporate the installation ID into these metrics, though; that
will require tagging to work properly in DataDog.

### DIFF
--- a/components/builder-api/src/metrics.rs
+++ b/components/builder-api/src/metrics.rs
@@ -1,0 +1,33 @@
+// Copyright (c) 2018 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Centralized definition of all Builder API metrics that we
+//! wish to track.
+
+use std::borrow::Cow;
+use bldr_core::metrics;
+
+pub enum Counter {
+    GithubEvent
+}
+
+impl metrics::CounterMetric for Counter {}
+
+impl metrics::Metric for Counter {
+    fn id(&self) -> Cow<'static, str> {
+        match *self {
+            Counter::GithubEvent => "github.event".into(),
+        }
+    }
+}
--- a/components/builder-api/src/metrics.rs
+++ b/components/builder-api/src/metrics.rs
@@ -19,7 +19,7 @@ use std::borrow::Cow;
 use bldr_core::metrics;
 
 pub enum Counter {
-    GithubEvent
+    GitHubEvent
 }
 
 impl metrics::CounterMetric for Counter {}
@@ -27,7 +27,7 @@ impl metrics::CounterMetric for Counter {}
 impl metrics::Metric for Counter {
     fn id(&self) -> Cow<'static, str> {
         match *self {
-            Counter::GithubEvent => "github.event".into(),
+            Counter::GitHubEvent => "github.event".into(),
         }
     }
 }
--- a/components/builder-api/src/metrics.rs
+++ b/components/builder-api/src/metrics.rs
@@ -19,7 +19,7 @@ use std::borrow::Cow;
 use bldr_core::metrics;
 
 pub enum Counter {
-    GitHubEvent
+    GitHubEvent,
 }
 
 impl metrics::CounterMetric for Counter {}
--- a/components/builder-core/src/metrics.rs
+++ b/components/builder-core/src/metrics.rs
@@ -45,11 +45,11 @@ pub trait Metric {
 
 /// Marker trait for counter metrics (don't want to accidentally
 /// increment a gauge metric!)
-pub trait CounterMetric {}
+pub trait CounterMetric : Metric {}
 
 /// Increment the given metric by one
 pub fn incr<C>(counter: C)
-    where C: Metric + CounterMetric {
+    where C: CounterMetric {
     match sender().send((MetricType::Counter,
                          MetricOperation::Increment,
                          counter.id(),
--- a/components/builder-core/src/metrics.rs
+++ b/components/builder-core/src/metrics.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::borrow::Cow;
 use std::sync::{Once, ONCE_INIT};
 use std::sync::mpsc::{channel, sync_channel, Sender, Receiver, SyncSender};
 use std::thread;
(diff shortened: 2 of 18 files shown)


thesentinels:

Thank you for contributing!

components/builder-api/src/metrics.rs
```diff
@@ -0,0 +1,33 @@
...
+// limitations under the License.
+
+//! Centralized definition of all Builder API metrics that we
+//! wish to track.
+
+use std::borrow::Cow;
+use bldr_core::metrics;
+
+pub enum Counter {
+    GithubEvent
```
raskchanky: I hate to be the pedant, but technically this is `GitHub`, and is spelled that way in many other places across our code.

components/builder-core/src/metrics.rs
```diff
@@ -24,44 +25,57 @@ pub const APP_NAME: &'static str = "bldr";
...
-    SearchPackages,
+pub type ApiEndpoint = &'static str;
+pub type InstallationId = u32;
+
+// Public Interface
+////////////////////////////////////////////////////////////////////////
+
+/// Metric identifiers will usually be static `str`s, but some may
+/// need to be dynamically-generated `String`s. With a `Cow`, we can
+/// accept either.
```
raskchanky: That's pretty awesome.  I need to read up on `Cow`.

christophermaier: Page 300 of _Programming Rust_, Chapter 13: _Borrow and ToOwned at Work: The Humble Cow_

🐮

components/builder-api/src/metrics.rs
christophermaier: Consistency is good... I'll change it.

Thank you for your pedantry 😄

components/builder-core/src/metrics.rs
```diff
@@ -24,44 +25,57 @@ pub const APP_NAME: &'static str = "bldr";
...
-pub enum Gauge {
-    PackageCount,
+/// Marker trait for counter metrics (don't want to accidentally
+/// increment a gauge metric!)
+pub trait CounterMetric {}
+
+/// Increment the given metric by one
+pub fn incr<C>(counter: C)
+    where C: Metric + CounterMetric {
+    match sender().send((MetricType::Counter,
```
raskchanky: Where does `sender()` come from here?

components/builder-core/src/metrics.rs
```diff
@@ -24,44 +25,57 @@ pub const APP_NAME: &'static str = "bldr";
...
-// Supported metrics
-#[derive(Debug, Clone)]
-pub enum Gauge {
-    PackageCount,
+/// Marker trait for counter metrics (don't want to accidentally
+/// increment a gauge metric!)
+pub trait CounterMetric {}
+
+/// Increment the given metric by one
+pub fn incr<C>(counter: C)
```
chefsalim: Should incr be defined as a CounterMetric trait fn?

components/builder-core/src/metrics.rs
```diff
@@ -98,22 +112,14 @@ fn receive(rz: SyncSender<()>, rx: Receiver<MetricTuple>) {
...
             Some(ref mut cli) => {
                 match mtyp {
                     MetricType::Counter => {
                         match mop {
-                            MetricOperation::Increment => cli.incr(mid),
-                            MetricOperation::Decrement => cli.decr(mid),
-                            _ => error!("Unexpected metric operation: {:?}", mop),
-                        }
-                    }
-                    MetricType::Gauge => {
```
chefsalim: It would be nice to keep Gauge code in... at least as a sample or comment on how to add it later.

components/builder-core/src/metrics.rs
```diff
@@ -138,117 +144,3 @@ fn statsd_client() -> Option<Client> {
...
-
-impl Metric for Gauge {
-    fn id(&self) -> &'static str {
-        match *self {
-            Gauge::PackageCount => "package-count",
-        }
-    }
-}
-
-#[cfg(test)]
```
chefsalim: Tests would be nice to preserve as well..

components/builder-core/src/metrics.rs
```diff
@@ -24,44 +25,57 @@ pub const APP_NAME: &'static str = "bldr";
...
+    fn id(&self) -> MetricId;
 }
 
-// Supported metrics
-#[derive(Debug, Clone)]
-pub enum Gauge {
-    PackageCount,
+/// Marker trait for counter metrics (don't want to accidentally
+/// increment a gauge metric!)
+pub trait CounterMetric {}
```
chefsalim: Should CounterMetric also inherit from Counter?  I _think_ that could save having 2 impls later..?

components/github-api-client/src/metrics.rs
```diff
@@ -0,0 +1,47 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Centralized definition of all Github API client metrics that we
+//! wish to track.
+
+use std::borrow::Cow;
+use builder_core::metrics;
+
+pub type Endpoint = &'static str;
```
chefsalim: As a design choice the original code decided _not_ to allow names to come in as arbitrary strings.  Your use case makes sense here... we could think of having enums as well to have a bit more control over the strings.

chefsalim:
@christophermaier Overall this is an excellent refactor!! Just have a few thoughts to consider in the comments.

components/builder-core/src/metrics.rs
christophermaier: There is no core `Counter`... that's just what I was conventionally calling the crate-scoped metric definitions. This trait was just added to prevent passing a future `Gauge` metric to an `incr` function, etc.

components/builder-core/src/metrics.rs
christophermaier: Possibly? I didn't do it here to ensure that all the implementation was concentrated here. The idea is that you declare minimal types for your own metrics, and then pass those as input to one core set of functions. 

We could probably achieve the same thing with default implementations for `incr` and `decr` in `CounterMetric`.  I kinda wanted to see how this way played out for a little while. Also, I found a [dogstatsd](https://github.com/mcasper/dogstatsd-rs) crate that I was thinking about trying out next to get tagging, so I will probably be back in here soon anyway.

components/builder-core/src/metrics.rs
christophermaier: Little further down... it was one of the original pieces of code in this module I kept.

components/builder-core/src/metrics.rs
christophermaier: Re-implementing it won't be too difficult, we've got Git history to help us, and we weren't using it anywhere anyway, so I opted to remove it as dead code.

components/builder-core/src/metrics.rs
christophermaier: None of these tests actually did anything meaningful, and most were ignored. I could try to add meaningful ones, but it may take a while, and would likely end up just testing the statsd client we're using.

I suppose implementing things like `incr` in the `CounterMetric` trait could be useful for that, since we'd be able to more easily test the behavioral contract.

components/builder-core/src/metrics.rs
chefsalim: Right, the minimal type is goodness.. my thinking though is, if we defined incr as a default fn on the CounterMetric trait, then you'd never need to know about the global incr function at all... you'd just declare your Counter enum, impl the id function, and just call incr on it.

components/builder-core/src/metrics.rs
chefsalim: Yeah they were the barest minimum I could think of.. we can add some in the next pass.

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/builder/builds/350897486).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/builder/builds/350897486).

components/github-api-client/src/client.rs
```diff
@@ -27,10 +28,52 @@ use serde_json;
...
+    /// Retrieve the actual token content for use in HTTP calls.
+    pub fn inner_token(&self) -> &str {
+        self.inner_token.as_ref()
+    }
+}
+
+// Temporary type... will go away soon once we transition away from
+// personal Github tokens. Useful for expressing expectations through
+// explicitly typing, though.
+pub type UserToken = TokenString;
```
fnichol: This'll work great, however if it was a [newtype](https://rustbyexample.com/generics/new_types.html)-style struct, simply deleting this type later would show every place `UserToken` was used. I'm not 100% sure either way, but I believe that by type aliasing `UserToken` to `String` (ultimately), that the compiler swaps the alias name for the underlying type (as well as exposes all methods of the underlying type via the alias). Maybe someone will check me on this--it's one reason I use the "newtype" pattern so much.

christophermaier: Ah, yeah... that's probably a good tweak to make. In the one callsite we have, I did use `(token as UserToken)`, so we'll still get a compiler error when it gets removed, but that behavior is up to the programmer to remember 😄 

Good call.

PR #305, opened 2018-03-15
Remove generated code from git

elliott-davis:
I am removing this after realizing that we are all checking in different version of code generated by different versions of protoc. After this change, the build will be the single source of truth for all protocol changes. Also, I have removed the need to run `cargo build --features protocols` as it now runs automatically.

Closes: https://github.com/habitat-sh/habitat/issues/4290


thesentinels:

Thank you for contributing!

elliott-davis:
This is broken. Turns out we all have protofuf v3 installed and the hab package is still on v2. I'll take the time tomorrow to upgrade to 3.

baumanj:
Does this resolve https://github.com/habitat-sh/habitat/issues/4290, @elliott-davis? If so, go ahead and self assign for your much deserved glory.

components/libbuild-protocols.rs
```diff
@@ -48,7 +27,7 @@ mod protocols {
                 continue;
             }
             if file.metadata().unwrap().is_file() {
-                files.push(file.path());
+                files.push(file.path().to_string_lossy().into_owned());
```
baumanj: Seem like something has gone horribly wrong here if we somehow have invalid unicode in `file.path()`. I'd be more comfortable with
```rust
                  files.push(file.path().to_str().unwrap().into());
```

elliott-davis:
@baumanj it does resolve that issue! I think it would be good to discuss the implications of this here or in that issue. Personally, I don't see a problem with the build server generating these protobuf files. It should uncover problems with dependencies changing (which is how I found this in the first place) since the build will break immediately.

components/builder-api/Cargo.toml
```diff
@@ -12,11 +12,11 @@ path = "src/main.rs"
 doc = false
 
 [dependencies]
-clippy = {version = "*", optional = true}
-constant_time_eq = "*"
 base64 = "*"
 bodyparser = "*"
 builder-http-gateway = { path = "../builder-http-gateway" }
+clippy = {version = "*", optional = true}
+constant_time_eq = "*"
```
raskchanky: Thank you.

components/libbuild-protocols.rs
```diff
@@ -1,45 +1,24 @@
...
+        let protocols = protocol_files();
+        protoc_rust::run(protoc_rust::Args {
+            out_dir: "src/message",
+            input: protocols
+                .iter()
+                .map(AsRef::as_ref)
+                .collect::<Vec<&str>>()
+                .as_slice(),
+            includes: &["protocols"],
+        }).expect("protoc");
```
raskchanky: Perhaps the message that `expect` emits could be a little more descriptive?

raskchanky:
So, just to make sure I understand:

* Currently (w/o this PR) a developer runs `cargo build --features protocols` and whatever version of `protoc` said developer has on their workstation reads in the `*.proto` files and spits out a bunch of Rust code in the `src/message` subdir, which then gets compiled into the rest of our code.  The files inside `src/message` are checked into git because w/o their presence, everything would break.
* With this PR, there's no longer a need to run `cargo build --features protocols` because the protocols get re-compiled automatically every time `cargo build` is run.  `*.proto` files are still read in and Rust code is still emitted to the `src/message` subdir, just like before, with the main difference being that code is now git ignored and we use a library to generate it instead of a `protoc` binary.

Do I have that correct?

If I do, then this seems like a great change.  One downside I see off the cuff is an increased compile time for the `builder-protocols` crate, since it has to re-compile all the proto definitions every time, instead of only when they change.  Honestly, though, this crate doesn't change frequently and the increased convenience (not to mention correctness from not having to deal with different versions of `protoc`) feels worth it.  And, if someone really felt like the speed bump was onerous, you could probably implement some crude make-like logic in `build.rs` to check file timestamps and conditionally compile based on that.

elliott-davis:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/builder/builds/357423159).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/builder/builds/357423159).

chefsalim:
One thing that we're missing (that may be orthogonal), is how to enable a simple "cargo check" outside of the hab build?  I think we should continue to support that..

raskchanky:
@chefsalim I don't think I understand.  What's preventing `cargo check` from working with this as is?

raskchanky:
Kinda weird, but it looks like thesentinels didn't merge this.  I'll merge manually.

chefsalim:
@raskchanky there are no instructions or automation on how to enable a protocols build outside of the studio build - it's not this PR that's the cause, but in general we've lost that capability...

elliott-davis:
@chefsalim `cargo check` now automatically generates the proto messages :tada: 

As far as running cargo inside the dev studio, there is still some work to be done.

raskchanky:
image

@chefsalim Maybe I'm missing something here?  Seems fine to me.

raskchanky:
image

Full `cargo build` seems to work too.

raskchanky:
Note that both screenshots above were done outside the studio @chefsalim

chefsalim:
Right, but if you spin up a new VM, it's not going to work - you're not going to have any protobuf libraries installed..

raskchanky:
I don't think this PR removed any protobuf libraries, though, did it?  I don't see that in the diff.  So unless I'm misunderstanding, that would've also been a problem before this PR got merged?

chefsalim:
Yes, that's correct - it's an orthogonal issue, this PR is not the cause

chefsalim:
Opened new issue to track the above discussion https://github.com/habitat-sh/builder/issues/344

PR #325, opened 2018-03-21
Minor config cleanup for builder worker

chefsalim:
Minor tweak to remove un-needed configs from worker


chefsalim:
@elliott-davis the dev env config.sh was already pruned so we're good there

PR #336, opened 2018-03-27
Make `sed -i` work on both linux and Mac OS

baumanj:
Because the Mac OS version of sed's -i option requires an argument but linux's
does not, the way to get inline ovewrite behavior on both is to supply a zero-
length argument with no spaces after the -i.

### DIFF
--- a/terraform/instances.tf
+++ b/terraform/instances.tf
@@ -58,8 +58,8 @@ resource "null_resource" "api_provision" {
   provisioner "remote-exec" {
     inline = [
       "DD_INSTALL_ONLY=true DD_API_KEY=${var.datadog_api_key} /bin/bash -c \"$(curl -L https://raw.githubusercontent.com/DataDog/dd-agent/master/packaging/datadog-agent/source/install_agent.sh)\"",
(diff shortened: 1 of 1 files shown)


elliott-davis:
Do we need to make this change considering we'll never run terraform on a mac?

raskchanky:
Just to clarify, since @elliott-davis's comment confused me initially - the code in question is inside remote provisioners, which will always run on Linux (we don't run anything on mac servers).

baumanj:
I agree that we don't *need* to make this change if we never run terraform on a Mac, but I wasn't 100% sure that was the case. Also, since there doesn't seem to be any downside, I thought it was best to use the form which definitely works in both environments.

I was taking the approach of "if you find a problem in one place, fix _all_ the similar forms". This has the benefit of reducing the work if we should makes this error again and someone corrects it, then does a search for all the places where there's `"sed -i "`, this won't come up and force them to think about this again.

If we think there's a risk to making this change, I'm fine abandoning it, but if it's not totally safe, that implies there's something else going on here we don't understand.

elliott-davis:
I'm personally :-1: on this change since we know it works as-is and will never get run on a mac.

PR #338, opened 2018-03-27
Remove docker pinnings and add sbin to worker path

elliott-davis:
Yes, this will break some of the isolation in the worker environment but the alternative is building app armor as a hab package.

### DIFF
--- a/components/builder-worker/src/runner/docker.rs
+++ b/components/builder-worker/src/runner/docker.rs
@@ -12,9 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::env as std_env;
 use std::fs::{self, File};
 use std::io::{self, Write};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::process::{Child, Command, ExitStatus, Stdio};
 
 use hab_core::env;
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/builder-worker/src/runner/docker.rs
```diff
@@ -165,9 +166,15 @@ impl<'a> DockerExporter<'a> {
...
-        );
+        // TED: feels bad but we need to add sbin to the path
+        // so Dockerd can get to apparmor_parser
+        let env_paths = [
+            &*DOCKERD_PROGRAM.parent().expect(
+                "Dockerd parent directory exists",
+            ),
+            &Path::new("/sbin"),
+        ];
+        let env_path = join_paths(env_paths.iter()).unwrap().to_os_string();
```
fnichol: When it comes to functions from the stdlib (not methods on a struct), we sometimes keep the module name around so it's a little clearer where the function is coming from. For example:

```
use std::env;

// ...

env::join_path(env_paths.iter()).unwrap().to_os_string();
```

That doesn't change anything in the impl., but I was thrown for a second, not realizing there was a `join_paths()` function in the stdlib.

fnichol: I that I left us an interesting return type of `io::Result<Child>` meaning that calling `?` or `try!()` won't work in this method. It might be worth considering swaping that `.unwrap()` for an `.expect("Cannot join PATH elements for dockerd spawn") or equivalent call--this would give us a bit more information if the worker thread panicked here which is pretty unlikely as we control all the path elements in that function call.

elliott-davis: I actually had it as `env::join_paths` but got a collision with `hcore_env` and changed it :rofl:

Do you think it would be best to alias `std::env as std_env`?

fnichol: Oh right! That would work!

PR #340, opened 2018-03-28
Support optional on-prem analytics

cnunciato:
This change adds a new configuration section to allow on-prem customers to share usage data with us via Heap Analytics.

### DIFF
--- a/components/builder-web/app/actions/users.ts
+++ b/components/builder-web/app/actions/users.ts
@@ -16,6 +16,7 @@ import { authenticate, removeSession, loadOAuthProvider, loadBldrSessionState, l
 import { addNotification, SUCCESS, DANGER } from './notifications';
 import { BuilderApiClient } from '../client/builder-api';
 import { Browser } from '../browser';
+import config from '../config';
 
 export const CLEAR_ACCESS_TOKENS = 'CLEAR_ACCESS_TOKENS';
 export const CLEAR_NEW_ACCESS_TOKEN = 'CLEAR_NEW_ACCESS_TOKEN';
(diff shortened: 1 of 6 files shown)


components/builder-web/index.html
```diff
@@ -18,8 +24,18 @@
     <link id="hab-css" rel="stylesheet" href="/assets/app.css">
     <link rel="search" href="/opensearch.xml" type="application/opensearchdescription+xml" title="habitat.sh">
     <base href="/">
+
+    <!-- Heap -->
```
ryankeairns: @cnunciato I think the reason we're seeing the data appear in two places (in Heap) is that we're sending it directly to Heap with this snippet, and then again to Heap via Segment using the snippet below this. That would explain why we're seeing the company.id and company.name on the Production project in Heap, since where sending those two bits of data using a segment.identity call where Segment is using our Production Heap project ID.

Perhaps the solution here would be to use a different Segment snippet (Source) that points to the Development project. Let's look at Segment together when you have a few minutes.

ryankeairns: I did some poking around on Segment... I think what we need to do is create a new Source on Segment (localhost/local habitat dev) and add a new Heap destination to that Source (that uses the Heap 'Development' environment project ID).

Then we can send the identity info straight to heap using these Segment calls (that translate to the Heap `identity` and `track` calls --> https://segment.com/docs/destinations/heap/

components/builder-web/app/actions/users.ts
```diff
@@ -180,9 +181,15 @@ export function setDeletingAccessToken(payload) {
 
 function notifySegment(data: any) {
   const segment = window['analytics'];
-
   if (segment && typeof segment.identify === 'function') {
-    segment.identify(data.id, { email: data.email, name: data.name });
+    segment.identify(data.id, {
```
ryankeairns: Based upon my prior comment, I suspect this might become `analytics.identify` when it's all said and done. See these Segment docs --> https://segment.com/docs/destinations/heap/#identify

cnunciato:
After some experimentation with @ryankeairns, it looks like we need to remove the Heap script entirely and just use Segment, and we should make the Segment "write key" configurable instead, since that's what'll be configured on the receiving end to pass data through to Heap. Making these changes now.

ryankeairns:
tenor-122660932

PR #349, opened 2018-03-28
Check for jobsrv_enabled when syncing packages

chefsalim:
This change adds a jobsrv_enabled config value to sessionsrv, so that we don't attempt to sync package metadata with the jobsrv if it is not present.

tenor-96440966


thesentinels:

Thank you for contributing!

PR #356, opened 2018-03-29
Shorten filepath to get around containerd file path char limit

elliott-davis:
When the `--exec-root` file path is longer than 67 characters
containerd fails to start becuase it can't create it's gRPC sockets.

### DIFF
--- a/components/builder-worker/src/runner/docker.rs
+++ b/components/builder-worker/src/runner/docker.rs
@@ -192,12 +192,14 @@ impl<'a> DockerExporter<'a> {
         }
         cmd.arg("-H");
         cmd.arg(self.dockerd_sock());
+        // TED: Containerd has a file path limit of 67 characters and we were at 70 for most of ours.
+        // Shortning up the directories here to give us some breathing room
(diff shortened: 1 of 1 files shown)


components/builder-worker/src/runner/docker.rs
```diff
@@ -192,12 +192,14 @@ impl<'a> DockerExporter<'a> {
...
+        // Shortning up the directories here to give us some breathing room
         cmd.arg("--pidfile");
-        cmd.arg(root.join("var/run/docker.pid"));
+        cmd.arg(root.join("v/r/docker.pid"));
         cmd.arg("--data-root");
-        cmd.arg(root.join("var/lib/docker"));
+        cmd.arg(root.join("v/l/docker"));
         cmd.arg("--exec-root");
-        cmd.arg(root.join("var/run/docker"));
+        cmd.arg(root.join("v/r/d"));
```
raskchanky: Do these paths actually exist?

elliott-davis: they get created by docker

PR #357, opened 2018-03-29
Update bldr job status functionality

chefsalim:
This is the services update for https://github.com/habitat-sh/habitat/pull/4841

Getting the schedule status is now much more lightweight. By default, we only query the last most recent 10 job groups when querying by origin. Also we only return detailed project information when needed instead of always querying and returning it to the caller.
 
tenor-2406262


thesentinels:

Thank you for contributing!

PR #359, opened 2018-03-31
Add gitlab support

chefsalim:
Add support for Gitlab auth. The changes are mainly to the Github client/provider, and are backward compatible. A new config is added to builder-api for the redirect URL.  The config is also added to habitat-env-sample - devs will need to update their habitat-env, re-enter the studio, and run start-builder again to get their dev config updated.

tenor-113124898


thesentinels:

Thank you for contributing!

cnunciato:
Nice! 

I think at least for the UI portion of this, I'd prefer we created a `GitLab` class and an entry in the `OAuthProviderType` enum, since we use those classes for two purposes -- one to explicitly identify the params required for each type of service, second for the icon and labeling you see on the Sign In and Profile views.

To save you the hassle, I've put together a quick commit that I can push onto your branch if you like -- just say the word. 😄

chefsalim:
@cnunciato That is indeed a good idea and what we'll get to soon. For now, I explicitly don't want to add a new provider for the UI because of the way we are driving both front and and back end configs with the same environment variables in the setup scripts.  I think it's ok for the UI to say GitHub for now. We'll be doing some more cleanup here shortly, and that will be one of the things that will be covered in the next refactor.

ryankeairns:
+1 what @cnunciato said. I personally don't think it's'ok for the UI to say GitHub and letting these small things pass is a habit that we don't want to develop. Even if we hope to refactor soon, leaving this type of thing as-is degrades confidence and quality - which is something we actively trying to improve upon here at Chef.

chefsalim:
@ryankeairns I get your point, but we should be able to README this.  We're presenting this as a MVP, and I'd like to also avoid throwaway work which may be entailed.  Having a broken system is probably worse than a working system with a single UI imperfection while we kick the tires working closely with the customer(s).

ryankeairns:
@chefsalim  I personally don't think this particular situation is acceptable even as in MVP since we are sharing this with customers. At a minimum, I would prefer that we change the button text to simply read 'Sign In' and hide the github logo for now. Thanks!

components/github-api-client/src/client.rs
```diff
@@ -248,14 +252,32 @@ impl OAuthClient for GitHubClient {
...
-        let user: User = serde_json::from_str(&body).map_err(
+        let user: serde_json::Value = serde_json::from_str(&body).map_err(
             OAuthError::Serialization,
         )?;
 
+        // Gitlab API has username instead of login
+        let username = if user["login"].is_string() {
+            user["login"].to_string()
+        } else {
+            assert!(user["username"].is_string());
```
baumanj: What are the circumstances in which this `assert!` can fail? If I'm reading the code right, it's when the Github or Gitlab response contains neither `"login"` nor `"username"`. Is it possible they could change the format of their response? What do we gain by panicking over an error with a message?

chefsalim: I should probably add a comment here for a TBD to more cleanly split up the API handling between Github and Gitlab as we do the next round of auth refactor, and maybe go back to the structured deserialization.  I'll do that separately.

raskchanky: Yeah I think structured deserialization is the right long term plan.

PR #362, opened 2018-04-03
Strip quotes from serde strings

chefsalim:
The serde Value to_string function does not strip quotes from strings.  We need to do the as_... functions to get the quotation marks to get stripped.  This fixes a regression where the logged in user is not correctly recognized.

tenor-217342806


thesentinels:

Thank you for contributing!

PR #363, opened 2018-04-03
Update TF scripts to correctly deploy acceptance

elliott-davis:

### DIFF
--- a/terraform/instances.tf
+++ b/terraform/instances.tf
@@ -102,13 +102,14 @@ resource "null_resource" "api_provision" {
   provisioner "remote-exec" {
     inline = [
       "chmod +x /tmp/install_base_packages.sh",
+      "sudo /tmp/install_base_packages.sh",
       "sudo mv /home/ubuntu/hab-sup.service /etc/systemd/system/hab-sup.service",
       "sudo systemctl daemon-reload",
       "sudo systemctl start hab-sup",
       "sudo systemctl enable hab-sup",
-      "sudo hab svc load habitat/builder-api --group ${var.env} --bind router:builder-router.${var.env} --strategy at-once --url ${var.bldr_url} --channel ${var.release_channel}",
-      "sudo hab svc load habitat/builder-api-proxy --group ${var.env} --bind http:builder-api.${var.env} --strategy at-once --url ${var.bldr_url} --channel ${var.release_channel}",
-      "sudo hab svc load core/sumologic --group ${var.env} --strategy at-once --url ${var.bldr_url} --channel ${var.release_channel}",
+      "sudo hab svc load habitat/builder-api --force --group ${var.env} --bind router:builder-router.${var.env} --strategy at-once --url ${var.bldr_url} --channel ${var.release_channel}",
+      "sudo hab svc load habitat/builder-api-proxy --force --group ${var.env} --bind http:builder-api.${var.env} --strategy at-once --url ${var.bldr_url} --channel ${var.release_channel}",
+      "sudo hab svc load core/sumologic --force --group ${var.env} --strategy at-once --url ${var.bldr_url} --channel ${var.release_channel}",
     ]
   }
 }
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

terraform/instances.tf
```diff
@@ -102,13 +102,14 @@ resource "null_resource" "api_provision" {
   provisioner "remote-exec" {
     inline = [
       "chmod +x /tmp/install_base_packages.sh",
+      "sudo /tmp/install_base_packages.sh",
```
christophermaier: OMG

terraform/instances.tf
```diff
@@ -102,13 +102,14 @@ resource "null_resource" "api_provision" {
...
       "sudo systemctl daemon-reload",
       "sudo systemctl start hab-sup",
       "sudo systemctl enable hab-sup",
-      "sudo hab svc load habitat/builder-api --group ${var.env} --bind router:builder-router.${var.env} --strategy at-once --url ${var.bldr_url} --channel ${var.release_channel}",
-      "sudo hab svc load habitat/builder-api-proxy --group ${var.env} --bind http:builder-api.${var.env} --strategy at-once --url ${var.bldr_url} --channel ${var.release_channel}",
-      "sudo hab svc load core/sumologic --group ${var.env} --strategy at-once --url ${var.bldr_url} --channel ${var.release_channel}",
+      "sudo hab svc load habitat/builder-api --force --group ${var.env} --bind router:builder-router.${var.env} --strategy at-once --url ${var.bldr_url} --channel ${var.release_channel}",
+      "sudo hab svc load habitat/builder-api-proxy --force --group ${var.env} --bind http:builder-api.${var.env} --strategy at-once --url ${var.bldr_url} --channel ${var.release_channel}",
+      "sudo hab svc load core/sumologic --force --group ${var.env} --strategy at-once --url ${var.bldr_url} --channel ${var.release_channel}",
```
christophermaier: Why `--force`? Presumably these commands are being run on fresh instances, right?

terraform/s3.tf
```diff
@@ -18,7 +18,7 @@ resource "aws_s3_bucket" "jobs" {
   region = "${var.aws_region}"
 
   lifecycle {
-    prevent_destroy = true
+    prevent_destroy = false
```
christophermaier: What was the reason for this change?

terraform/scripts/foundation.sh
```diff
@@ -5,10 +5,14 @@
 
 set -eux
 
-sudo mount /dev/xvdf /mnt
-echo '/dev/xvdf /hab     ext4   defaults 0 0' | sudo tee -a /etc/fstab
-sudo mkdir -p /mnt/hab
-sudo ln -s /mnt/hab /hab
+if mountpoint -q /mnt; then
+    echo "Already Mounted"
```
christophermaier: What's the situation where this would already be mounted?

terraform/www.tf
```diff
@@ -34,7 +34,7 @@ resource "aws_s3_bucket" "www" {
   acl    = "public-read"
 
   lifecycle {
-    prevent_destroy = true
+    prevent_destroy = false
```
christophermaier: ditto here.

terraform/instances.tf
elliott-davis: unless one fails, then you end up in fail hell

terraform/scripts/foundation.sh
elliott-davis: Same reason for the force above. If provisioning fails for some other reason, it will re-run the null resource and fail

terraform/instances.tf
christophermaier: Wouldn't you taint those resources and start fresh? `--force` would only come into play when there's already a spec file for that package present.

terraform/instances.tf
chefsalim: Wow - how was this ever working?

terraform/instances.tf
chefsalim: Yeah @christophermaier has a good point.. if we're in a bad state on initial provision, we'd want to re-create that instance until we could deploy clean.

terraform/scripts/foundation.sh
chefsalim: Yeah this feels like the wrong strategy as well...

terraform/instances.tf
elliott-davis: iirc Terraform should be idempotent and convergent so I think these changes make sense.

terraform/scripts/foundation.sh
elliott-davis: Same answer as above on force.

terraform/instances.tf
christophermaier: Best to follow the "Nuke it from orbit... it's the only way to be sure" school of infrastructure management here.

elliott-davis: Yeah that's interesting since we're using a null resource for provisioning. The null resource is tainted but it never taints the VM. I'll double back and see if we can change the resource type to.make this safer

chefsalim:
Closing this PR as I don't think we're going to accept it in its current form.  I'll open a new PR to fix up the install_base_packages.sh omission.

PR #370, opened 2018-04-11
Make nginx body size configurable

chefsalim:
We currently restrict the size of http client body to 1GB in the builder-api-proxy, with no way to make it configurable.  This change adds a config option for scenarios where we need to change the max size (for example, larger package upload sizes)


thesentinels:

Thank you for contributing!

PR #372, opened 2018-04-12
Adds S3 backend to builder depot

eeyun:


thesentinels:

Thank you for contributing!

eeyun:
I need to put this work on pause. If someone else feels the desire to carry this they can, otherwise I'll return to it after kubecon.

### Notes
This changes the artifact storage backend to either amazon S3 or an S3 compatible  API. This means that for on-premise builder clusters, we will need to deploy a minio service alongside the depot and API. There are plenty of memory optimizations that can be performed in this code and there are currently no scripted tests. This PR also has not yet been functionally tested. As such a review pass, optimization pass, and functional testing need to be performed before this is merged.

 
## Changes
After this PR the way artifacts are handled changes but hopefully not dramatically. A majority of the validation and safety check we perform local to the depot hard disk are still performed in this manner. We are now however treating that storage as a sort of short term proxy. On a PUT request, we write the package to disk in a temporary location. We perform all of our request and package validations (including determining whether or not the package already exists in the backend) and once these constraints are satisfied we upload to the backend. After we've validated the upload we remove the package from local storage.

On a GET request we perform the inverse operation. We pull the package from the backend and write it to a temporary location on disk, we perform necessary validations and we hand the package off to the client. Once we've transferred the package we remove it from local storage.

## Future State
- S3 has a hard limit of 5gb per single PUT. Rusoto supports incrementally uploading chunks of a file and realistically our upload function should use this API instead of the one we're using on the branch currently. We don't have any packages yet that are larger than 5gb but someday, it wouldn't surprise me.
- In an ideal situation we could theoretically simplify the package upload function in the server by directly connecting the client to s3 and streaming their request. This would require significant effort in rusoto's implementation of futures as well as a nontrivial refactor of the server's handling of package uploads and downloads.
- S3 has the option of tiered storage levels (Faster vs slower). If we were to add a concept of hot vs cold packages to the depot we could theoretically use varying storage backends per package based on some arbitrary qualifier of the package provided by a request.
- Ultimately if there was a significant refactor of the server code the abstractions used for handling of requests could likely be simplified. In it's current state this is the simplest implementation of a new backend.

eeyun:
Also it looks like travis isn't actually using latest rust. This PR follows `use` syntax from rust 1.25

components/builder-depot/src/server.rs
```diff
@@ -841,9 +857,21 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
...
         };
 
-    if origin_package_found && depot.archive(&ident, &target_from_artifact).is_some() {
-        return Ok(Response::with(status::Conflict));
-    };
+    // If an S3 backend is set in the config file
+    // check whether the package already exists in
+    // the configured bucket, otherwise validate on disk
+    if origin_package_found {
+        let s3_config = depot.config.s3.to_owned();
```
elliott-davis: I see this is used in a few places. Would it make sense to move the S3 connection into the global context that is passed around with the request handers?

components/builder-depot/src/server.rs
```diff
@@ -841,9 +857,21 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
...
+        let s3handler = s3::S3Handler::new(s3_config);
+        match s3handler.exists(&ident, &target_from_artifact) {
+            Ok(pot_hocket) => {
```
elliott-davis: A more descriptive name here might be helpful for future humans

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    // Helper function for programmatic creation of
+    // the s3 object key
+    pub fn s3_key<T>(&self, ident: &T, target: &PackageTarget) -> PathBuf
+    where
+        T: Identifiable,
+    {
+        Path::new(ident.origin())
+            .join(format!("{}", ident.name()))
```
elliott-davis: You might want to leverage this here: https://github.com/habitat-sh/core/blob/master/components/core/src/package/ident.rs#L93-L99

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,189 @@
...
+
+impl S3Handler {
+    // The S3 Handler struct contains all of the credential
+    // and target information that we should need to perfom
+    // any backend operations
+    pub fn new(config: S3Cfg) -> Self {
+        let region =
+            match config.endpoint {
+                Some(ref value) => {
+                    Region::Custom { name: "minio_s3".to_owned(), endpoint: value.to_string() }
```
chefsalim: Instead of hardcoding the region name here, it might be nicer to just use the aws_region config.  We could put some sane defaults for the minio config for local dev - eg, "minio_s3" and "http://localhost:9000"

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,189 @@
...
+        let mut request = ListObjectsRequest::default();
+        let key = self.s3_key(ident, target).to_string_lossy().into_owned();
+        request.bucket = self.bucket.to_owned();
+        match self.client.list_buckets().sync() {
+            Ok(bucket_list) => {
+                match bucket_list.buckets {
+                    Some(bkts) => {
+                        for bucket in bkts {
+                            if bucket.name.unwrap() == request.bucket {
+                                match self.client.list_objects(&request).sync() {
```
chefsalim: Some concerns here: 1) synchronous blocking operation;  2) large result set (we are retrieving list of all objects in our S3 bucket?); 3) according to the API, list_objects returns up to a max of 1000.  This doesn't seem feasible if the idea is to iterate through all objects we have stored looking for a key match..

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,189 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Pkg storage backend variant which uses S3 (or an API compatible clone) for
+//! hart storage.
+//!
+//! Has been tested against AWS S3.
+//!
+//! All packages are stored in a single bucket, using the fully qualified
+//! package ident as the key, prefixed with `pkgs`
```
chefsalim: Why is the "pkgs" key prefix needed / desired?

components/builder-depot/src/backend/s3.rs
chefsalim: We should consider key optimization instead of just sticking with the canonical package name - eg, start keys with the reversed release timestamp to avoid AWS partition hotspots. https://minops.com/blog/2015/01/aws-s3-performance-tuning/

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,189 @@
...
+        let mut request = PutObjectRequest::default();
+        request.key = key;
+        request.bucket = buck;
+        request.body = Some(object);
+        request.content_type = Some("application/x-tar".to_string());
+
+        // TODO (ian): A maximum single PUT to S3 is 5gb
+        // rusoto supports PUT in chunks.  This upload
+        // should chunk the upload request to stay <limit
+        match self.client.put_object(&request).sync() {
```
chefsalim: Definitely a concern that we won't have multipart, and we are doing a sync upload.. we should deeply instrument the S3 code with metrics and logging so we have a pretty good understanding of what the perf looks like under different loads.

components/builder-api/src/server/mod.rs
```diff
@@ -41,23 +41,24 @@ impl HttpGateway for ApiSrv {
     fn add_middleware(config: Arc<Self::Config>, chain: &mut iron::Chain) {
         chain.link(persistent::Read::<Self::Config>::both(config.clone()));
 
-        chain.link(persistent::Read::<OAuthCli>::both(OAuth2Client::new(
-            config.oauth.clone(),
-        )));
+        chain.link(persistent::Read::<OAuthCli>::both(
```
chefsalim: Need to rebase w/ rustfmt 0.38 formatting?

components/builder-depot/src/backend/s3.rs
eeyun: This is actually an incorrect comment, we don't store in a path prefixed with `pkgs`, will get this removed.

components/builder-depot/src/backend/s3.rs
eeyun: Yeah I'll have to think about this one. This part of the code hasn't been exercised yet. Right now nothing is actually hardcoded, the way it functions is you can either provide an "endpoint" value to the config or you can provide a "region". The "name" in this case is a bit of boilerplate from the rusoto crate, the endpoint is the actual configured value.

Right now we provide a default value to region and not endpoint. It should be trivial to flip those if we want to effectively default to local mode over S3.

components/builder-depot/src/backend/s3.rs
eeyun: Yes, the API list objects limit is definitely going to be a concern. The operations themselves are futures, so these could be performed in an async way. Will have to think about how to accomplish that.

components/builder-depot/src/backend/s3.rs
eeyun: We definitely _will_ have multi-part upload. It just wasn't implemented when I pushed the branch up so I wanted to make sure I had my note in here to get this switched over :)

components/builder-depot/src/backend/s3.rs
chefsalim: Cool, yeah that would be great if we flip over to local mode default.

components/builder-depot/src/backend/s3.rs
chefsalim: Well, maybe we need to think about whether we need/should have an existence check at all.  We essentially are already keeping a secondary index via the metadata in originsrv.  So we could assume that we have the package in S3 already if the originsrv metadata says we do (post-migration).  So this is only a sanity validation for upload that isn't really serving much need in the S3 world.

components/builder-depot/src/error.rs
```diff
@@ -163,3 +181,15 @@ impl From<hab_net::NetError> for Error {
         Error::NetError(err)
     }
 }
+
+impl From<rusoto_s3::ListBucketsError> for Error {
```
chefsalim: A while ago it was decided we'd stop doing the From's for errors and use .map_err instead.. so we could just remove these.

components/builder-api/src/server/mod.rs
eeyun: I think I'm on 0.9.0. I can pin back to 0.3.8 but is there a reason this part of the codebase is pinned back this far?

components/builder-depot/src/backend/s3.rs
eeyun: You might be correct about that. It's worth discussing. I think initially i put this validation in here because its the operation we were doing with locally stored artifacts. I guess my question would be, what about a theoretical edge case where the data exists in orignsrv but the package never actually successfully uploaded to s3?

components/builder-api/src/server/mod.rs
chefsalim: 0.3.8 is actually newer (go figure :)

components/builder-depot/src/backend/s3.rs
chefsalim: If we upload the package first, and only write to originsrv on a successful upload that should not happen.  We just need to be able to idempotently upload to S3.

components/builder-depot/src/error.rs
eeyun: Oh! I didn't know that. Will get these removed as well.

components/builder-api/src/server/mod.rs
eeyun: WHAT. Ugh, I'm all turned around. I'll get this updated then!

components/builder-depot/src/backend/s3.rs
eeyun: I'll read this and adjust accordingly.

eeyun: Ok, because of the way we're generating these keys i don't know if we actually need to change this. Right now we generate a pretty long path which effectively serve to group all of our packages into logical blobs of /origin/packagename/version long before we ever get to the hartfile. I think this optimization should be kept in mind in case we start to see any issues, but based on that blog post i think what we're doing is probably totally fine!

components/builder-depot/src/backend/s3.rs
eeyun: Ok this should be corrected now, minio is the default and this abstraction has been cleaned up a bit!

components/builder-depot/src/backend/s3.rs
eeyun: EoD note on this, I haven't touched this bit yet. Planning to fix this in the morning and theoretically remove this whole function.

components/builder-depot/src/backend/s3.rs
eeyun: Latest commit implements multi-part uploads, still a raw commit, I didn't get as far as I would have liked but all the functionality should now be in for this the code itself just needs to be cleaned up and potentially optimized.

components/builder-api/src/server/mod.rs
eeyun: This is corrected, I got my workstation all updated XD

components/builder-depot/src/error.rs
eeyun: I haven't yet cleaned these out, will do this in the morning as well.

components/builder-depot/src/backend/s3.rs
chefsalim: Apparently it's the first 6 or so characters at the beginning of the key that really matters for the partitioning.. we're probably fine at the current TPS but if we do hit a limit later it's going to be much harder to switch the key scheme.

eeyun: I can switch it if you think its worth it, it'd be better to do this before we migrate, but the blog post seemed to indicate that creating logical groupings with a pathlike prefix to the key would also serve to group objects together on a partition? Unless I'm misunderstanding what they meant by that, that's basically what we're doing now with our origin.

elliott-davis: I think since most of our operations are going to be GET request we should be fine fronting our bucket with cloudfront https://docs.aws.amazon.com/AmazonS3/latest/dev/request-rate-perf-considerations.html

eeyun: Yeah the section in there on grouping is what i was referring to. I think we're good on this.

components/builder-depot/src/server.rs
```diff
@@ -1241,6 +1262,19 @@ fn download_package(req: &mut Request) -> IronResult<Response> {
 
     match route_message::<OriginPackageGet, OriginPackage>(req, &ident_req) {
         Ok(package) => {
+            let temp_path = depot.archive_path(package.get_ident(), &agent_target);
```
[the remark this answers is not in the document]
eeyun: This is possible however in doing so we no longer shovel around a PathBuf and instead shovel around an open File, that means further refactor around how we write archive to disk on download which then affects how we write the archive to disk on upload. I'd be ok making this change but I'd need further input and validation before going that path.

components/builder-api/src/config.rs
```diff
@@ -171,6 +173,12 @@ mod tests {
         client_id = "0c2f738a7d0bd300de10"
         client_secret = "438223113eeb6e7edf2d2f91a232b72de72b9bdf"
 
+        [s3]
```
chefsalim: Another thought on how to structure the config - it might be cleaner to call out the backend provider specifically. It lets us provide things like endpoint for minio, vs a region for AWS more naturally. For example:

[S3]
backend = "minio"
access_key = "dev"
secret_key = "secret"
endpoint = "http://localhost:9000"
bucket = "dev"

[S3]
backend = "aws"
access_key = "AWS_KEY"
secret_key = "AWS_SECRET"
region = "AWS_REGION"
bucket = "AWS_BUCKET"

Also as discussed the default should be for minio.

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,322 @@
...
+    //            debug!("{:?}", e);
+    //            Err(Error::BadBucket(e))
+    //        }
+    //    }
+    //}
+
+    // This function checks whether or not the
+    // configured bucket exists in the configured
+    // backend.
+    fn exists(&self) -> Result<bool> {
```
chefsalim: Might as well name it 'bucket_exists' ..

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,322 @@
...
+                return Err(Error::ListBuckets(e));
+            }
+        }
+    }
+
+    pub fn create_bucket(&self) -> Result<()> {
+        let mut request = CreateBucketRequest::default();
+        request.bucket = self.bucket.clone();
+
+        if self.exists().unwrap() {
```
chefsalim: Might be nicer to have the exists check outside of create bucket.. if !bucket_exists() { create_bucket() }.. :)

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    pub fn upload(
+        &self,
+        hart: &PathBuf,
+        ident: &PackageIdent,
+        target: &PackageTarget,
+    ) -> Result<()> {
+        let key = self.s3_key(ident, target).to_string_lossy().into_owned();
+        let file = File::open(hart)?;
+        let size = file.metadata().unwrap().len() as usize;
```
chefsalim: Check result instead of unwrap?

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,322 @@
...
+        let min_limit: usize = 49152 * 1024;
```
chefsalim: Might be worth making it a constant w/ a comment on why 49152?

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,322 @@
...
+        if size < min_limit {
+            self.single_upload(key, file)
+        } else {
+            self.multipart_upload(key, file)
+        }
```
chefsalim: One other thing to consider is adding timing data for upload and downloads, similiar to how the graph operations are instrumented - https://github.com/habitat-sh/builder/blob/master/components/builder-jobsrv/src/server/mod.rs#L153

Also, it would be good to add metrics so we can also track them on a dashboard like we do github calls.

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    pub fn download(
```
chefsalim: Can/should we have multi-part downloads as well?

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,322 @@
...
+
+        let payload = self.client.get_object(&request).sync();
+        let body = match payload {
+            Ok(response) => response.body,
+            Err(e) => {
+                return Err(Error::PackageDownload(e));
+            }
+        };
+
+        let file = body.expect("Downloaded pkg archive empty!").concat2();
```
chefsalim: Check result instead of expect..?

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,322 @@
...
+        let _result = DepotUtil::write_archive(&loc, file.wait().unwrap());
```
chefsalim: Unwrap safe? Also check result of the write_archive?

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,322 @@
...
+        Ok(())
+    }
+
+    fn single_upload(&self, key: String, hart: File) -> Result<()> {
+        let mut reader = BufReader::new(hart);
+        let mut object: Vec<u8> = Vec::new();
+        let bucket = self.bucket.clone();
+        let _result = reader.read_to_end(&mut object);
```
chefsalim: Check result?

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,322 @@
...
+        let mut mprequest = CreateMultipartUploadRequest::default();
+        mprequest.key = key.clone();
+        mprequest.bucket = self.bucket.clone();
+
+        match self.client.create_multipart_upload(&mprequest).sync() {
+            Ok(output) => {
+                let mut request = UploadPartRequest::default();
+                request.key = key.clone();
+                request.bucket = self.bucket.clone();
+                request.upload_id = output.upload_id.clone().unwrap();
```
chefsalim: Unwrap ok?

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,322 @@
...
+
+                let cap: usize = 49152 * 1024;
```
chefsalim: Consider constant as per above..

components/builder-depot/src/config.rs
```diff
@@ -94,6 +99,35 @@ impl GatewayCfg for Config {
...
+    pub secret_key: String,
+    pub bucket_name: String,
+    pub backend: S3Backend,
+    pub endpoint: String,
+}
+
+impl Default for S3Cfg {
+    fn default() -> Self {
+        S3Cfg {
+            key_id: String::from(""),
```
chefsalim: Might be good to match all the defaults we put in the default.toml here as well..

components/builder-depot/src/error.rs
```diff
@@ -54,8 +64,12 @@ impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let msg = match *self {
             Error::BadPort(ref e) => format!("{} is an invalid port. Valid range 1-65535.", e),
+            Error::CaughtPanic(ref msg, ref source) => {
```
chefsalim: Seems ominous.. is this actually a thing?

components/builder-depot/src/lib.rs
```diff
@@ -114,6 +120,13 @@ impl DepotUtil {
             ))
     }
 
+    fn write_archive(filename: &PathBuf, body: Vec<u8>) -> Result<PackageArchive> {
+        let file = File::create(&filename)?;
+        let mut write = BufWriter::new(file);
+        let _payload = write.write_all(&body);
```
chefsalim: Check result and return appropriately?

components/builder-depot/src/server.rs
```diff
@@ -841,9 +849,11 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
             }
         };
 
-    if origin_package_found && depot.archive(&ident, &target_from_artifact).is_some() {
+    // Using the S3 backend if package exists in originsrv
+    // we make the assumption that the package exists in s3
+    if origin_package_found {
```
chefsalim: You can return directly from inside the match, and get rid of the origin_package_found variable entirely..

components/builder-depot/src/server.rs
```diff
@@ -869,6 +879,13 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
 
     let filename = depot.archive_path(&ident, &target_from_artifact);
 
+    let trident = ident.to_owned().into();
```
chefsalim: Nitpick - consider rename to something like temp_ident instead of 'trident' ? :)

components/builder-depot/src/server.rs
```diff
@@ -869,6 +879,13 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
...
+    if s3handler
+        .upload(&temp_path, &trident, &target_from_artifact)
+        .is_err()
+    {
+        return Ok(Response::with(status::InternalServerError));
```
chefsalim: Add an info! or warn! output here?

components/builder-depot/src/server.rs
```diff
@@ -1241,6 +1260,15 @@ fn download_package(req: &mut Request) -> IronResult<Response> {
 
     match route_message::<OriginPackageGet, OriginPackage>(req, &ident_req) {
         Ok(package) => {
+            let temp_path = depot.archive_path(package.get_ident(), &agent_target);
+            let trident = ident.to_owned().into();
```
chefsalim: Naming nitpick.. use temp_ident instead of trident?

components/builder-api/src/config.rs
eeyun: yes indeed, perhaps I have this locally but I thought I'd actually already done this! I will validated and get this in. I ran into this myself last week!

components/builder-depot/src/backend/s3.rs
eeyun: excellent idea!

components/builder-depot/src/backend/s3.rs
eeyun: Absolutely.

components/builder-depot/src/backend/s3.rs
eeyun: We do, the download function simply checks the size of the object being uploaded before calling the appropriate function either single GET or multi-part!

components/builder-depot/src/error.rs
eeyun: Nope I don't believe so, I think I must have added this in an early iteration and forgotten to remove it.

components/builder-api/src/config.rs
eeyun: Yeah it appears I hadn't pushed this up! Also I've duplicated these settings into the default.toml

components/builder-depot/src/backend/s3.rs
eeyun: Perhaps. For now this code path can't actually get executed. I left it in for once we decide whether or not we want on-prem users (or ourselves for that matter) to have to use an S3 account with elevated privileges. Using this functionality would require whatever service account to have both PUT, GET, and CREATE permissions in s3. Significantly smaller issue for Minio but still worth a discussion before any further work is performed here.

components/builder-depot/src/backend/s3.rs
eeyun: This shouldn't be necessary, all files have metadata and I believe all rust File types have access to that metadata.

components/builder-depot/src/backend/s3.rs
eeyun: Absolutely. Do we want metrics on upload requests, successful uploads, and failed uploads?

components/builder-depot/src/backend/s3.rs
eeyun: This isn't actually a result, the result was pulled out before this expect, this is actually an expectation of an Option. Is using a match here still what you'd like me to do?

components/builder-depot/src/backend/s3.rs
eeyun: This isn't an actual result type, I think I'll change this variable name. This is just being used to execute the `.read_to_end()`

components/builder-depot/src/backend/s3.rs
eeyun: Yep this should be good, the rusoto library has this struct parameter listed with the Option type but AWS multipart upload requests should always return an upload id in the case of an OK response. This _should_ be ok I believe.

components/builder-depot/src/config.rs
eeyun: Yep I had this locally as well :)

components/builder-depot/src/server.rs
eeyun: Oh duh, excellent catch!

components/builder-depot/src/backend/s3.rs
chefsalim: That's probably fine.. we occasionally do get errors pop up with file handles etc though.. I saw other instances of metadata checking result..

components/builder-depot/src/backend/s3.rs
chefsalim: Ah, great!

components/builder-api/src/config.rs
chefsalim: 👍

components/builder-depot/src/backend/s3.rs
chefsalim: Ah ok, seems safe as we're never pushing empty packages.. just don't know what to expect with S3 APIs :)

components/builder-depot/src/backend/s3.rs
chefsalim: OK (may be good to also add an 'unwrap safe' comment at the end of the line...)

components/builder-depot/src/error.rs
```diff
@@ -25,26 +25,35 @@ use hab_net;
...
 #[derive(Debug)]
 pub enum Error {
     BadPort(String),
     ChannelAlreadyExists(String),
     ChannelDoesNotExist(String),
+    CreateBucketError(rusoto_s3::CreateBucketError),
     DepotClientError(depot_client::Error),
     HabitatCore(hab_core::Error),
     HabitatNet(hab_net::error::LibError),
-    NetError(hab_net::NetError),
```
eeyun: Not sure where that went... I don't recall removing this.

eeyun: Perhaps I removed this because it wasn't being used. Im not seeing any errors or warnings in compilation about it so I'm gonna go with it.

.studiorc
```diff
@@ -140,10 +140,14 @@ _start-builder() {
     start-builder originsrv
     start-builder sessionsrv
     start-builder worker
+    start-builder minio
     sleep 2
     generate_bldr_keys
     upload_github_keys
     echo "Builder Started: Navigate to http://localhost/#/pkgs/core to access the web UI."
+    echo "Minio login available at http://localhost:9000"
+    echo "Username: AWSKEYIDIBELIEVE"
```
chefsalim: Minor - would be nice to default the key/secret to something simple and easy to remember/use, like "dev" / "dev".

components/builder-api/habitat/config/config.toml
```diff
@@ -27,3 +27,8 @@ path = "{{pkg.svc_data_path}}"
 log_dir = "{{pkg.svc_var_path}}"
 key_dir = "{{pkg.svc_files_path}}"
 {{toToml cfg.depot}}
+
+{{#if cfg.s3 ~}}
```
chefsalim: Do we even need the #if cfg here?

components/builder-api/habitat/default.toml
```diff
@@ -27,3 +27,10 @@ jobsrv_enabled          = true
 [segment]
 url       = "https://api.segment.io"
 write_key = ""
+
+[s3]
+backend = "Minio"
+key_id = "AWSKEYIDIBELIEVE"
+secret_key = "wellIwanttobelieveanywaysbutwhoknowssincethisissecret"
+endpoint = "http://localhost:9000"
+bucket_name = "habitat-builder-dev-artifact-store"
```
chefsalim: Minor - would be nice to change to "habitat-builder-local-artifact-store" (then we only need to change the key/secret for on-prem installations)

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,347 @@
...
+    client: S3Client<StaticProvider, RequestDispatcher>,
+    bucket: String,
+}
+
+impl S3Handler {
+    // The S3 Handler struct contains all of the credential
+    // and target information that we should need to perfom
+    // any backend operations
+    pub fn new(config: S3Cfg) -> Self {
+        let region = match config.backend {
```
chefsalim: Nice..

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,347 @@
...
+                ident.origin(),
+                ident.name(),
+                ident.version().unwrap(),
+                ident.release().unwrap(),
+                target.architecture,
+                target.platform
+            ))
+    }
+
+    // Helper function to validate whether or not a package
```
chefsalim: Remove the commented function?

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,347 @@
...
+    fn bucket_exists(&self) -> Result<bool> {
+        let artifactbucket = self.bucket.to_owned();
+        match self.client.list_buckets().sync() {
+            Ok(bucket_list) => match bucket_list.buckets {
+                Some(buckets) => {
+                    for bucket in buckets {
+                        if bucket.name.unwrap() == artifactbucket {
+                            return Ok(true);
+                        } else {
+                            return Ok(false);
```
chefsalim: This logic does not seem right.. remove the else?

chefsalim: You could also maybe simplify to something like:
return Ok(buckets.iter().any(|&b| b.name.unwrap() == artifactbucket)))

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,347 @@
...
+                return Err(Error::ListBuckets(e));
+            }
+        }
+    }
+
+    pub fn create_bucket(&self) -> Result<()> {
+        let mut request = CreateBucketRequest::default();
+        request.bucket = self.bucket.clone();
+
+        if self.bucket_exists().unwrap() {
```
chefsalim: I don't know if I'd do an unwrap here... if you're confident there's no chance of error, then just change the bucket_exists function to only return a bool, not a result.

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,347 @@
...
+        &self,
+        hart: &PathBuf,
+        ident: &PackageIdent,
+        target: &PackageTarget,
+    ) -> Result<()> {
+        Counter::UploadRequests.increment();
+        let key = self.s3_key(ident, target).to_string_lossy().into_owned();
+        let file = File::open(hart)?;
+        let size = file.metadata().unwrap().len() as usize;
+        //let min_limit: usize = 49152 * 1024;
```
chefsalim: Remove comment?

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,347 @@
...
+        if size < MINLIMIT {
+            self.single_upload(key, file)
```
chefsalim: Might be nice to also throw in a couple of counters for single vs. multiple upload counts..

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,347 @@
...
+
+        let mut request = PutObjectRequest::default();
+        request.key = key;
+        request.bucket = bucket;
+        request.body = Some(object);
+
+        match self.client.put_object(&request).sync() {
+            Ok(_) => {
+                let end_time = PreciseTime::now();
+                info!("Upload stats ({} sec):", start_time.to(end_time));
```
chefsalim: Would be super nice to have the hart file name in the info

components/builder-depot/src/backend/s3.rs
```diff
@@ -0,0 +1,347 @@
...
+        mprequest.bucket = self.bucket.clone();
+
+        match self.client.create_multipart_upload(&mprequest).sync() {
+            Ok(output) => {
+                let mut request = UploadPartRequest::default();
+                request.key = key.clone();
+                request.bucket = self.bucket.clone();
+                request.upload_id = output.upload_id.clone().unwrap();
+
+                //let cap: usize = 49152 * 1024;
```
chefsalim: Remove comment?

components/builder-api/habitat/default.toml
chefsalim: Actually, "habitat-builder-artifact-store.default" might be a better choice.  Then, we can just replace "default" with the environment name for acceptance / live.  Dev and on-prem both just use the "default" nomenclature.  This would match up with how we are identifying services.

components/builder-depot/src/server.rs
```diff
@@ -868,6 +866,14 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
 
     let filename = depot.archive_path(&ident, &target_from_artifact);
```
chefsalim: I think we can completely remove the archive_path function, and just use the temp file as the path.  In the new S3 world, we should not need to have a carefully crafted archive path.

chefsalim: This will also simplify and tighten up some other stuff.  For example, we are holding a read lock on depot for quite a while - if we remove the archive_path and archive_parent functions, we can tighten up the lock usage to only the place where we check for the targets.

components/builder-depot/src/server.rs
```diff
@@ -880,7 +886,7 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
     }
 
     info!("File added to Depot at {}", filename.to_string_lossy());
```
chefsalim: We can probably remove this from here :)

components/builder-depot/src/server.rs
```diff
@@ -1236,6 +1244,15 @@ fn download_package(req: &mut Request) -> IronResult<Response> {
...
         Ok(package) => {
+            let temp_path = depot.archive_path(package.get_ident(), &target);
+            let temp_ident = ident.to_owned().into();
+            match s3handler.download(&temp_path, &temp_ident, &target) {
+                Ok(_target) => {
+                    Response::with(status::Ok);
+                    _target
+                }
+                Err(_) => return Ok(Response::with(status::NotFound)),
+            }
```
chefsalim: I'm not sure I understand this logic.  We are downloading into a temp path, but we're not renaming it into an archive path.  How is it getting translated into the archive path?  We should either just completely get rid of the archive_path function (which I think is the right thing to do - because on prod we're going to serve mostly from fastly, and on-prem we'll serve from a local store), or if we really really want to use an on-disk cache, then it should be a much simpler path than constructing an elaborate archive path.

components/builder-depot/src/server.rs
```diff
@@ -2512,6 +2532,7 @@ fn download_response_for_archive(archive: PackageArchive) -> IronResult<Response
     };
     response.headers.set(disp);
     response.headers.set(XFileName(archive.file_name()));
+    let _cleanup = remove_file(&archive.path);
```
chefsalim: We shouldn't need this if we remove the pre-constructed paths

support/builder/s3migrate.sh
```diff
@@ -0,0 +1,311 @@
...
+    # traversing the data path and uploads to s3
+    if [ ${#artifacts[@]} -eq 0 ]; then
+        echo ""
+        echo "No artifacts found in Artifact storage path!"
+        echo "Exiting!"
+        exit 1
+    fi
+        keyMsg
+    for hart in ${artifacts[@]}; do
+        uploadMsg
```
chefsalim: If we're echoing a message per hart, it would be nice to also echo the hart file we're uploading..

support/builder/s3migrate.sh
```diff
@@ -0,0 +1,311 @@
...
+    done
+}
+
+generateKey() {
+    # Generates the s3 item key from the hartfile
+    FILEIN="${1}"
+    BASE="${FILEIN##*/}"
+
+    # Get the
+    ORIGIN="$(hab pkg info -j ${FILEIN} | jq -r .origin)"
```
chefsalim: Do the hab pkg info -j command just one instead of 4 times?

support/builder/s3migrate.sh
```diff
@@ -0,0 +1,311 @@
...
+        fi
+    done
+
+    if [ -z $SYSTEM ]; then
+        return 1
+    else
+        echo "${ORIGIN}/${PKGNAME}/${PKGVER}/${RELEASE}/${ARCH}/${SYSTEM}/${BASE}"
+    fi
+}
+setBucket() {
```
chefsalim: Minor : whitespace?

support/builder/s3migrate.sh
```diff
@@ -0,0 +1,311 @@
+#!/bin/bash
+set -euo pipefail
```
chefsalim: I assume we've run shellcheck or other linters on the script, if not, would be good to run those..

support/builder/s3migrate.sh
```diff
@@ -0,0 +1,311 @@
...
+    else
+        echo "${ORIGIN}/${PKGNAME}/${PKGVER}/${RELEASE}/${ARCH}/${SYSTEM}/${BASE}"
+    fi
+}
+setBucket() {
+    # Configure the bucket name to use for the upload
+    echo ""
+    echo "==========================================================="
+    echo "Please enter a target bucket name and press [ENTER]:"
+    read bucket_name
```
chefsalim: Instead of a target bucket name, it would be easier to just have them input the environment name.  Then we could just always have the bucket base name the same, and only change the environment as discussed.  That would be easier and likely less error prone than trying to remember the long bucket name.  We could also check for environments "live", "acceptance", or "default" and only allow those if we wanted.

support/builder/s3migrate.sh
```diff
@@ -0,0 +1,311 @@
...
+        fi
+    fi
+}
+
+setRegion() {
+    # Sets the region used by the bucket
+    echo ""
+    echo "==========================================================="
+    echo "Please enter the region or the minio endpoint for your bucket and press [ENTER]:"
+    echo "(us-west-1 us-east-1 || http://localhost:9000 )"
```
chefsalim: Only ask for this if the endpoint is AWS?

support/builder/s3migrate.sh
```diff
@@ -0,0 +1,311 @@
...
+    credsConfigured=false
+    if [ $s3type = "minio" ]; then
+        echo ""
+        echo "It looks like you specified a migration to minio!"
+        echo "Minio will use whatever credentials you've configured it with."
+        echo "If those credentials don't match your aws credentials file, you"
+        echo "must specify those custom credentials."
+    fi
+
+    CREDS=( "${HOME}/.aws/credentials" "${HOME}/.aws/config" "/root/.aws/credentials" "/root/.aws/config")
```
chefsalim: This doesn't work if you have AWS credentials for other tooling but don't want minio creds to be the same as AWS creds.  It would be great to ask if you want to use AWS creds for minio (default to No).

support/builder/s3migrate.sh
```diff
@@ -0,0 +1,311 @@
...
+        if [[ "${response}" =~ ^([yY][eE][sS]|[yY])+$ ]]; then
+            echo "==========================================================="
+            echo "###########################################################"
+            echo "### CAUTION: This is a destructive operation and cannot ###"
+            echo "### be undone or reversed. It is 100% entirely permanent ##"
+            echo "###########################################################"
+            echo "==========================================================="
+            read -r -p "Are you certain you would like to clean local storage? [y/N]" response2
+                if [[ "${response2}" =~ ^([yY][eE][sS]|[yY])+$ ]]; then
+                    for art in ${artifacts[@]}; do
```
chefsalim: Could just rm -rf the artifact root dir...would be faster.

support/builder/s3migrate.sh
eeyun: Yep, indeed I actually have a local copy that fixes this. Will push it momentarily.

support/builder/s3migrate.sh
eeyun: Nope, we need this for on-prem as well. the region/endpoint are used with minio as well. The local copy I have tries to detect the minio instance from a supervisor ring, but still allows you to specify something custom if you've got more than one instance appropriately named.

support/builder/s3migrate.sh
eeyun: Unfortunately, this won't work. In AWS S3 bucket names have to be _globally_ unique. Which means you can't use a bucket by the same name as any other bucket. This tool quite unfortunately has to support a whole bunch of potential use cases :(

support/builder/s3migrate.sh
eeyun: yep thats exactly how this works!

components/builder-api/habitat/config/config.toml
eeyun: Oh no we do not! :) good catch!

components/builder-api/habitat/default.toml
eeyun: We can definitely change this to something like that, I haven't tested buckets with `.` characters in them, will give that a shot. Though as a note in context to a comment further down the page. If this bucket exists in S3, that naming scheme won't really work.

components/builder-depot/habitat/plan.sh
```diff
@@ -0,0 +1,22 @@
+pkg_name=builder-minio
```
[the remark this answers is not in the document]
eeyun: In the studiorc I think we actually pre-pend `builder` to the services being run. This service is a totally new one and I don't think it . I thought about naming it builder-depot but that felt like a misnomer because its not actually the depot its _just_ minio and nothing else. We can't bundle this in to builder-api either without converting builder-api to a composite plan and that felt like an unnecessary change.

support/builder/s3migrate.sh
eeyun: Yep, shellcheck full-up has been fully run and corrected.

support/builder/s3migrate.sh
chefsalim: Hmm, that's unfortunate.  This brings up one more concern - the migration script could mess up something already configured - eg, we mistakenly fat finger and type in the prod bucket when doing a dev test for example.  Perhaps there should be a check that migration doesn't start if the bucket exists AND there are already items in that bucket?

support/builder/s3migrate.sh
chefsalim: 👍

components/builder-depot/habitat/plan.sh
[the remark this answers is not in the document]
chefsalim: It makes me pretty uncomfortable to have builder-depot component package up a service. Builder-depot will get merged into builder-api at some point.  It's not meant to package new services.  I think we should just create a new component that is just builder-minio, and not overload builder-depot.  That should fulfill your needs and keep things cleanly separated.

support/builder/s3migrate.sh
```diff
@@ -0,0 +1,360 @@
...
+                    read region_name
+                    export AWS_REGION=${region_name}
+                    return
+                fi
+            fi
+        done
+    else
+        echo ""
+        echo "==========================================================="
+        echo "Please enter the region or the minio endpoint for your bucket and press [ENTER]:"
```
chefsalim: Should we be asking about minio endpoints if the s3type is not minio?

support/builder/s3migrate.sh
```diff
@@ -0,0 +1,360 @@
...
+            aws s3api create-bucket --bucket "${bucket_name}" --region "${AWS_REGION}"
+            export S3_BUCKET=${bucket_name}
+        fi
+    fi
+}
+
+setRegion() {
+    # Sets the region used by the bucket
+    if [ "${s3type}" == 'minio' ]; then
+        sgroups=("default" "dev" "prod")
```
chefsalim: Add "acceptance" to this list as well?

support/builder/s3migrate.sh
```diff
@@ -0,0 +1,360 @@
...
+        echo "Exiting!"
+        exit 1
+    fi
+        keyMsg
+    for hart in "${artifacts[@]}"; do
+        uploadMsg
+        keyId=$(generateKey "${hart}")
+        putArtifacts "${hart}" "${S3_BUCKET}" "${AWS_REGION}" "${keyId}"
+    done
+}
```
chefsalim: Super minor - would be nice to echo out a final count of packages successfully migrated.

components/builder-depot/habitat/hooks/run
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+exec 2>&1
+
+source "{{pkg.svc_config_path}}/config.sh"
+
+exec minio server \
```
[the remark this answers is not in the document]
chefsalim: Ah, you're right... cool.. :)  Getting this out of builder-depot will make it much clearer as per my other comment.

components/builder-depot/habitat/plan.sh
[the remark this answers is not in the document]
eeyun: Sure thing. I'll get that switched over.

components/builder-depot/src/server.rs
eeyun: This has been simplified. The archive path is actually just the hartfile + /pkgs. We're no longer generating a complex path we simply use /pkgs as the caching location. A follow-on PR could definitely be made to switch all of this pathing over to using a tempdir or tempfile abstraction but doing so would be a more significat refactor than originally thought. For now hopefully the simplification is a reasonable enough middle-ground.

components/builder-depot/src/server.rs
eeyun: Because of not using a tempfile or tempdir this is still obligatory. Re: my comment above

support/builder/s3migrate.sh
eeyun: Yeah I'll add a warning and confirmation on this!

support/builder/s3migrate.sh
eeyun: Nope, excellent catch!

components/builder-depot/src/server.rs
chefsalim: Ok, cool!

components/builder-depot/src/server.rs
chefsalim: 👍

PR #378, opened 2018-04-17
Fix regression in use of JobGroupGet message

chefsalim:
This fixes a regression from an earlier commit (https://github.com/habitat-sh/builder/commit/048bb282bf173f1d55cf3a68a1cb3b9458c6f0ea) - reverse dependency builds will not work correctly if the include_projects field is not set.


thesentinels:

Thank you for contributing!

PR #386, opened 2018-04-19
OAuth refactor

chefsalim:
This is a significant refactor of the OAuth infrastructure. The key goals are simplification, consolidation, and readiness for addition of OpenId Connect provider(s).  The OAuth model moves from having multiple clients/multiple crates for the OAuth providers, to a single-client/single-crate model, where the client has lightweight providers that plug in dynamically based on the config. 

The configuration is also modified to consolidate all the OAuth config under an [oauth] namespace in the config toml. This guarantees that there is only a single OAuth config, vs the possibility of having multiple configs for OAuth providers present. One thing to note is that the github-api-client remains in place, but it is only responsible for Github App-based calls now.

Devs: note that you will need to update your habitat-env and re-enter the dev studio to pick up the correct configs. The configs on acceptance/prod will also need to be updated when this merges.

Important: With this change, we no longer support the Github token for the hab cli auth.  This means anyone who has not yet moved to using Personal Access Tokens will be required to move when this goes live.

tenor-243626886


thesentinels:

Thank you for contributing!

rsertelon:
Just a comment for the oldee GitHub token. I think having multiple PAT is really important. In my case I use habitat on several machines (up to 3) and I cannot reuse the same PAT easily from builder :) whereas I could generate tokens from GitHub quite easily ;)

chefsalim:
Hi @rsertelon, could you elaborate a bit more on your workflow, on why the same PAT cannot be used across machines, but multiple ones can?  thanks!

components/builder-api/src/server/mod.rs
```diff
@@ -79,28 +71,8 @@ impl HttpGateway for ApiSrv {
...
-            let c = config.github.clone();
-            let client = GitHubClient::new(c.unwrap());
-            basic = Authenticated::new(client.clone(), config.depot.key_dir.clone());
-            admin = Authenticated::new(client, PathBuf::new()).require(FeatureFlags::ADMIN);
-        } else {
-            let c = config.bitbucket.clone();
-            let client = BitbucketClient::new(c.unwrap());
-            basic = Authenticated::new(client.clone(), config.depot.key_dir.clone());
-            admin = Authenticated::new(client, PathBuf::new()).require(FeatureFlags::ADMIN);
-        }
```
raskchanky: So much better, wow.

components/github-api-client/src/config.rs
```diff
@@ -14,16 +14,6 @@
...
-/// URL to GitHub Token endpoint
-pub const DEFAULT_GITHUB_TOKEN_URL: &'static str = "https://github.com/login/oauth/access_token";
-/// Default Client ID providing a value in development environments only.
-///
-/// See https://developer.github.com/apps
-pub const DEV_GITHUB_CLIENT_ID: &'static str = "Iv1.732260b62f84db15";
-/// Default Client Secret providing a value in development environments only.
-///
-/// See https://developer.github.com/apps
-pub const DEV_GITHUB_CLIENT_SECRET: &'static str = "fc7654ed8c65ccfe014cd339a55e3538f935027a";
```
raskchanky: I didn't realize we don't need any of this stuff for the GH API calls we're doing (for connecting plans to Builder).

components/oauth-client/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "oauth-client"
+version = "0.0.0"
+authors = ["Josh Black <raskchanky@gmail.com>"]
```
raskchanky: After all the work you've done, I think your name is certainly justified in being here as well 😁

components/oauth-client/src/client.rs
```diff
@@ -0,0 +1,70 @@
...
+                match reqwest::Proxy::http(&url) {
+                    Ok(p) => {
+                        client.proxy(p);
+                    }
+                    Err(e) => warn!("Invalid proxy url: {}, err: {:?}", url, e),
+                }
+            }
+        }
+
+        let provider: Box<OAuth2Provider> = match &config.provider[..] {
```
raskchanky: Wow, I've never seen the `[..]` syntax before.  Looking at the [Rust syntax index](https://doc.rust-lang.org/stable/book/first-edition/syntax-index.html) makes it seem like that's for collection indexing?  But I don't think I follow how that works if the provider is a boxed trait object.

components/oauth-client/Cargo.toml
chefsalim: :)

components/oauth-client/src/client.rs
chefsalim: This is a bit of an known issue in Rust where you can't just match a &String to a static literal - https://github.com/rust-lang/rust/issues/28606

So you have to do an explicit conversion.. the [..] syntax is basically converting the String to a string slice str (there's other ways to do it as well).

raskchanky: Wow, far out.  Thanks for explaining.

PR #391, opened 2018-04-23
Add infra for OpenId Connect auth

chefsalim:
This PR adds support for Azure AD via an OpenId Connect-based provider that is added to the Builder OAuth infra. The Azure AD provider itself is quite straightforward and uses the userinfo endpoint rather than the id token. 

In addition, there are some updates to the dev environment in this PR that allow for building the builder-api-proxy package via the dev studio.

tenor-89352130


thesentinels:

Thank you for contributing!

chefsalim:
Good catch @raskchanky .. the provider protobuf should be updated.. easy to miss as we currently have code that falls back to OAuthProvider::None, but we should error out on that instead - I'll fix it.

PR #408, opened 2018-05-01
Don't add audit entry for non-build

chefsalim:
Update job audit code to not write entry if a build is not actually issued (eg, for uploads that don't result in any builds being issued).

tenor-261897219

### DIFF
--- a/components/builder-jobsrv/src/server/handlers.rs
+++ b/components/builder-jobsrv/src/server/handlers.rs
@@ -401,22 +401,23 @@ pub fn job_group_create(
             None => state.datastore.create_job_group(&msg, projects)?,
         };
         state.schedule_cli.notify()?;
+
+        // Add audit entry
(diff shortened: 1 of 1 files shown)


raskchanky:
Would it be helpful to have an audit entry that specifically tells us there was no build, as a way to eliminate any ambiguity in the audit log?

chefsalim:
Right, that's actually what we sort of get right now - entries with group id 0, that signifies no build.  80-90% of the current audit logging is basically this. It ends up being like an upload log, but not quite since it doesn't track builder uploads.  When (and if) we add upload audit, we'll probably not want to do it in the jobs audit table though, and for now, I'd rather not have no-op entries clutter up the jobs audit table.

raskchanky:
Fair enough.

The DCO check looks like it's hung on this so I'm going to close the PR and re-open it to see if that fixes it.

PR #410, opened 2018-05-01
Fix duplicate job cancels on job timeout

chefsalim:
This fixes a missed case in the job cancelation logic - when a job cancels due to timeout, but the worker continues to run (eg, if it is in an operation that takes a while to cancel), then the job would continue to get attempt cancelation on every heartbeat, resulting in large number of status messages for the cancel operation.

tenor-248050285


thesentinels:

Thank you for contributing!

PR #411, opened 2018-05-02
Re-start worker on cancel

chefsalim:
Currently, there is no good way to cancel a studio build in progress.  The build worker can get into a situation where a studio build is hung (for example, a custom do_build can drop into a debugger with some Python builds, etc).  Exiting the process and letting the supervisor re-start the worker seems to be the only workable solution atm.

tenor-19705610

### DIFF
--- a/components/builder-worker/src/server.rs
+++ b/components/builder-worker/src/server.rs
@@ -152,7 +152,12 @@ impl Server {
             let reply = self.runner_cli.recv_ack()?;
             self.fe_sock.send(reply, 0)?;
         }
-        Ok(())
+
+        // Currently, there's no good way to cancel a studio build that is in progress.
+        // Terminate the worker process (it will get restarted by the supervisor) in
+        // order to handle the case where a build is hung.
+        warn!("Exiting worker (cancel received)");
+        ::std::process::exit(0);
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/builder-worker/src/server.rs
```diff
@@ -152,7 +152,12 @@ impl Server {
             let reply = self.runner_cli.recv_ack()?;
             self.fe_sock.send(reply, 0)?;
         }
-        Ok(())
+
+        // Currently, there's no good way to cancel a studio build that is in progress.
+        // Terminate the worker process (it will get restarted by the supervisor) in
+        // order to handle the case where a build is hung.
+        warn!("Exiting worker (cancel received)");
+        ::std::process::exit(0);
```
raskchanky: If the build is hung, and we're effectively killing the worker, should we exit with a non-zero exit code?

chefsalim: Probably fine to do that too.. I don't think it really matters here since we're not using the worker exit codes for anything.  One thing to clarify though - all job cancel requests will now exit the worker, not just hung builds.

PR #417, opened 2018-05-04
Use OpenID for authenticating against GitLab

rarenerd:
Authenticating against GitLab was failing because the scope was missing in the request, and it was missing in the protocol definition.

This PR adds the missing pieces and switches to using OpenID for authenticating with GitLab. I've tested this against gitlab.com and a private GitLab EE instance.

These are the URLs that need to be configured:
```
OAUTH_USERINFO_URL="https://gitlab.com/oauth/userinfo"
OAUTH_AUTHORIZE_URL="https://gitlab.com/oauth/authorize"
OAUTH_TOKEN_URL="https://gitlab.com/oauth/token"
```

I did not find a place to document this, so that is not included in this PR.

### DIFF
--- a/components/builder-protocol/protocols/sessionsrv.proto
+++ b/components/builder-protocol/protocols/sessionsrv.proto
@@ -6,6 +6,7 @@ enum OAuthProvider {
   None = 1;
   Bitbucket = 2;
   AzureAD = 3;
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

PR #418, opened 2018-05-04
Add support for fetching packages from an upstream depot

raskchanky:
This adds support for fetching packages from an upstream depot, as you might have guessed from my clever and obscure PR title.

This is achieved through one new line of configuration in the `[depot]` section of `builder-api`'s config file:

```toml
[depot]
upstream_depot = "http://example.com"
```

This is optional, and it's disabled by default.  Some sites, e.g. public Builder, do not need an upstream.

The primary driver for this is on-prem depots and the need to install packages via `hab pkg install`.  Before `hab pkg install` actually installs packages, it tries to fetch the details for a package from the depot, both to see what the latest version is, and verify that it actually exists.  This is the place where the bulk of the action happens.

In the event that we fail to find the package the user wants to install, rather than returning a 404, we attempt to fetch the package from the upstream depot (if one is configured).  Then we fake the upload process, as though the package had been uploaded manually.  Note that the owner of the package in this particular case will be Builder itself.  We also put it into the `stable` channel, because that's where `hab pkg install` expects to find it.  All of this happens in-band with the request to install a package, in an effort to provide the user with immediate results.

Additionally, any time someone makes a request to view the details of a package, we check the upstream depot in a background thread and pull in any updates that are available.

In order to make all of this work, there was a significant amount of refactoring inside `builder-depot`, which is notoriously a bit hairy.  I've tested:
* `hab pkg install` on a package that doesn't exist at all in my local depot (but does exist in the upstream)
* `hab pkg install` on a package that doesn't exist in either local or upstream
* `hab pkg install` on a package that exists in my local depot but is not currently installed
* `hab pkg upload` on a brand new package that doesn't exist in the depot
* `hab pkg upload` on a package that already exists in the depot

As far as I can tell, everything works, but I would appreciate anyone that reviews this PR to run everything through its paces as much as you can, particularly if you already have a local Builder dev env setup.

Also note the accompanying PR to `builder-depot-client` here: https://github.com/habitat-sh/habitat/pull/5034

Closes https://github.com/habitat-sh/builder/issues/150

### DIFF
--- a/components/builder-depot/src/server.rs
+++ b/components/builder-depot/src/server.rs
@@ -14,17 +14,21 @@
 
 use std::fs::{self, File};
 use std::path::PathBuf;
-use std::io::{BufWriter, Read, Write};
+use std::io::{self, BufWriter, Read, Write};
 use std::result;
 use std::str::{FromStr, from_utf8};
+use std::thread::spawn;
 
-use bldr_core::metrics::CounterMetric;
+use bldr_core::access_token::{BUILDER_ACCOUNT_ID, BUILDER_ACCOUNT_NAME};
 use bldr_core::helpers::transition_visibility;
+use bldr_core::metrics::CounterMetric;
 use bodyparser;
+use depot_client::{Client as DepotClient, DisplayProgress};
 use hab_core::package::{ident, FromArchive, Identifiable, PackageArchive, PackageIdent,
                         PackageTarget};
 use hab_core::crypto::keys::{parse_key_str, parse_name_with_rev, PairType};
 use hab_core::crypto::BoxKeyPair;
+use http_gateway::conn::RouteBroker;
 use http_gateway::http::controller::*;
 use http_gateway::http::helpers::{self, all_visibilities, check_origin_access, check_origin_owner,
                                   dont_cache_response, get_param, get_session_user_name,
@@ -53,6 +57,7 @@ use url;
 use uuid::Uuid;
 
 use super::DepotUtil;
+use config::Config;
 use error::{Error, Result};
 use metrics::Counter;
 use handlers;
@@ -420,6 +425,7 @@ fn write_archive(filename: &PathBuf, body: &mut Body) -> Result<PackageArchive> 
     let mut writer = BufWriter::new(file);
     let mut written: i64 = 0;
     let mut buf = [0u8; 100000]; // Our byte buffer
+
     loop {
         let len = body.read(&mut buf)?; // Raise IO errors
         match len {
@@ -437,6 +443,7 @@ fn write_archive(filename: &PathBuf, body: &mut Body) -> Result<PackageArchive> 
             }
         };
     }
+
     Ok(PackageArchive::new(filename))
 }
 
@@ -784,7 +791,17 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
     let temp_name = format!("{}.tmp", Uuid::new_v4());
     let temp_path = parent_path.join(temp_name);
 
-    let mut archive = write_archive(&temp_path, &mut req.body)?;
+    let mut archive = match write_archive(&temp_path, &mut req.body) {
+        Ok(a) => a,
+        Err(e) => {
+            warn!("Error writing archive to disk: {:?}", &e);
+            return Ok(Response::with((
+                status::InternalServerError,
+                format!("ds:up:7, err={:?}", e),
+            )));
+        }
+    };
+
     debug!("Package Archive: {:#?}", archive);
 
     let target_from_artifact = match archive.target() {
@@ -845,37 +862,8 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
 
     // Check with scheduler to ensure we don't have circular deps, if configured
     if depot.config.jobsrv_enabled {
-        let mut pcr_req = JobGraphPackagePreCreate::new();
-        pcr_req.set_ident(format!("{}", ident));
-        pcr_req.set_target(target_from_artifact.to_string());
-
-        let mut pcr_deps = protobuf::RepeatedField::new();
-        let deps_from_artifact = match archive.deps() {
-            Ok(deps) => deps,
-            Err(e) => {
-                info!("Could not get deps from {:#?}: {:#?}", archive, e);
-                return Ok(Response::with((status::UnprocessableEntity, "ds:up:4")));
-            }
-        };
-
-        for ident in deps_from_artifact {
-            let dep_str = format!("{}", ident);
-            pcr_deps.push(dep_str);
-        }
-        pcr_req.set_deps(pcr_deps);
-
-        match route_message::<JobGraphPackagePreCreate, NetOk>(req, &pcr_req) {
-            Ok(_) => (),
-            Err(err) => {
-                if err.get_code() == ErrCode::ENTITY_CONFLICT {
-                    warn!(
-                        "Failed package circular dependency check: {:?}, err: {:?}",
-                        ident, err
-                    );
-                    return Ok(Response::with(status::FailedDependency));
-                }
-                return Ok(render_net_error(&err));
-            }
+        if let Err(r) = check_circular_deps(req, &ident, &target_from_artifact, &mut archive) {
+            return Ok(r);
         }
     }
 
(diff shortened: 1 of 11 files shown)


thesentinels:

Thank you for contributing!

rsertelon:
How are signing keys handled? If my package requires a new upstream origin, will `hab pkg install` be able to download the public key through the 'local' builder?

raskchanky:
@rsertelon I'm not sure I know what you mean by `new upstream origin`.

rsertelon:
Like imagine first setup. Hab starts with no pub keys and uses 'local' builder. It downloads my origin pub key fine and will download core packages through 'local' builder. Will the core pub key be downloaded from 'local' builder so that hab can verify packages?

raskchanky:
@rsertelon I think so?  But I will test that to be sure.

components/builder-depot/src/server.rs
```diff
@@ -2307,6 +2264,393 @@ fn target_from_headers(user_agent_header: &UserAgent) -> result::Result<PackageT
...
+    // is here. We want to pass None as the last parameter to fetch_package, because
+    // this is running on a server and there's no TTY, so we have no need for
+    // a progress bar. Passing None by itself doesn't work, because the compiler
+    // complains that it can't infer the type for None. Passing None::<DisplayProgress>
+    // doesn't work because the compiler complains that it doesn't know the size of
+    // that parameter. So in order to get this to compile, we need to create this 100%
+    // useless struct that implements the DisplayProgress trait, just so we can cast
+    // None to it when we call the function. If someone has a better way of doing this,
+    // fantastic. I will happily change this code and subscribe to your newsletter.
+    struct NoProgress {}
```
chefsalim: Builder has the same issue when making depot client calls - the pattern that was adopted is to create "x_" versions of the calls that don't need to have the progress UI passed in.  If we do that for fetch_package and create a x_fetch_package API (similar to x_download), it should remove the need for the fake progress struct.  Newsletter subscription coming soon.. :)

components/builder-depot/src/server.rs
```diff
@@ -1779,12 +1691,50 @@ fn show_package(req: &mut Request) -> IronResult<Response> {
                 session_id,
                 &ident.get_origin(),
             ));
-            request.set_ident(ident);
+            request.set_ident(ident.clone());
+            let u = helpers::extract_query_value("upstream", req);
+
+            // If the "upstream" query parameter is not passed, then we can assume it's a normal
+            // API request and we should check for an update in the background.
```
chefsalim: I'm not sure I understand the need for the upstream parameter..

components/builder-depot/src/server.rs
```diff
@@ -2307,6 +2264,393 @@ fn target_from_headers(user_agent_header: &UserAgent) -> result::Result<PackageT
...
+    };
+
+    for ident in deps_from_artifact {
+        let dep_str = format!("{}", ident);
+        pcr_deps.push(dep_str);
+    }
+    pcr_req.set_deps(pcr_deps);
+
+    // We need to do it this way instead of via route_message because this function can't be passed
+    // a Request struct, because it's sometimes called from a background thread, and Request
```
chefsalim: Since the upstream already checks for circular deps, I don't believe we need to do that check in the downstream (it shouldn't be possible to introduce a circular dependency by downloading a _new_ package from the upstream, since by definition nothing else is pointing to the package in the downstream).

components/builder-depot/src/server.rs
```diff
@@ -2307,6 +2264,393 @@ fn target_from_headers(user_agent_header: &UserAgent) -> result::Result<PackageT
...
+        return;
+    }
+
+    let depot_cli = depot_client_from_config(&config).unwrap();
+    let upstream_depot = config.upstream_depot.unwrap(); // unwrap is fine because we just checked if it was None
+    let oauth_token = helpers::get_optional_oauth_token(req);
+
+    // Let's do the actual work that takes time in the background so the main request handler can
+    // continue on its way.
+    spawn(move || {
```
chefsalim: Instead of spawning new threads for requests, we may want to consider having a long lived background thread (similar to how we are handling worker manager and scheduler threads) that is responsible for pulling refresh tasks from a list of idents.  The main thread then just has to add an ident to the refresh list (if it's not already present), and let the worker manage churning thru that list every N second intervals.  The main thread could also add multiple idents to refresh (eg, any updated dependencies for the package that is being fetched could all be added at once).

adamhjk:
I'm stoked for this feature. When you combine it with proxy support, this is exactly the flow you want in the Enterprise :tm: - you install a local depot, then just start working like normal. It's dope.

tenor-43624980

components/builder-depot/src/server.rs
```diff
@@ -2307,6 +2264,393 @@ fn target_from_headers(user_agent_header: &UserAgent) -> result::Result<PackageT
...
+    let depot = d.clone();
+    let config = d.config.clone();
+
+    if config.upstream_depot.is_none() {
+        return;
+    }
+
+    let depot_cli = depot_client_from_config(&config).unwrap();
+    let upstream_depot = config.upstream_depot.unwrap(); // unwrap is fine because we just checked if it was None
+    let oauth_token = helpers::get_optional_oauth_token(req);
```
chefsalim: I don't think this is right - the upstream and downstream will / should not share tokens.  We can either just make upstream checks without a token for now (public packages only), or configure an upstream token separately.

components/builder-depot/src/server.rs
```diff
@@ -784,7 +791,16 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
     let temp_name = format!("{}.tmp", Uuid::new_v4());
     let temp_path = parent_path.join(temp_name);
 
-    let mut archive = write_archive(&temp_path, &mut req.body)?;
+    let mut archive = match write_archive(&temp_path, &mut req.body) {
+        Ok(a) => a,
+        Err(_) => {
+            return Ok(Response::with((
+                status::InternalServerError,
+                "Internal Habitat error",
```
chefsalim: Minor - use ideomatic format "xxx:yyy:NNN" for the error string?

components/builder-depot/src/server.rs
```diff
@@ -784,7 +791,16 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
     let temp_name = format!("{}.tmp", Uuid::new_v4());
     let temp_path = parent_path.join(temp_name);
 
-    let mut archive = write_archive(&temp_path, &mut req.body)?;
+    let mut archive = match write_archive(&temp_path, &mut req.body) {
+        Ok(a) => a,
+        Err(_) => {
```
chefsalim: Minor - write out a warn! with the original error code?

components/builder-depot/src/server.rs
```diff
@@ -2602,6 +2946,16 @@ where
     r
 }
 
+fn depot_client_from_config(config: &Config) -> Option<DepotClient> {
+    if let Some(ref upstream_depot) = config.upstream_depot {
+        // TODO: we keep running into this issue where PRODUCT and VERSION aren't available and
+        // hardcoding values. how do we avoid this?
+        Some(DepotClient::new(upstream_depot, "hab", "0.55.0", None).unwrap())
```
chefsalim: The product should probably not be "hab" .. maybe "depot", and leave version at 0.0.0 until we can retrieve the build version?

components/builder-depot/src/server.rs
```diff
@@ -2307,6 +2264,393 @@ fn target_from_headers(user_agent_header: &UserAgent) -> result::Result<PackageT
...
+                Ok(p) => p,
+                Err(e) => {
+                    info!("Error building package from archive: {:#?}", e);
+                    return Err(Error::HabitatCore(e));
+                }
+            };
+
+            let config = depot.config.clone();
+
+            if let Err(e) = process_upload_for_package_archive(
```
chefsalim: It could be a lot less code change and cleaner if we use the package upload endpoint, instead of commonizing the code and calling it from here.   So eventually, we could take the upstream refresh module and run it from anywhere (eg, I could even see it moving to the originsrv instead of the depot after we do the S3 migration).  If we use the endpoint, we can leave the package upload and download code almost exactly as it is currently.

components/builder-depot/src/server.rs
```diff
@@ -1779,12 +1691,50 @@ fn show_package(req: &mut Request) -> IronResult<Response> {
...
+                    // If the package is not found and it's the hab CLI making the request and
+                    // upstream=t is passed in the query string, we can assume someone is trying to
+                    // hab pkg install this particular package, and we want to synchronously check
+                    // the upstream depot before returning a 404 (assuming an upstream depot is
+                    // configured)
+                    if config.upstream_depot.is_some() && code == ErrCode::ENTITY_NOT_FOUND
+                        && helpers::is_request_from_hab(req)
+                        && u.unwrap_or("f".to_string()) == "t"
+                    {
+                        let depot_cli = depot_client_from_config(&config).unwrap();
```
chefsalim: It would be nice if we could encapsulate the use of the depot client to only the refresh code (which could become it's own module/struct - more encapsulated and portable).

chefsalim:
OK added a first pass of comments.  This is a fairly tricky feature with a lot of potential for regressions in the core path so I'm inclined to be a bit more cautious here in my feedback. My main recommendations would be to more tightly encapsulate the refresh code, avoid doing any refresh work on the main thread, use the public endpoint instead of calling upload functions directly, and limit changes to the core upload/download code path to the minimum required (it should be very little to none needed).  Overall this is great progress, it's going to be an outstanding feature @raskchanky

components/builder-depot/src/server.rs
raskchanky: Yeah, good call.

components/builder-depot/src/server.rs
raskchanky: Sure, I can change that.  FWIW, I didn't write any of this error code.  I just moved it from the bottom of the file to here, but your suggestion is a good one.

components/builder-depot/src/server.rs
```diff
@@ -1779,12 +1691,50 @@ fn show_package(req: &mut Request) -> IronResult<Response> {
...
             {
                 Ok(id) => ident = id.into(),
-                Err(err) => return Ok(render_net_error(&err)),
+                Err(err) => {
+                    let code = err.get_code();
+
+                    // If the package is not found and it's the hab CLI making the request and
+                    // upstream=t is passed in the query string, we can assume someone is trying to
+                    // hab pkg install this particular package, and we want to synchronously check
+                    // the upstream depot before returning a 404 (assuming an upstream depot is
```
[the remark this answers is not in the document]
raskchanky: We can, but my main thought here was the UX.  If I run `hab pkg install` and the package I want to install doesn't exist in the depot that I'm pointed at, what should happen?

1) We return the normal error that we normally do when a package doesn't exist, which says "package doesn't exist" or something similar, and then we check the upstream in the background.  That's great from an efficiency perspective, but it's a bummer from a user's perspective.  Now they might think they're stuck, unable to get the package they need.  If they try it again, of course, it will likely succeed this time, assuming the upstream has it and we've downloaded it in the background and enough time has elapsed between their first request and their second, but I'm guessing many (most?) users wouldn't think to try again.

2) We check and download it synchronously within the request.  This felt like the best UX to me because you get immediate results.  The biggest downside here is the person is left waiting a bit longer than normal, because the depot has to download the package from the upstream and process it before it can send it along to the user.  But this synchronous downloading of the upstream package only happens when we've detected that the user is running `hab pkg install`, so they're likely ok with waiting a little bit to get the package they want.

@ryankeairns do you have any thoughts on the best UX here?

components/builder-depot/src/server.rs
raskchanky: The need for it revolved around the UX choice I made for downloading a package via `hab pkg install`.  Since I went with synchronously checking the upstream depot in-band, I needed a way to differentiate between that kind of request and a normal API call to e.g. `/v1/depot/pkgs/core/redis/1.2.3/20180101123456`

components/builder-depot/src/server.rs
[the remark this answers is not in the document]
ryankeairns: In either case, erring on the side of clarity would be prudent. For example, in the first scenario, could you provide words to inform them to try again? "That package doesn't exist/could not be found. If you are certain it does exist in the target repo, then you might try again."

In the second case, a short message like "...we're trying to find it in [here]" or "...we're downloading/processing from the upstream depot..." would inform them why it may feel slow.

components/builder-depot/src/server.rs
raskchanky: I think this is a great idea, but it feels like it would require significant changes to this PR.  Perhaps that's something we can pick up in a future refactor.

components/builder-depot/src/server.rs
raskchanky: Ah, good point.  I think for now it's probably best to just support public packages only.  I'll make the appropriate changes.

components/builder-depot/src/server.rs
raskchanky: I think this is a great idea, but it feels like a pretty big change to make at this point in the PR, for what is essentially a nice-to-have.  I will make a note to file an issue for this kind of enhancement after this PR merges.

components/builder-depot/src/server.rs
raskchanky: Good call

components/builder-depot/src/server.rs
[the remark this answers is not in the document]
chefsalim: Returning "Not Found" to the user seems perfectly valid (and correct) - the package literally isn't in the local depot. We can tweak the cli UX, but it might be better to simply advise admins to bootstrap core packages using the on-prem tool we already have for now - that minimizes (maybe even eliminates) the UX issue.  The "not found" case is a very small subset of the staleness case, I don't think we should worry about "not found".  The entire strategy for now is to let the update happen in the background, and we should stick with that strategy.  This will all have a very nice fix once we have notifications from the upstream - we'll refresh (in the background) on notification from the upstream, thus greatly minimizing the staleness window.

components/builder-depot/src/server.rs
chefsalim: 👍

components/builder-depot/src/server.rs
chefsalim: 👍

components/builder-depot/src/server.rs
chefsalim: Sounds good - should be an easy copy-pasta from other of the other modules

components/builder-depot/src/server.rs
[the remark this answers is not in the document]
raskchanky: Ok, I'll make that change.

raskchanky:
@chefsalim I think I've addressed your initial pass of comments.  Let me know if you have any further thoughts.

components/builder-depot/src/server.rs
chefsalim: Sadly, these seem like the only options for now - if we go with the no-op struct, we could move it to the depot client as well so that we could take advantage of it from other places as well.

raskchanky: Yeah, fair enough.

PR #426, opened 2018-05-10
Add proxy timeout config to builder-api-proxy

chefsalim:
This change adds a proxy timeout config setting to the builder-api-proxy. It is useful when we need to adjust the service to handle upload/download of large files.

tenor-167176814

### DIFF
--- a/components/builder-api-proxy/habitat/default.toml
+++ b/components/builder-api-proxy/habitat/default.toml
@@ -37,6 +37,8 @@ worker_connections   = 8000
 worker_processes     = "auto"
 worker_rlimit_nofile = 8192
 max_body_size        = "1024m"
+proxy_send_timeout   = 60
+proxy_read_timeout   = 60
 
 [http]
 keepalive_timeout = "20s"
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/builder-api-proxy/habitat/default.toml
```diff
@@ -37,6 +37,8 @@ worker_connections   = 8000
 worker_processes     = "auto"
 worker_rlimit_nofile = 8192
 max_body_size        = "1024m"
+proxy_send_timeout   = 60
+proxy_read_timeout   = 60
```
christophermaier: Does this need to be synced with the 120 timeout you added on the client?

chefsalim: No, this is basically the same default as what we have today. For the cases where we need to increase it (in on-prem depot), the values will be updated via user.toml automatically.  The hab client 120sec is a purposely larger default since it's a bit more cumbersome to have to change it via the env variable.

Issue #422, opened 2018-05-08
Optimize the upstream depot processing


raskchanky:
There is some room to optimize the code that downloads packages from upstream depots right now:

* Instead of spawning a new thread each time we want to check for an update, we could have one long running background thread that pulls tasks from a list of idents to refresh.  The main thread just pushes idents to that list whenever something needs to be updated.  This would also let us seed that list with a list of packages to install the first time the depot starts up, which would eliminate the need for some of our current bootstrapping scripts.

* The logic to pull packages down and upload them could use builder-depot-client, which would make the entire process more encapsulated and portable.

chefsalim:
A couple of other items:
* Throttling of calls to upstream - the upstream processing can put a lot of extra pressure on API calls to the upstream - having a long lived thread should allow more optimization on how often and how many calls we are doing to the upstream
* Better logging - it would be nice to have a persistent log (similar to the builder-scheduler.log on disk) that audits the refreshes (including failures).

raskchanky:
This is done

PR #432, opened 2018-05-18
Refactor upstream processing

chefsalim:
This change refactors the on-prem upstream processing:
* Add a single background thread that does the upstream package handling
* Bunch up requests so we don't incur duplicate checks over short periods
* Add option of specifying multiple origins (current default: core only)
* Add on-disk log file for upstream-based update success/failures
* Refactored DepotUtil to become a trait and remove extra object from Request
* Fixed a bug where non-qualified idents were not getting handled
* Removed some dead code (doctor.rs)


raskchanky:
Closes https://github.com/habitat-sh/builder/issues/422

components/builder-depot/src/config.rs
```diff
@@ -52,6 +52,8 @@ pub struct Config {
     /// Upstream depot to pull packages from if someone tries to install from this depot and they
     /// aren't present. This is optional because e.g. public Builder doesn't have an upstream.
     pub upstream_depot: Option<String>,
+    // Origins for which we pull from upstream (default: core)
+    pub upstream_origins: String,
```
raskchanky: Why a `String` instead of a `Vec<String>`?

chefsalim: Ah good point - I have it set up to take a comma separated string with origins, but having it be a Vec at the config level would be better - I'll update it.

components/builder-depot/src/upstream.rs
```diff
@@ -0,0 +1,241 @@
...
+            Some(DepotClient::new(upstream_depot, "builder-upstream", "0.0.0", None).unwrap())
+        } else {
+            None
+        };
+
+        let log_path = cfg.log_dir.clone();
+        let mut logger = Logger::init(PathBuf::from(log_path), "builder-upstream.log");
+
+        let want_origins: HashSet<String> =
+            cfg.upstream_origins.iter().map(|s| s.to_owned()).collect();
```
raskchanky: Is the `to_owned()` necessary here?  Isn't `cfg.upstream_origins` already a `Vec<String>`?

raskchanky: In fact, since you're just converting a `Vec` to a `HashSet`, you could probably do that with something like:

`let want_origins: HashSet<String> = HashSet::from_iter(cfg.upstream_origins);`

chefsalim: 👍 will try that

chefsalim: ah no, its not that simple - the ownership does need to change as we cant hand the ownership of the strings from the borrowed config to the hashset.  so the from_iter fails with a borrow checker error.

raskchanky: Ah, I forgot about the borrowed config.  Oh well 😁

chefsalim: one can run, but not hide from the borrow checker :)

PR #439, opened 2018-05-23
Refactor the way targets are calculated.

raskchanky:
Specifically, via a URL query parameter if present, with a fallback to
the UserAgent header if not.

This is in support of including windows packages in the on-prem bootstrap archive.


thesentinels:

Thank you for contributing!

PR #440, opened 2018-05-23
Bug fixes for upstream handling

chefsalim:
This fixes some issues with the upstream handling:
1. The package version comparison between local and remote was incorrect if the local package that triggered the check was an older version
2. Packages could get downloaded multiple times when triggered with a non-qualified ident
3. The logging could incorrectly log an update in some cases where the download did not actually occur

tenor-102525593


thesentinels:

Thank you for contributing!

PR #441, opened 2018-05-24
Remove un-needed files

chefsalim:
Builder doesn't track versions explicitly, so we don't need a VERSION file.  Also builder releases are fluid and we don't need to track a CHANGELOG.


thesentinels:

Thank you for contributing!

PR #445, opened 2018-05-25
Fix perf issue in show_package

chefsalim:
This change is a fix/workaround for a perf issue in the builder API where we have been needing to crack open package archives to determine whether the package is a service. This flag is only used by the web UI to determine whether to show the "run / export" commands in the package details side bar (mostly cosmetic). We will eventually need to get to a better solution to determine this status.

tenor-14200822


thesentinels:

Thank you for contributing!

PR #449, opened 2018-05-29
Remove archive check from show_package

chefsalim:
This change removes a costly check from show_package API call.  If we have valid metadata in originsrv for a package, we will assume that it's archive is in the correct location as well.  This will be a valid assumption as we move package storage to S3 shortly. We also don't need to re-check the target since we already pass it in to the OriginPackageGet call.

tenor-13670767

### DIFF
--- a/components/builder-depot/src/server.rs
+++ b/components/builder-depot/src/server.rs
@@ -1764,26 +1764,7 @@ fn show_package(req: &mut Request) -> IronResult<Response> {
         request.set_ident(ident.clone());
 
         match route_message::<OriginPackageGet, OriginPackage>(req, &request) {
-            Ok(pkg) => {
-                let lock = req.get::<persistent::State<Config>>()
-                    .expect("depot not found");
-
-                let depot = lock.read().expect("depot read lock is poisoned");
-                let pkg_target = PackageTarget::from_str(&target);
-
-                // If the target doesn't parse or there's no archive on disk, return 404
-                if pkg_target.is_err() || !depot.archive(&ident, &pkg_target.unwrap()).is_some() {
-                    return Ok(Response::with(status::NotFound));
-                };
-
-                // If the request was for a fully qualified ident, cache the response, otherwise do
-                // not cache
-                if qualified {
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/builder-depot/src/server.rs
```diff
@@ -1764,26 +1764,7 @@ fn show_package(req: &mut Request) -> IronResult<Response> {
...
-            Ok(pkg) => {
-                let lock = req.get::<persistent::State<Config>>()
-                    .expect("depot not found");
-
-                let depot = lock.read().expect("depot read lock is poisoned");
-                let pkg_target = PackageTarget::from_str(&target);
-
-                // If the target doesn't parse or there's no archive on disk, return 404
-                if pkg_target.is_err() || !depot.archive(&ident, &pkg_target.unwrap()).is_some() {
-                    return Ok(Response::with(status::NotFound));
```
raskchanky: This error handling for the target feels valuable to keep, but is probably best moved to the top of this function where we create the `target` variable.  `target_from_headers` already makes the `PackageTarget::from_str()` call, but we blindly unwrap it instead of checking for errors and returning `NotFound`.

Also, we should be checking the target that we get from the URL, if one exists.

raskchanky:
I have an uneasy feeling about approving this while we still have packages on disk.  I remember there were some instances of files on disk and database metadata getting out of sync.  I searched the code but couldn't find the place where we used to handle that or at least warn about it.

components/builder-depot/src/server.rs
chefsalim: Good point - yes, a target check up front and removal / removal of blind unwraps seems like the right thing - I'll play with it a bit more.

chefsalim:
@raskchanky I don't know of any instances of being out of sync on prod - the only scenarios are in cases like dev (and rarely acceptance) where we're moving metadata around independently of the on-disk bits.  But even in that case, we will fail gracefully when trying to d/l the package itself, which is the only time we should ever be touching the hart file. show_package should just be about what's in our DB.

PR #452, opened 2018-06-01
False positive "already started" messages in the dev-env w/ v0.56.0

eeyun:
Seeing false positive "already started" messages in the builder dev-env after upgrading to hab v0.56.0. I _think_ it's related to the exit code provided by `hab sup status` now we grep for appropriate output to determine whether a service is started.

### DIFF
--- a/.studiorc
+++ b/.studiorc
@@ -164,7 +164,7 @@ start-builder() { stop-on-failure _start-builder "$@"; }
 _load-if-not-loaded() {
   local pkg_ident
   pkg_ident=$1
-  if hab sup status "$pkg_ident" > /dev/null; then
+  if hab sup status "${pkg_ident}" | grep "${pkg_ident}" > /dev/null; then
     echo "$pkg_ident is already loaded"
   else
     hab svc load "$@"
@@ -173,7 +173,7 @@ _load-if-not-loaded() {
 load-if-not-loaded() { stop-on-failure _load-if-not-loaded "$@"; }
 
 start-datastore() {
-  if hab sup status habitat/builder-datastore > /dev/null; then
+if hab sup status habitat/builder-datastore | grep "builder-datastore" > /dev/null; then
     echo "habitat/builder-datastore is already loaded"
   else
     init-datastore


thesentinels:

Thank you for contributing!

eeyun:
Exit code is 127. I think the last time we discussed this that exit code was intentional?

raskchanky:
Yeah, that's a launcher exit code, if I'm not mistaken.

raskchanky:
I'm wrong.  The launcher exit codes are 84 and 86.

https://github.com/habitat-sh/habitat/blob/master/components/launcher-protocol/src/lib.rs#L36-L39

.studiorc
```diff
@@ -173,7 +173,7 @@ _load-if-not-loaded() {
 load-if-not-loaded() { stop-on-failure _load-if-not-loaded "$@"; }
 
 start-datastore() {
-  if hab sup status habitat/builder-datastore > /dev/null; then
+if hab sup status habitat/builder-datastore | grep "builder-datastore" > /dev/null; then
```
raskchanky: Should this be `grep "habitat/builder-datastore"`?

chefsalim: Just "builder-datastore" seems sufficient?

raskchanky:
Are these changes still necessary after https://github.com/habitat-sh/habitat/pull/5166 merged?  @baumanj

.studiorc
raskchanky: My point was just to make the grepping consistent, given that on line 167, the grep will resolve to something like `grep habitat/builder-router` or whatever the package name is.

baumanj:
Good point, @raskchanky. This change *is* necessary after https://github.com/habitat-sh/habitat/pull/5166, but should be removed once https://github.com/habitat-sh/habitat/issues/5172 is fixed. The former prevents the supervisor from crashing, but it still has the wrong exit code when no services are loaded and so the grep is required for now, but that's a temporary fix which we'd like to remove since depending on the details of the output isn't as robust as just checking the exit code.

eeyun:
Yeah this fix got incidentally pulled in to the backend PRs I'm going to close this one.

PR #468, opened 2018-06-08
[WIP] Adds s3bulk uploader tool

eeyun:
In order to safely migrate prod/acceptance to s3 this tool will get leveraged by the s3migrate script, doing so should cut our migration time down astronomically.

A majority of this PR is literally vendored deps following the standard golang community pattern. 


elliott-davis:
Also, I wouldn't sweat the vendor thing:
* https://github.com/habitat-sh/habitat-operator/tree/master/vendor
* https://github.com/kubernetes/apimachinery/tree/master/vendor
* https://github.com/kubernetes/kubernetes/tree/master/vendor
* https://github.com/cockroachdb/vendored/tree/c0f3dc60e5b8b3b2f4f322491f31b042c5900a5e

chefsalim:
The vendoring is definitely ugly - is that a Go pattern?  Ideally we would vendor in what we need as part of plan build time - not possible?

raskchanky:
It's a Go pattern, which will likely change in the near future after their new dep tool becomes stable.  I can't see that it really matters though.  It accomplishes a goal, and when we don't need it any more, we delete it.  I don't understand the uproar here.

tools/s3bulk/main.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+}
+
+// upload function for workers
+// uploads a given file to s3
+func (worker *Worker) upload(file string) (string, error) {
+	res, err := exec.Command("hab", "pkg", "info", "-j", fmt.Sprintf("%s%s", worker.SourceDir, file)).Output()
+
+	if err != nil {
+		panic(err)
```
chefsalim: Emit a warning and return instead of panic? (assuming the panic exits the process)

tools/s3bulk/main.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	var pkgInfo PkgInfo
+
+	err = json.Unmarshal(res, &pkgInfo)
+	if err != nil {
+		panic(err)
```
chefsalim: Warning instead of panic?

tools/s3bulk/main.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	// s3 destination file path
+	var platform string
+
+	if strings.Contains(file, "windows") {
```
chefsalim: More resilient to split the filename and explicitly check the platform component instead of just a contains?

tools/s3bulk/main.go
```diff
@@ -0,0 +1,237 @@
...
+		platform = "windows"
+	} else {
```
chefsalim: More resilient to check for linux here, and bail with a warning if not found instead of defaulting?

tools/s3bulk/main.go
```diff
@@ -0,0 +1,237 @@
...
+
+	filename := strings.Split(file, "/")
+	destfile := fmt.Sprintf("/%s/%s/%s/%s/x86_64/%s/%s", pkgInfo.Origin, pkgInfo.Name, pkgInfo.Version, pkgInfo.Release, platform, filename[len(filename)-1])
+	worker.println("uploading to " + destfile)
+
+	// open and read file
+	f, err := os.Open(worker.SourceDir + file)
+
+	if err != nil {
+		return "Couldn't open file", err
```
chefsalim: Print out the filename/path along with the err?

tools/s3bulk/main.go
```diff
@@ -0,0 +1,237 @@
...
+	fileBytes := bytes.NewReader(buffer)
+
+	params := &s3.PutObjectInput{
+		Bucket: aws.String(worker.Bucket),
+		Key:    aws.String(destfile),
+		Body:   fileBytes,
+		ACL:    aws.String(worker.Acl),
+	}
+
+
```
chefsalim: whitespace

tools/s3bulk/main.go
elliott-davis: Are you thinking of the case where someone has a package named windows?

tools/s3bulk/main.go
```diff
@@ -0,0 +1,237 @@
...
+// doUploads function for workers
+//
+// reads from the file channel (queue),
+// calls upload function for each,
+// then moves uploaded files to worker.DestDir
+func (worker *Worker) doUploads() {
+
+	defer worker.Wg.Done() // notify parent when I complete
+
+	worker.println("doUploads() started")
```
chefsalim: change to just "Uploads started"?  (this will also run in on-prem instances so it's not just us seeing the output)

tools/s3bulk/main.go
```diff
@@ -0,0 +1,237 @@
...
+func main() {
+
+	bucketFlag := flag.String("bucket", "my-s3-bucket", "s3 bucket to upload to")
+	subfolderFlag := flag.String("subfolder", "", "subfolder in s3 bucket, can be blank")
+	num_workersFlag := flag.Int("workers", 100, "number of upload workers to use")
+	regionFlag := flag.String("region", "eu-west-1", "aws region")
+	aclFlag := flag.String("acl", "private", "s3 upload acl - use either private or public")
+	sourceDirFlag := flag.String("sourcedir", "files/", "source directory")
+	destDirFlag := flag.String("destdir", "files-uploaded/", "dest dir for uploaded files (on local box)")
+    endpointFlag := flag.String("endpoint", "http://localhost:9000", "S3 URI in case of local S3")
```
chefsalim: Indent?

tools/s3bulk/main.go
```diff
@@ -0,0 +1,237 @@
...
+	flag.Parse()
+
+	bucket := *bucketFlag
+	subfolder := *subfolderFlag
+	num_workers := *num_workersFlag
+	region := *regionFlag
+	acl := *aclFlag
+	sourceDir := *sourceDirFlag
+	destDir := *destDirFlag
+    endpoint := *endpointFlag
```
chefsalim: Indent?

tools/s3bulk/main.go
```diff
@@ -0,0 +1,237 @@
...
+	// set up s3 credentials from environment variables
+	// these are shared to every worker
+	creds := credentials.NewEnvCredentials()
+
+	fmt.Println("Starting " + strconv.Itoa(num_workers) + " workers...")
+
+	// create the desired number of workers
+	for i := 1; i <= num_workers; i++ {
+		// make a new worker
+        awsConfig := &aws.Config{Region: aws.String(region), Credentials: creds, LogLevel: aws.LogLevel(1), S3ForcePathStyle: aws.Bool(true)}
```
chefsalim: Indent

tools/s3bulk/main.go
```diff
@@ -0,0 +1,237 @@
...
+		if err != nil {
+			worker.println("error uploading" + file + ": " + response + " " + err.Error())
+		} else {
+			worker.println(response)
+			// make destination directory if needed
+			filename := path.Base(file)
+			directory := strings.Replace(file, "/"+filename, "", 1)
+			os.MkdirAll(worker.DestDir+directory, 0775)
+			// move file
+			os.Rename(worker.SourceDir+file, worker.DestDir+file)
```
chefsalim: I think it would be  better to just leave the artifacts in place as is instead of a move right away.. do cleanup separately..eg, fir on-prem allow operator to cleanup after everything is verified to be working well.

tools/s3bulk/main.go
```diff
@@ -0,0 +1,237 @@
...
+	// loop until I receive "" as a termination signal
+	for {
+		file := <-worker.File_channel
+		if file == "" {
+			break
+		}
+		worker.println("File to upload: " + file)
+		response, err := worker.upload(file)
+		if err != nil {
+			worker.println("error uploading" + file + ": " + response + " " + err.Error())
```
chefsalim: Will this error out if the artifact is already present in S3?  If so, maybe phrase it as more of a warning in that case.. the tool should be friendly if run a second time.

tools/s3bulk/main.go
```diff
@@ -0,0 +1,237 @@
...
+	worker.println("doUploads() finished")
+}
+
+// function to print out messages prefixed with worker-[id]
+func (worker *Worker) println(message string) {
+	fmt.Println("Worker-" + strconv.Itoa(worker.Id) + ": " + message)
+}
+
+func main() {
+
```
chefsalim: whitepsace

tools/s3bulk/plan.sh
```diff
@@ -0,0 +1,36 @@
+pkg_name=s3bulk
```
chefsalim: Minor - maybe rename a bit more descriptive - eg, s3bulkupload

tools/s3bulk/main.go
chefsalim: Yeah.. "xwindows" is still a thing :)

chefsalim:
@raskchanky no worries.. no one has to die.. I'll be ok putting this in a standalone repo.. then we can nuke it from outer space some day.

eeyun:
Ok I'm closing this PR and creating a new repo to house this code, will also apply all of salim's suggestions in that new repo and link it back here.

PR #470, opened 2018-06-11
Updates to S3migrate & tweaks to s3backend

eeyun:
A couple of minor tweaks here
1. downcased the backend enum so that in the default.toml's etc we don't have a camel-case "Aws" or "Minio".
2. Changes default.toml to match the default values in the rust component.
3. Most importantly, modifies the s3migrate.sh tool to use the s3bulk go tool.

### DIFF
--- a/components/builder-depot/src/config.rs
+++ b/components/builder-depot/src/config.rs
@@ -103,6 +103,7 @@ impl GatewayCfg for Config {
 }
 
 #[derive(Debug, Clone, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
 pub enum S3Backend {
     Aws,
     Minio,
@@ -188,7 +189,7 @@ mod tests {
         port = 9000
 
         [s3]
-        backend = "Aws"
+        backend = "aws"
         endpoint = "https://aws"
         bucket_name = "mybucket"
 
--- a/support/builder/s3migrate.sh
+++ b/support/builder/s3migrate.sh
@@ -25,7 +25,7 @@ keyMsg() {
 
 uploadMsg() {
     echo ""
-    echo "Uploading hartfile."
+    echo "Uploading hartfiles."
     echo "======================================"
 }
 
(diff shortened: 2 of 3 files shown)


components/builder-depot/src/config.rs
```diff
@@ -101,11 +101,13 @@ impl GatewayCfg for Config {
...
 }
+// maybe lowercase this shit
 
-#[derive(Debug, Clone, Deserialize, PartialEq)]
+#[derive(Debug, Clone, Deserialize)]
+#[allow(non_camel_case_types)]
 pub enum S3Backend {
-    Aws,
-    Minio,
+    aws,
```
chefsalim: I'm not sure lowercasing the enum itself is the way to go.. all our enums are mixed case.  We may need to add a FromStr somewhere instead?

support/builder/s3migrate.sh
```diff
@@ -197,23 +172,26 @@ setRegion() {
...
         echo ""
-        aws s3api put-object --bucket "${BUCKET}" --key "${KEY}"  --body "${BODY}" --region "${REGION}"
+        s3-bulk-upload -backend="aws" -region="${REGION}" -sourcedir="${DIR}" -bucket="${BUCKET}"
     fi
 }
 
 genS3Config() {
-    aws configure
+    echo ""
+    read -r -p "AWS Access Key ID [DEFAULTKEYID]: " access_key_id
```
chefsalim: Are these defaults in brackets correct and do they get applied if the user just hits enter?

eeyun: These defaults _used_ to be correct but they aren't after our tweaks from friday, let me update this so that both of these things are true.

components/builder-depot/src/config.rs
eeyun: I'm not sure exactly how that would work, but let me take a look and see if we can.

support/builder/s3migrate.sh
eeyun: Actually, the thing about changing these is that this function gets used both for S3 as well as minio. I originally had these DEFAULTKEYID and DEFAULTSECRETACCESSKEY because they more closely mirrored what an expected string for the s3 side of things would look like. Ergonomically does it make more sense for us to keep these this way, or should I reconfigure to match the rest of the code?

chefsalim: Right, I'm not sure how the default values here actually get set.. you could just remove them, I think that would be fine

eeyun: Yeah I think thats the right thing. Better not to confuse people, the defaults _can_ get set quite easily but from a UX perspective that might add further confusion

components/builder-depot/src/config.rs
```diff
@@ -101,8 +101,10 @@ impl GatewayCfg for Config {
         self.routers.as_slice()
     }
 }
+// maybe lowercase this shit
```
chefsalim: Is this a todo?  what else is remaining?

PR #472, opened 2018-06-11
Add new active directory provider

elliott-davis:
There is no username in oidc for AD so we duplicate it for our oauth impl

### DIFF
--- a/components/builder-protocol/protocols/sessionsrv.proto
+++ b/components/builder-protocol/protocols/sessionsrv.proto
@@ -8,6 +8,7 @@ enum OAuthProvider {
   AzureAD = 3;
   GitLab = 4;
   Okta = 5;
+  ActiveDirectory = 6;
 }
 
 message Account {
--- a/components/builder-web/app/oauth-providers/index.ts
+++ b/components/builder-web/app/oauth-providers/index.ts
@@ -1,9 +1,10 @@
 enum OAuthProviderType {
+  ActiveDirectory = 'active-directory',
   AzureAD = 'azure-ad',
   GitHub = 'github',
   GitLab = 'gitlab',
   Bitbucket = 'bitbucket',
-  Okta = 'okta'
+  Okta = 'okta',
 }
 
 export abstract class OAuthProvider {
@@ -43,6 +44,8 @@ export abstract class OAuthProvider {
 
   static fromConfig(type: string, clientID: string, authorizeUrl: string, redirectUrl: string, signupUrl: string, state: string): OAuthProvider {
     switch (type) {
+      case OAuthProviderType.ActiveDirectory:
+        return new ActiveDirectoryProvider(clientID, authorizeUrl, redirectUrl, signupUrl, state);
       case OAuthProviderType.AzureAD:
         return new AzureADProvider(clientID, authorizeUrl, redirectUrl, signupUrl, state);
       case OAuthProviderType.GitHub:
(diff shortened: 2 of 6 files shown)


thesentinels:

Thank you for contributing!

components/builder-web/app/oauth-providers/index.ts
```diff
@@ -63,6 +66,29 @@ export abstract class OAuthProvider {
...
+      redirectUrl,
+      signupUrl,
+      true,
+      {
+        client_id: clientID,
+        redirect_uri: redirectUrl,
+        response_type: 'code',
+        state: state,
+        scope: 'openid',
+        nonce: 0 // Ok since we're not using id_token
```
cnunciato: Are all of these properties actually required, or does the destination service just ignore the ones it doesn't need? I only ask because I'd like to restrict them to only what's needed.

components/builder-web/app/oauth-providers/index.ts
```diff
@@ -43,6 +44,8 @@ export abstract class OAuthProvider {
 
   static fromConfig(type: string, clientID: string, authorizeUrl: string, redirectUrl: string, signupUrl: string, state: string): OAuthProvider {
     switch (type) {
+      case OAuthProviderType.ActiveDirectory:
+        return new ActiveDirectoryProvider(clientID, authorizeUrl, redirectUrl, signupUrl, state);
```
cnunciato: Same comment as below here: I'd just ask that we be sure these arguments are limited to those required specifically by this provider.

components/builder-protocol/protocols/sessionsrv.proto
```diff
@@ -8,6 +8,7 @@ enum OAuthProvider {
   AzureAD = 3;
   GitLab = 4;
   Okta = 5;
+  ActiveDirectory = 6;
```
cnunciato: Are there any other flavors of AD we know of that this naming might conflict with or otherwise complicate?

chefsalim:
This looks good to me.. I can't tell though where the upn related change is between this and azure-ad.

components/builder-web/app/oauth-providers/index.ts
elliott-davis: I tried limiting them (for example removing the signup url) and the typechecker went crazy. It looks like this is duplicated in every provider from what I can tell.

components/builder-protocol/protocols/sessionsrv.proto
elliott-davis: perhaps? I think it'll depend on your AD setup. I'd say since this is the first case it's okay and we can expand it from there

components/builder-web/app/oauth-providers/index.ts
elliott-davis: not entirely sure - I won't like, I carbon copied these from all the others

elliott-davis:
@chefsalim mostly here https://github.com/habitat-sh/builder/pull/472/files#diff-5b2be4477477bda3d98a48fff325934aR34 - there just wasn't a upn field returned from their AD over oidc

chefsalim:
Ah I see the upn change now.. it's smaller than I thought it would be

components/builder-protocol/protocols/sessionsrv.proto
chefsalim: Yep - this can be our generic non-azure AD provider

elliott-davis:
that one line change got big pretty fast

PR #482, opened 2018-06-14
Adds support for SSL with signed certificates to Builder-Minio package

eeyun:

### DIFF
--- a/components/builder-minio/habitat/hooks/init
+++ b/components/builder-minio/habitat/hooks/init
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+exec 2>&1
+
+{{#if cfg.use_ssl}}
+    mkdir -p {{pkg.svc_config_path}}/certs
+    cp {{pkg.svc_files_path}}/private.key {{pkg.svc_config_path}}/certs/private.key
+    cp {{pkg.svc_files_path}}/public.crt {{pkg.svc_config_path}}/certs/private.crt
+{{/if ~}}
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

components/builder-minio/habitat/hooks/init
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+exec 2>&1
+
+{{#if cfg.use_ssl}}
+    mkdir -p {{pkg.svc_config_path}}/certs
+    ln -s {{pkg.svc_files_path}}/private.key {{pkg.svc_config_path}}/certs/private.key
```
chefsalim: Should the key file names be configurable?

eeyun: They should be, but minio doesn't support that, they have to be named this specific way.

components/builder-minio/habitat/hooks/init
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+exec 2>&1
+
+{{#if cfg.use_ssl}}
+    mkdir -p {{pkg.svc_config_path}}/certs
+    cp {{pkg.svc_files_path}}/private.key {{pkg.svc_config_path}}/certs/private.key
```
eeyun: By switching these to `cp` from `ln` we'll get a loop on init failure until the files get uploaded to the service group which is exactly what we want here.

chefsalim: 👍

PR #487, opened 2018-06-16
Add audit logs for package promotion and demotion

raskchanky:
This writes an entry into an `audit` table every time a package is promoted or demoted.  It includes the package, the channel, what type of operation (promote/demote), the source that triggered the promotion/demotion (hab CLI, Builder UI, unknown), and the user that triggered it.  It mostly follows the pattern of audit logs from jobsrv, with a few slight tweaks.

### DIFF
--- a/components/builder-protocol/protocols/originsrv.proto
+++ b/components/builder-protocol/protocols/originsrv.proto
@@ -15,6 +15,27 @@ message AccountInvitationListResponse {
   repeated OriginInvitation invitations = 2;
 }
 
+enum PackageChannelTrigger {
+  Unknown = 0;
+  BuilderUI = 1;
+  HabClient = 2;
+}
+
+enum PackageChannelOperation {
+  Promote = 0;
+  Demote = 1;
+}
+
+message PackageChannelAudit {
+  optional uint64 package_id = 1;
+  optional uint64 channel_id = 2;
+  optional PackageChannelOperation operation = 3;
+  optional PackageChannelTrigger trigger = 4;
+  optional uint64 requester_id = 5;
+  optional string requester_name = 6;
+  optional uint64 origin_id = 7;
+}
+
 message CheckOriginAccessRequest {
   oneof account_info {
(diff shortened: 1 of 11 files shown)


components/builder-protocol/protocols/originsrv.proto
```diff
@@ -15,6 +15,27 @@ message AccountInvitationListResponse {
...
+}
+
+message PackageChannelAudit {
+  optional uint64 package_id = 1;
+  optional uint64 channel_id = 2;
+  optional PackageChannelOperation operation = 3;
+  optional PackageChannelTrigger trigger = 4;
+  optional uint64 requester_id = 5;
+  optional string requester_name = 6;
+  optional string origin_name = 7;
```
chefsalim: Is origin name here also meant to be persisted?

chefsalim:
Overall looks pretty good to me @raskchanky ... this doesn't catch the promotion/demotion that is done via 'hab bldr job group promote/demote' and we should add that to the audit log as well, but if you want you can do that in a follow-up PR.

components/builder-protocol/protocols/originsrv.proto
raskchanky: No, just for routing purposes.

chefsalim:
tenor-154226273

components/builder-protocol/protocols/originsrv.proto
chefsalim: 👍

raskchanky: Although now that I'm thinking about it some more, it's probably appropriate for this to be `origin_id` instead of `origin_name` and for it to be persisted.  I'll update this.

raskchanky:
@chefsalim I added in the `origin_id` to the protocol message and persisted it, which felt more consistent with how the rest of the protocol messages work.  Did you want to review fresh or shall I just merge this when Travis is green?

PR #490, opened 2018-06-19
Use tmpfile for s3 transitional caching

eeyun:
Signed-off-by: Ian Henry ihenry@chef.io

### DIFF
--- a/components/builder-depot/src/lib.rs
+++ b/components/builder-depot/src/lib.rs
@@ -77,7 +77,7 @@ use iron::typemap;
 pub trait DepotUtil {
     fn archive<T: Identifiable>(&self, ident: &T, target: &PackageTarget)
         -> Option<PackageArchive>;
-    fn archive_path<T: Identifiable>(&self, ident: &T, target: &PackageTarget) -> PathBuf;
+    fn archive_name<T: Identifiable>(&self, ident: &T, target: &PackageTarget) -> PathBuf;
     fn packages_path(&self) -> PathBuf;
     fn write_archive(filename: &PathBuf, body: Vec<u8>) -> Result<PackageArchive>;
 }
@@ -85,11 +85,14 @@ pub trait DepotUtil {
 impl DepotUtil for config::Config {
     // Return a PackageArchive representing the given package. None is returned if Builder
     // doesn't have an archive for the given package.
+    // TODO: This doesn't actually get called by the depot server process any longer
+    // It might be good to validate whether or not this is even required anymore
+    // and if it's not, remove the heck out of it!
     fn archive<T>(&self, ident: &T, target: &PackageTarget) -> Option<PackageArchive>
     where
         T: Identifiable,
     {
-        let file = self.archive_path(ident, target);
+        let file = self.archive_name(ident, target);
         match fs::metadata(&file) {
             Ok(_) => Some(PackageArchive::new(file)),
             Err(_) => None,
@@ -98,8 +101,8 @@ impl DepotUtil for config::Config {
 
     // Return a formatted string representing the filename of an archive for the given package
     // identifier pieces.
-    fn archive_path<T: Identifiable>(&self, ident: &T, target: &PackageTarget) -> PathBuf {
-        self.packages_path().join(format!(
+    fn archive_name<T: Identifiable>(&self, ident: &T, target: &PackageTarget) -> PathBuf {
+        PathBuf::from(format!(
             "{}-{}-{}-{}-{}-{}.hart",
             ident.origin(),
             ident.name(),
--- a/components/builder-depot/src/server.rs
+++ b/components/builder-depot/src/server.rs
@@ -41,6 +41,7 @@ use hyper::mime::{Attr, Mime, SubLevel, TopLevel, Value};
 use iron::headers::{ContentType, UserAgent};
 use iron::middleware::BeforeMiddleware;
 use iron::request::Body;
+use iron::typemap;
 use persistent;
 use protobuf;
 use protocol::jobsrv::{JobGraphPackagePreCreate, JobGraphPackageStats, JobGraphPackageStatsGet,
@@ -52,6 +53,7 @@ use regex::Regex;
 use router::{Params, Router};
 use segment_api_client::SegmentClient;
 use serde_json;
+use tempfile::{tempdir_in, TempDir};
 use url;
 use uuid::Uuid;
 
@@ -789,8 +791,10 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
     };
 
     // Create a temp file at the archive location
+    let dir = tempdir_in(depot.packages_path()).expect("Unable to create a tempdir!");
+    let file_path = dir.path();
     let temp_name = format!("{}.tmp", Uuid::new_v4());
-    let temp_path = parent_path.join(temp_name);
+    let temp_path = parent_path.join(file_path).join(temp_name);
 
     let mut archive = match write_archive(&temp_path, &mut req.body) {
         Ok(a) => a,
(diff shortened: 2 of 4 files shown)


thesentinels:

Thank you for contributing!

eeyun:
One bug remaining in this, will have to get to it tomorrow. The remaining bug is that on download an extra copy of the requested package gets left hanging around in the `archive_path`. Upload and download functionality are all in and tested working.

components/builder-depot/src/server.rs
```diff
@@ -1257,24 +1260,17 @@ fn download_package(req: &mut Request) -> IronResult<Response> {
 
     match route_message::<OriginPackageGet, OriginPackage>(req, &ident_req) {
         Ok(package) => {
-            let temp_path = depot.archive_path(package.get_ident(), &target);
+            let dir = tempdir().expect("Unable to create a tempdir!");
```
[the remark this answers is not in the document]
chefsalim: Hmm yeah that's a good point - you may have to use a different pattern - eg, see https://github.com/habitat-sh/habitat/blob/master/components/common/src/ui.rs#L541

components/builder-depot/src/server.rs
```diff
@@ -1257,24 +1260,17 @@ fn download_package(req: &mut Request) -> IronResult<Response> {
...
             match s3handler.download(&temp_path, &temp_ident, &target) {
-                Ok(_target) => {
-                    Response::with(status::Ok);
-                    _target
-                }
-                Err(_) => return Ok(Response::with(status::NotFound)),
-            }
-            if let Some(archive) = depot.archive(package.get_ident(), &target) {
-                match fs::metadata(&archive.path) {
+                Ok(archive) => match fs::metadata(&archive.path) {
```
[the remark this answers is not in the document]
eeyun: This is mostly just to check that the full file was downloaded into the temporary path before firing it off in the response to the client. If that feels unnecessary that should also be a pretty trivial change!

components/builder-depot/src/server.rs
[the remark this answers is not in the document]
raskchanky: Your assumption about the variable being in scope is correct - it's in scope for the duration of the match block.  The problem I was referring to involves this sequence of events:
* Client requests hart file
* Depot downloads hart file from S3 and stores it in temp dir
* Depot constructs an HTTP response, including headers for content disposition and file name of the hart file (note that the body of the HTTP response does not contain the contents of the hart file at this point)
* Depot returns the HTTP response
* Immediately after the HTTP response is returned from this function, the temp dir goes out of scope and is deleted

At this point the client has an HTTP response that points to a file that doesn't exist any longer.  I'm not certain of the mechanism by which the client and/or server takes the HTTP response, recognizes that it's supposed to contain a file and puts that file contents into the body of the HTTP response, but it's not present in our code, that I can see.

I think the safer, albeit more laborious, approach is similar to what Salim linked, where we generate a directory with a UUID for a name, drop the file from S3 in there, and download from there.  At some point, we need a way to clean that up, but honestly I don't think that has to be part of this PR.  I'm imagining some kind of background thread that watches for those directories and removes them after X amount of time.  Right now though, it's sufficient to just leave a comment in the code that stuff isn't getting cleaned up and open an issue to circle back to it.

I think getting the download stuff more bulletproof from a concurrency perspective trumps cleanup at this point.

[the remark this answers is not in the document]
eeyun: Ok, sorry for all the questions I just want to make sure I have a full understanding - I _thought_ that putting the body of the file into the http response was happening here:
https://github.com/habitat-sh/builder/blob/c463dc8df24af89fec1e9c3510c1d2e15a4e18c9/components/builder-depot/src/server.rs#L1270

We validate that the file exists in the tempdir, and then we pass the archive through to the response function, I just kind of figured since this was inside the match block that this would all be occurring before dropping out of scope:
https://github.com/habitat-sh/builder/blob/c463dc8df24af89fec1e9c3510c1d2e15a4e18c9/components/builder-depot/src/server.rs#L2532

I'd prefer to not go the path of generating a dir based on uuid from the name of the file because of the situation @fnichol was detailing during the outage of like having multiple requests for the same thing and potentially clobbering each other. But we could definitely just use a more durable tempdir type that doesn't get removed unless it's explicitly closed in the code. But then we could end up filling that disk up realllllllly fast.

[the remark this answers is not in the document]
eeyun: Oh I think I misunderstood your comment, sorry about that @raskchanky I see what you're saying.

eeyun:
To explain why this now works. @fnichol and I tore apart the iron docs and source code until we found the extensions primitive for the iron response type. It's intended use is as extensible storage for data associated with the type. By injecting our directory into the extension it should guarantee us that the data continues to exist until the response is fully written and sent!

components/builder-depot/src/lib.rs
```diff
@@ -98,8 +98,8 @@ impl DepotUtil for config::Config {
 
     // Return a formatted string representing the filename of an archive for the given package
     // identifier pieces.
-    fn archive_path<T: Identifiable>(&self, ident: &T, target: &PackageTarget) -> PathBuf {
-        self.packages_path().join(format!(
```
raskchanky: I don't understand this change.  Don't we lose the root directory by removing `self.packages_path()`?  What is the root directory of the new `PathBuf` that's returned?

components/builder-depot/src/server.rs
```diff
@@ -865,7 +869,7 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
         }
     }
 
-    let filename = depot.archive_path(&ident, &target_from_artifact);
```
raskchanky: Ohh, is this why the change was made above to `archive_name`?  Because you specifically don't want it to return an absolute path any more?  Because you're getting your absolute path here, from `file_path`?

components/builder-depot/src/lib.rs
eeyun: Yep, so basically we need this function to return _only_ the hartfile name as a PathBuf so it can be cleanly appended to the tempdir when a download or upload occurs. Before, this function would return the harfile name appended to a specific path which was less composable and added some dirtiness when creating that tempdir and dropping the downloaded files in. It was effectively nesting an unnecessary pathing structure inside the tempdir.

components/builder-depot/src/lib.rs
```diff
@@ -89,7 +89,7 @@ impl DepotUtil for config::Config {
     where
         T: Identifiable,
     {
-        let file = self.archive_path(ident, target);
+        let file = self.archive_name(ident, target);
         match fs::metadata(&file) {
```
chefsalim: Is the metadata check deterministic, since the file now no longer a fully qualified path?

eeyun: I _think_ so since it's now only ever called in the context of being within a `/pkgs/tempfile`

chefsalim: I wonder if it's safer to pass in the parent path and do a fully qualified check.. or remove the metadata check (as it gets checked or will fail later anyways if the file isn't there).

components/builder-depot/src/server.rs
```diff
@@ -1257,27 +1261,19 @@ fn download_package(req: &mut Request) -> IronResult<Response> {
...
-                match fs::metadata(&archive.path) {
-                    Ok(_) => download_response_for_archive(archive),
-                    Err(_) => Ok(Response::with(status::NotFound)),
-                }
-            } else {
-                // This can happen if the package is not found in the file system for some reason
-                error!("Inconsistentcy between metadata and filesystem!");
-                Ok(Response::with(status::InternalServerError))
+            match s3handler.download(&file_path, &temp_ident, &target) {
+                Ok(archive) => match fs::metadata(&archive.path) {
```
chefsalim: This metadata check is also redundant as it is after write_archive in the d/load..

components/builder-depot/src/lib.rs
eeyun: We can't know what the parent path is anymore as the parent path is generated dynamically by the request. We just know that we will be checking a hartfile at `/pkgs/<some_tempdir>/`

eeyun: Actually I take this back, it looks like after the refactor that function never actually gets called.

PR #501, opened 2018-06-23
Update API node terraform for S3

chefsalim:
This removes the need for Terraform to attach a volume for artifacts to the API node.  

This should only be merged when the corresponding change in cloud-environments is also merged, and after validation is complete.

tenor-114339315

### DIFF
--- a/terraform/instances.tf
+++ b/terraform/instances.tf
@@ -23,25 +23,10 @@ resource "aws_instance" "api" {
     agent       = "${var.connection_agent}"
   }
 
-  tags {
-    Name          = "builder-api-${count.index}"
-    X-Contact     = "The Habitat Maintainers <humans@habitat.sh>"
-    X-Environment = "${var.env}"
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

terraform/instances.tf
```diff
@@ -23,25 +23,10 @@ resource "aws_instance" "api" {
...
-    X-ManagedBy   = "Terraform"
-  }
-}
-
-resource "null_resource" "api_provision" {
-  triggers {
-    ebs_volume = "${aws_volume_attachment.api.id}"
-  }
-
-  connection {
```
eeyun: Ok It's been a hot minute since I've written any terraform templates. Did this `connection` resource get removed intentionally?

chefsalim: Yep we no longer need a separate "api_provision" resource.. everything is now merged into the "api" resource.

PR #515, opened 2018-07-04
Add auditing for job groups.

raskchanky:
I also added some new tables and functions for package auditing, which
effectively abandons the existing tables. This felt like the right move,
though, given the need for multiple auditing tables. Having one called
just "audit" seemed like a bad idea in hindsight. We can leave the old
stuff around for awhile and delete it later on down the road.

It's also worth noting that this code hasn't been tested yet. I spent the entire day trying to get builds working in my local dev env and couldn't.  If someone has builds working in their env, maybe they can test this, or alternatively, we could test it in acceptance.


thesentinels:

Thank you for contributing!

chefsalim:
tenor-209758101

raskchanky:
@chefsalim I'll fix that

PR #520, opened 2018-07-06
Mount drive to /mnt, NOT /hab

christophermaier:
This made for unhappy VMs after a restart.

### DIFF
--- a/terraform/scripts/foundation.sh
+++ b/terraform/scripts/foundation.sh
@@ -6,7 +6,7 @@
 set -eux
 
 sudo mount /dev/xvdf /mnt
-echo '/dev/xvdf /hab     ext4   defaults 0 0' | sudo tee -a /etc/fstab
+echo '/dev/xvdf /mnt     ext4   defaults 0 0' | sudo tee -a /etc/fstab
 sudo mkdir -p /mnt/hab
 sudo ln -s /mnt/hab /hab
 


thesentinels:

Thank you for contributing!

nellshamrell:
tenor

fnichol:
A great catch!!

tenor-46156363

fnichol:
A little background here…

Our instances running Builder services are all provisioned via Terraform and the disk setup is consistent across all instance types. To deal with the "Habitat stuff" distinctly from normal "system stuff" we allocate and attach an additional EBS volume (in AWS) which works out to `/dev/xvdf` on a running system and we mount this at `/mnt`. In that mounted volume we create a `hab/` directory which is the root for all the Habitat software, service state, etc. Finally, we have a symlink at `/hab` that points to `/mnt/hab` which effectively puts all Habitat software on the additional EBS volume.

Our mistake here in the Terraform provisioning was in setting up the `/etc/fstab` entry for our additional volume. We mistakenly tried to mount it at `/hab` (which is actually the symlink) rather than at `/mnt` (which is what Terraform does for us on the first boot and remains in effect until the instance restarts).

This change updates the `/etc/fstab` entry so that the instances can be restarted and can resume work afterwards. Turns out in our case that `/mnt` and `/hab` are pretty close in concept, and character count. Anyway, a great find and fix!!

tenor-257712054

PR #525, opened 2018-07-09
Fix shellcheck lints for CI supporting scripts

baumanj:
Starting with the least risky changes, this is another bite-sized step towards https://github.com/habitat-sh/habitat/issues/4170.

This reverts the `shellcheck` suppressions in favor of actually fixing the lints.

### DIFF
--- a/support/ci/fast_pass.sh
+++ b/support/ci/fast_pass.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
-# shellcheck disable=SC2086
 #
 # Check to see what directories have been affected by a change. If directories
 # have not been affected, exit 0
@@ -37,16 +36,19 @@ else
   echo
   echo "Checking for changed files since last merge commit:"
   echo
+  # shellcheck disable=2086
   echo_indented $CHANGED_FILES
   echo
 
   echo "Among the affected files:"
   echo
+  # shellcheck disable=2086
   echo_indented $AFFECTED_FILES
   echo
 
   echo "And in the affected directories:"
   echo
+  # shellcheck disable=2086
   echo_indented $AFFECTED_DIRS
   echo
 
--- a/support/ci/lint.sh
+++ b/support/ci/lint.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
-# shellcheck disable=SC2002,SC2059,SC2086,SC2154,SC2162
 
 # Fail if there are any unset variables and whenever a command returns a
 # non-zero exit code.
@@ -14,10 +13,10 @@ fi
 info() {
   case "${TERM:-}" in
     *term | xterm-* | rxvt | screen | screen-*)
-      printf -- "   \033[1;32m${program}: \033[1;37m$1\033[0m\n"
+      printf -- "   \033[1;32m%s: \033[1;37m%s\033[0m\n" "${program}" "$1"
       ;;
     *)
-      printf -- "   ${program}: $1\n"
+      printf -- "   %s: %s\n" "${program}" "$1"
       ;;
   esac
   return 0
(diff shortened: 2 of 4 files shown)


thesentinels:

Thank you for contributing!

support/ci/fast_pass.sh
```diff
@@ -37,16 +36,19 @@ else
   echo
   echo "Checking for changed files since last merge commit:"
   echo
+  # shellcheck disable=2086
```
baumanj: See https://github.com/koalaman/shellcheck/wiki/SC2086

These have to be disabled since we need to treat these variables like arrays and pass them to `echo_indented` as separate arguments. Since they're passed in as environment variables we can't use proper bash arrays.

support/ci/lint.sh
```diff
@@ -29,7 +28,7 @@ warn() {
       >&2 echo -e "   \033[1;32m${program}: \033[1;33mWARN \033[1;37m$1\033[0m"
       ;;
     *)
-      >&2 echo "   ${pkg_name}: WARN $1"
+      >&2 echo "   ${program}: WARN $1"
```
baumanj: This was just erroneous. Thankfully `shellcheck` surfaced it.

support/ci/lint.sh
```diff
@@ -14,10 +13,10 @@ fi
 info() {
   case "${TERM:-}" in
     *term | xterm-* | rxvt | screen | screen-*)
-      printf -- "   \033[1;32m${program}: \033[1;37m$1\033[0m\n"
+      printf -- "   \033[1;32m%s: \033[1;37m%s\033[0m\n" "${program}" "$1"
```
baumanj: See https://github.com/koalaman/shellcheck/wiki/SC2059

support/ci/lint.sh
```diff
@@ -89,7 +89,7 @@ else
   info "Selecting files from Git via: '$cmd'"
 fi
 
-eval "$cmd" | while read file; do
+eval "$cmd" | while read -r file; do
```
baumanj: See https://github.com/koalaman/shellcheck/wiki/SC2162#read-without--r-will-mangle-backslashes

support/ci/lint.sh
```diff
@@ -108,7 +108,7 @@ eval "$cmd" | while read file; do
       case $rf_exit in
         0|3)
           if echo "$output" | grep -q "Diff at line " >/dev/null; then
-            warn "File $file generates a diff after running rustfmt $rf_version"
+            warn "File $file generates a diff after running rustfmt"
```
baumanj: Previously we used `$rf_version` to check the version of `rustfmt`, but now we just rely on [pinning the `rust` version in `.travis.yml`](https://github.com/habitat-sh/builder/blob/master/.travis.yml#L42).

support/ci/lint.sh
```diff
@@ -132,13 +132,13 @@ eval "$cmd" | while read file; do
   esac
 done
 
-if [[ $(cat "$failed" | wc -l) -gt 0 ]]; then
+if [[ -s "$failed" ]]; then
```
baumanj: > -s _file_
> True if file exists and has a size greater than zero.

https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html#Bash-Conditional-Expressions

support/ci/lint.sh
```diff
@@ -132,13 +132,13 @@ eval "$cmd" | while read file; do
...
   echo
   echo
   warn "Summary: One or more files failed linting:"
-  cat "$failed" | while read file; do
+  while read -r file; do
     warn "  * $file"
-  done
+  done < "$failed"
```
baumanj: https://github.com/koalaman/shellcheck/wiki/SC2002#useless-cat-consider-cmd--file---or-cmd-file---instead

chefsalim:
tenor-153542863

PR #527, opened 2018-07-11
Fix shellcheck lints for plan files

baumanj:
Continuing with the second least risky changes, this is another bite-sized step towards habitat-sh/habitat#4170.

This reverts the shellcheck suppressions in favor of actually fixing the lints.

### DIFF
--- a/components/airlock/plan.sh
+++ b/components/airlock/plan.sh
@@ -1,4 +1,4 @@
-# shellcheck disable=SC2086,SC2154,SC2155
+# shellcheck disable=SC2034
 pkg_name=airlock
 pkg_origin=habitat
 pkg_maintainer="The Habitat Maintainers <humans@habitat.sh>"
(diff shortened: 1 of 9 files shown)


thesentinels:

Thank you for contributing!

components/airlock/plan.sh
```diff
@@ -43,7 +44,8 @@ do_prepare() {
   # Used to find libgcc_s.so.1 when compiling `build.rs` in dependencies. Since
   # this used only at build time, we will use the version found in the gcc
   # package proper--it won't find its way into the final binaries.
-  export LD_LIBRARY_PATH=$(pkg_path_for gcc)/lib
+  export LD_LIBRARY_PATH
+  LD_LIBRARY_PATH=$(pkg_path_for gcc)/lib
```
baumanj: See https://github.com/koalaman/shellcheck/wiki/SC2155#declare-and-assign-separately-to-avoid-masking-return-values

PR #528, opened 2018-07-11
[BREAKING/depot] Updates due to PackageTarget changes in habitat-sh/core#37

fnichol:
Breaking config.toml update
---------------------------

There is a potential user-impacting breaking change included with this
change. That is, prior to this change an operator of a
`habitat/builder-api` service would override the default `targets`
section of their configuration with a form similar to this:

```toml

[[targets]]
architecture = "x86_64"
platform = "windows"

[[targets]]
architecture = "x86_64"
platform = "linux"
```

(Note that the above example is *exactly* the defaults for the service
and therefore it would be extremely unlikely to see this in the wild)

With this change, an operator would want to update their configuration
with a form similar to this:

```toml

targets = ["x86_64-linux", "x86_64-windows"]
```

Scope of breaking change
------------------------

Given how our software is built and configured, there are likely **no**
impacted users, including the Habitat team operating the public Builder
service. Overriding the default supported package targets is likely
under-documented and would require a very unusual use case to even
warrant this. However, the chance of impacting a running service is
non-zero, hence why this is called out.

---

Here is a high level summary of all the other resulting changes:

* Exploit the newly available iterators on `PackageIdent` and
`PackageTarget` and the `PackageIdent.archive_name_with_target()`
implementation to generate the same S3 key strings in `s3_key()`. The
behavior of `s3_key()` didn't require internal knowledge of a struct,
hence it was moved into a private module function where adding unit
tests was made easier too.
* Update `s3_key()` to return a `Result<String>` rather than a `PathBuf`
so that its successful return could be used directly as a key and so
that any relevant failures could be tracked with Rust error propagation.
* Update the default `Config` to use the newly available supported
`PackageTarget` constants for `x86_64-linux` and `x86_64-windows`.
* Update the `DepotUtil::archive_name()` trait function signature to
operate directly with `PackageIdent`s rather than with generic
`Identifiable`s to gain access to some of `PackageIdent`'s behavior. So
rather than using generics, this change uses type conversion from a
`&OriginPackage` to a `PackageIdent` and from a `&OriginPackageIdent`
to a `PackageIdent`. Several unit tests were added to test the type
conversions.

Signed-off-by: Fletcher Nichol <fnichol@nichol.ca>

### DIFF
--- a/components/builder-depot/src/backend/s3.rs
+++ b/components/builder-depot/src/backend/s3.rs
@@ -26,12 +26,12 @@
 //! ID and a secret access key.
 use std::fs::File;
 use std::io::{BufRead, BufReader, Read};
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use std::str::FromStr;
 
 use bldr_core::metrics::CounterMetric;
 use futures::{Future, Stream};
-use hab_core::package::{Identifiable, PackageArchive, PackageIdent, PackageTarget};
+use hab_core::package::{PackageArchive, PackageIdent, PackageTarget};
 use iron::typemap::Key;
 use metrics::Counter;
 use rusoto::{credential::StaticProvider, reactor::RequestDispatcher, Region};
@@ -82,29 +82,6 @@ impl S3Handler {
         }
     }
 
-    // Helper function for programmatic creation of
-    // the s3 object key
-    fn s3_key<T>(&self, ident: &T, target: &PackageTarget) -> PathBuf
-    where
-        T: Identifiable,
-    {
-        Path::new(ident.origin())
-            .join(format!("{}", ident.name()))
-            .join(format!("{}", ident.version().unwrap()))
-            .join(format!("{}", ident.release().unwrap()))
-            .join(format!("{}", target.architecture))
-            .join(format!("{}", target.platform))
-            .join(format!(
-                "{}-{}-{}-{}-{}-{}.hart",
-                ident.origin(),
-                ident.name(),
-                ident.version().unwrap(),
-                ident.release().unwrap(),
-                target.architecture,
-                target.platform
-            ))
-    }
-
     // This function checks whether or not the
     // configured bucket exists in the configured
     // backend.
@@ -149,10 +126,10 @@ impl S3Handler {
         target: &PackageTarget,
     ) -> Result<()> {
         Counter::UploadRequests.increment();
-        let key = self.s3_key(ident, target).to_string_lossy().into_owned();
+        let key = s3_key(ident, target)?;
         let file = File::open(hart)?;
         let path_attr = &hart.clone().into_os_string().into_string().unwrap();
-        info!("S3Handler::upload request started for s3_key: {:?}", key);
+        info!("S3Handler::upload request started for s3_key: {}", key);
         let size = file.metadata().unwrap().len() as usize;
         if size < MINLIMIT {
             self.single_upload(key, file, &path_attr)
@@ -169,7 +146,7 @@ impl S3Handler {
     ) -> Result<PackageArchive> {
         Counter::DownloadRequests.increment();
         let mut request = GetObjectRequest::default();
-        let key = self.s3_key(ident, target).to_string_lossy().into_owned();
+        let key = s3_key(ident, target)?;
         request.bucket = self.bucket.to_owned();
         request.key = key;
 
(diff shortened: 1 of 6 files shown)


thesentinels:

Thank you for contributing!

fnichol:
References habitat-sh/habitat#5278

components/builder-depot/src/backend/s3.rs
```diff
@@ -310,3 +287,54 @@ pub struct S3Cli;
...
+    // are returned in case of failure
+    let hart_name = ident.archive_name_with_target(target)?;
+
+    Ok(format!(
+        "{}/{}/{}",
+        ident.iter().collect::<Vec<&str>>().join("/"),
+        target.iter().collect::<Vec<&str>>().join("/"),
+        hart_name
+    ))
+}
```
raskchanky: Is it my imagination, or is this returning a different kind of key from the function that it replaced?  It looks like the existing `s3_key` function returns a value of the form `X-X-X-X-X.hart` whereas this returns something that looks more like `X/X/X/X/X/X/X.hart`.  Is that going to break all of our artifact storage?  I don't understand why the format was changed.

chefsalim: Good catch - it does indeed look like things will break if we don't keep the exact same format.

components/builder-depot/src/config.rs
```diff
@@ -224,10 +213,8 @@ mod tests {
         assert_eq!(config.http.port, 9000);
         assert_eq!(&format!("{}", config.routers[0]), "172.18.0.2:9001");
         assert_eq!(config.targets.len(), 2);
-        assert_eq!(config.targets[0].platform, Platform::Linux);
-        assert_eq!(config.targets[0].architecture, Architecture::X86_64);
-        assert_eq!(config.targets[1].platform, Platform::Windows);
-        assert_eq!(config.targets[1].architecture, Architecture::X86_64);
+        assert_eq!(config.targets[0], target::X86_64_LINUX);
+        assert_eq!(config.targets[1], target::X86_64_WINDOWS);
```
chefsalim: There is code in builder-depot that does a target check like:
    if !depot.targets.contains(&target_from_artifact) ...

We need to either update that, or add a test here that ensures that targets cannot contain any other known target.

components/builder-depot/src/backend/s3.rs
fnichol: It's a bit odd to look at, but the former version generated a key of the form:

```
$pkg_origin/$pkg_name/$pkg_version/$pkg_release/${pkg_arch}/${pkg_sys}/${pkg_origin}-${pkg_name}-${pkg_version}-${pkg_release}-${pkg_target}.hart
```

That is, there are 3 main pieces of information:

* The package ident parts, as a path, i.e. `${pkg_ident}`
* The package target parts, as a path, i.e. `${pkg_target}`
* The .hart file name, i.e. `${pkg_artifact}`

So, putting it another way:

```
${pkg_ident}/${pkg_target}/${pkg_artifact}
```

---

The updated implementation solves this with:

* `${pkg_ident}` paths are created using an iterator over `PackageIdent`
* `${pkg_target}` paths are created using an iterator over `PackageTarget`
* `${pkg_artifact}` entry is created calling the `archive_name_with_target` method on `PackageIdent`

You can see this in the unit test provided below, where I split the lead path entries from the filenaname and joined them in the `format!` call (the line was ludicrously long otherwise).

Make sense?

fnichol: To be clear, this should preserve the exact same key format

components/builder-depot/src/config.rs
fnichol: Let me double check this code, but as we're comparing a `PackageIdent` with presumably a `Vec<PackageIdent`, the `Eq`/`PartialEq` implementations allow this to keep working. The value at those constants are internally types like `PackageTarget(Type::X86_64_Linux))` and `PackageTarget(Type::X86_64_Windows))` which is the same types in `depot.targets` (which I presume is derived from `Config`).

Anyway, I'll double check

fnichol: I'm seeing 3 call sites with the `depot.target.contains(&target)` pattern in the `server` module:

* https://github.com/habitat-sh/builder/blob/f46ac9e992d5637bacc97ba147f34421054404ce/components/builder-depot/src/server.rs#L819 in `upload_package()`
* https://github.com/habitat-sh/builder/blob/f46ac9e992d5637bacc97ba147f34421054404ce/components/builder-depot/src/server.rs#L1242 in `download_package()`
* https://github.com/habitat-sh/builder/blob/f46ac9e992d5637bacc97ba147f34421054404ce/components/builder-depot/src/server.rs#L2492 in `download_package_from_upstream_depot()`

In all 3 places, it's checking in `depot.targets` which is a read-only borrow of `Config` to get the `Vec<PackageTarget>` vec and uses [`contains`](https://doc.rust-lang.org/std/primitive.slice.html#method.contains) to get a `true`/`false` response. As the underlying type `PackageTarget` hasn't changed (only the internals), this is all good and continues to work 100% as before.

Yay!

components/builder-depot/src/backend/s3.rs
chefsalim: 👍

components/builder-depot/src/config.rs
chefsalim: 👍

fnichol:
Okay!

Issue #517, opened 2018-07-06
Builder API / DB load profiling/tuning


chefsalim:
We have been seeing the database get under duress with an increase in some front end calls (eg https://app.zenhub.com/workspace/o/habitat-sh/builder/issues/519)

Initial observations (via the pg_activity tool) showed that count_unique_graph_packages_v1 shows up as one of the top calls when the DB is under stress. This is an expensive call that we should look at either tuning, or caching results on the front end.

There may be other calls that we want to tune as well.  For example, some APIs calls have a high fanout level that we may want to look at refactoring as well.

cnunciato:
The regular expression in this query is probably worth looking into as a possible culprit: https://github.com/habitat-sh/builder/blob/1e70671da2ba147a184e291d20a6855eed89ed15/components/builder-jobsrv/src/migrations/2017-12-13-173340_initial_jobsrv_migration/up.sql#L388

PR #529, opened 2018-07-12
Tune some queries to be a bit faster

raskchanky:
It was difficult to come up with a comprehensive set of slow queries to test, because we don't get enough activity for `pg_activity` to show much, unless the db is under a lot of strain, and it hasn't been lately.

One way to track this over time would be to install the `pg_stat_statements` extension, but I didn't do that here.  Instead, I mostly tackled the queries that were represented in the GH issue.

There are 2 main changes worth noting here:

* Switching from matching via a regular expression to using `LIKE` resulted in roughly a 2x speed increase
* Rather than converting from a string to an array on the fly for every record in a result set, creating the array column and pre-populating it ahead of time, and using that, resulted in roughly a 10x speed increase.

Closes https://github.com/habitat-sh/builder/issues/517

### DIFF
--- a/components/builder-jobsrv/src/data_store.rs
+++ b/components/builder-jobsrv/src/data_store.rs
@@ -387,7 +387,7 @@ impl DataStore {
         let conn = self.pool.get_shard(0)?;
 
         let rows = conn.query(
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

PR #536, opened 2018-07-16
Sa/build opt in

chefsalim:
This change adds an "Auto Build" option to the Builder project settings. By default, the auto build option is set to Off.  If auto-build is off, Builder will not automatically kick off builds of the package if one of it's reverse dependencies re-builds, or if it detects a commit change via the Github webhook/bldr.toml path. Essentially, only manual builds via the Web UI or the hab client will be conducted.  

tenor-249913659

### DIFF
--- a/components/builder-originsrv/src/migrations/2018-06-29-222627_auto_build_opt_in/up.sql
+++ b/components/builder-originsrv/src/migrations/2018-06-29-222627_auto_build_opt_in/up.sql
@@ -1,0 +1,69 @@
+ALTER TABLE IF EXISTS origin_projects ADD COLUMN IF NOT EXISTS auto_build bool NOT NULL DEFAULT true;
+
+CREATE OR REPLACE FUNCTION update_origin_project_v4 (
+  project_id bigint,
+  project_origin_id bigint,
+  project_package_name text,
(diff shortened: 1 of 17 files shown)


thesentinels:

Thank you for contributing!

components/builder-originsrv/src/migrations/2018-06-29-222627_auto_build_opt_in/up.sql
```diff
@@ -0,0 +1,69 @@
+ALTER TABLE IF EXISTS origin_projects ADD COLUMN IF NOT EXISTS auto_build bool NOT NULL DEFAULT true;
```
raskchanky: I thought this was supposed to be off by default?

chefsalim: Yes that was the original plan.. however in the interest of not breaking any existing build workflows (eg, all core plans auto builds would immediately stop), it seems better to default it to true in the DB, and then have a migration path for turning off unneeded builds.  Note that the UI will default auto build to Off as default, so for new plan connections we will have the more conservative behavior.

Issue #494, opened 2018-06-20
Disable nginx caching


baumanj:
A zero-byte file cached by nginx was instrumental to the builder outage on Monday, June 18th. Since our hit rate with Fastly caching is very high, it's not clear if we really benefit from the added complexity that caching in nginx gives.

Also, the nginx cache depends upon the HTTP user agent header, which seems questionable. I propose we try out disabling the nginx cache unless we can show a good argument that it provides value. We've already seen that it can exact a cost.

chefsalim:
Couple of action items:
1. See if we can determine root cause - eg, is there a race condition we've missed
2. If not, add a config that turns off nginx caching, and we can experiment with having it off

PR #545, opened 2018-07-20
Nginx caching

raskchanky:
This PR is mostly about tuning the nginx cache.

* There's now a config directive that can be set to `false` to disable caching completely.
* If caching is enabled, and we want to clear the cache, we can do that with e.g. `curl -v -X PURGE -D - "https://bldr.habitat.sh/*"`
* The above command will only work when run from the API node itself, so internet randos can't purge our caches for us.
* Only 200, 301 and 302 responses are cached, and only for 10 minutes.  Previously, everything was cached indefinitely.

I don't have a ton of experience modifying nginx configuration files, so if I messed something up, I'm not surprised.  Please do let me know.

Unrelated, but I noticed an extraneous `BufWriter` in the rust code to write files that wasn't necessary, so I removed that.  I'm not sure if that was related to the zero byte file issue or not, but I'm always a fan of removing code if possible.

Finally, the `default.toml` changes are almost entirely formatting.  The only real change of substance is the `enable_caching` line in the `nginx` section.

Closes https://github.com/habitat-sh/builder/issues/494

### DIFF
--- a/components/builder-api-proxy/habitat/default.toml
+++ b/components/builder-api-proxy/habitat/default.toml
@@ -1,57 +1,58 @@
-app_url                 = "https://bldr.habitat.sh"
-cookie_domain           = ""
-demo_app_url            = "https://www.habitat.sh/demo/build-system/steps/1/"
-docs_url                = "https://www.habitat.sh/docs"
-enable_publisher_amazon = false
-enable_publisher_azure  = false
-enable_publisher_docker = false
-enable_builder          = false
-environment             = "production"
-hosted                  = false
-source_code_url         = "https://github.com/habitat-sh/habitat"
-status_url              = "https://status.habitat.sh/"
-tutorials_url           = "https://www.habitat.sh/learn"
-use_gravatar            = true
-www_url                 = "https://www.habitat.sh"
+app_url                   = "https://bldr.habitat.sh"
+cookie_domain             = ""
+demo_app_url              = "https://www.habitat.sh/demo/build-system/steps/1/"
+docs_url                  = "https://www.habitat.sh/docs"
+enable_publisher_amazon   = false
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

components/builder-api-proxy/habitat/default.toml
```diff
@@ -1,57 +1,58 @@
...
-enable_publisher_docker = false
-enable_builder          = false
-environment             = "production"
-hosted                  = false
-source_code_url         = "https://github.com/habitat-sh/habitat"
-status_url              = "https://status.habitat.sh/"
-tutorials_url           = "https://www.habitat.sh/learn"
-use_gravatar            = true
-www_url                 = "https://www.habitat.sh"
+app_url                   = "https://bldr.habitat.sh"
```
chefsalim: Convert over the tabs back to spaces in the toml files?

raskchanky: I used a plugin to do the alignment.  I had no idea it used tabs instead of spaces.  How can you tell the difference from the diff?

raskchanky: I went into this file in my editor, and I can arrow through the spaces one by one, as well as traverse the lines vertically, maintaining the same column position, at any point on any of these lines.  They don't behave like tabs.

raskchanky: I further configured my editor to show the ASCII character code of the character under the cursor, and ran through all these lines, and indeed, they are spaces, not tabs.

chefsalim: ah ok cool.. for some reason I assumed the bold whitespace was showing tabs, but looks like that is not the case

PR #550, opened 2018-07-25
Remove nginx cache purge option

chefsalim:
proxy_cache_purge is only available in Nginx Plus version, so we can't use it.

tenor-255530023


thesentinels:

Thank you for contributing!

PR #556, opened 2018-07-27
Migrate github-api-client and segment-api-client to reqwest

chefsalim:
This change migrates the github-api-client and segment-api-client to use the reqwest crate instead of hyper / habitat_http_client. It is part of a larger dependency cleanup effort that will simplify and update our dependencies, as well as remove the circular dependency we currently have between the habitat and builder repos.

tenor-105155218


thesentinels:

Thank you for contributing!

PR #557, opened 2018-07-27
Update JWT to use latest frank_jwt crate

chefsalim:
This migrates us away from a one-off, year old JWT crate to the current latest version of the original JWT crate.

tenor-19364684

### DIFF
--- a/components/github-api-client/Cargo.toml
+++ b/components/github-api-client/Cargo.toml
@@ -6,7 +6,7 @@ workspace = "../../"
 
 [dependencies]
 base64 = "*"
-frank_jwt = { git = "https://github.com/habitat-sh/frank_jwt", branch = "habitat" }
+frank_jwt = "*"
 reqwest = "*"
 log = "*"
 regex = "*"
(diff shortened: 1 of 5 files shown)


components/github-api-client/Cargo.toml
```diff
@@ -6,13 +6,13 @@ workspace = "../../"
...
 [dependencies]
 base64 = "*"
-frank_jwt = { git = "https://github.com/habitat-sh/frank_jwt", branch = "habitat" }
+frank_jwt = "*"
 reqwest = "*"
 log = "*"
 regex = "*"
-serde = ">= 1.0"
-serde_derive = ">= 1.0"
-serde_json = ">= 1.0"
```
raskchanky: I'm curious why these constraints were removed.  I put them in there on purpose awhile back because certain cargo changes were reverting serde to 0.8 versions.  I can't see a scenario where we'd want to use a version of serde < 1.0.

chefsalim: Ah it is a bit of leftover cleanup from a different set of changes (that should clean up the reversions)... I'll remove it from this one as I think it's a no-op either way.

PR #560, opened 2018-07-30
Adds validation of minio write to s3 uploads

eeyun:
OK. This adds some validation to the s3 upload functions. After the upload future returns it's status we chain a combinator onto the returned future - we take the configured s3 bucket and key and check that the object exists before the upload function returns it's status.

### DIFF
--- a/components/builder-depot/src/backend/s3.rs
+++ b/components/builder-depot/src/backend/s3.rs
@@ -24,6 +24,7 @@
 //!
 //! Currently the S3Handler must be configured with both an access key
 //! ID and a secret access key.
+use std::fmt::Display;
 use std::fs::File;
 use std::io::{BufRead, BufReader, Read};
 use std::path::PathBuf;
@@ -35,11 +36,9 @@ use hab_core::package::{PackageArchive, PackageIdent, PackageTarget};
 use iron::typemap::Key;
 use metrics::Counter;
 use rusoto::{credential::StaticProvider, reactor::RequestDispatcher, Region};
-use rusoto_s3::{
-    CompleteMultipartUploadRequest, CompletedMultipartUpload, CompletedPart, CreateBucketRequest,
-    CreateMultipartUploadRequest, GetObjectRequest, PutObjectRequest, S3, S3Client,
-    UploadPartRequest,
-};
+use rusoto_s3::{CompleteMultipartUploadRequest, CompletedMultipartUpload, CompletedPart,
+                CreateBucketRequest, CreateMultipartUploadRequest, GetObjectRequest,
+                HeadObjectRequest, PutObjectRequest, S3, S3Client, UploadPartRequest};
 use time::PreciseTime;
 
 use config::{S3Backend, S3Cfg};
@@ -103,6 +102,24 @@ impl S3Handler {
         }
     }
 
+    // This function checks whether an uploaded file
+    // exists in the configured s3 bucket. It should
+    // only get called from within an upload future.
+    fn object_exists(&self, object_key: &str) -> Result<()> {
+        let mut request = HeadObjectRequest::default();
+        request.bucket = self.bucket.clone();
+        request.key = object_key.to_string();
+
+        match self.client.head_object(&request).sync() {
+            Ok(object) => {
+                info!("Verified {} was written to minio!", object_key);
+                debug!("Head Object check returned: {:?}", object);
+                Ok(())
+            }
+            Err(e) => Err(Error::HeadObject(e)),
+        }
+    }
+
     pub fn create_bucket(&self) -> Result<()> {
         let mut request = CreateBucketRequest::default();
         request.bucket = self.bucket.clone();
@@ -121,20 +138,25 @@ impl S3Handler {
 
     pub fn upload(
         &self,
-        hart: &PathBuf,
+        hart_path: &PathBuf,
         ident: &PackageIdent,
         target: &PackageTarget,
     ) -> Result<()> {
         Counter::UploadRequests.increment();
         let key = s3_key(ident, target)?;
-        let file = File::open(hart)?;
-        let path_attr = &hart.clone().into_os_string().into_string().unwrap();
+        let file = File::open(hart_path)?;
+
         info!("S3Handler::upload request started for s3_key: {}", key);
+
         let size = file.metadata().unwrap().len() as usize;
+        let fqpi = hart_path.clone().into_os_string().into_string().unwrap();
+
         if size < MINLIMIT {
-            self.single_upload(key, file, &path_attr)
+            self.single_upload(&key, file, fqpi)
+                .and_then(move |_| self.object_exists(&key))
         } else {
-            self.multipart_upload(key, file, &path_attr)
+            self.multipart_upload(&key, file, fqpi)
+                .and_then(move |_| self.object_exists(&key))
         }
     }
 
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/builder-depot/src/backend/s3.rs
```diff
@@ -103,6 +102,25 @@ impl S3Handler {
...
+    // exists in the configured s3 bucket. It should
+    // only get called from within an upload future.
+    fn object_exists(&self, object_key: &str) -> Result<()> {
+        let mut request = ListObjectsRequest::default();
+        request.bucket = self.bucket.clone();
+        request.prefix = Some(object_key.to_string());
+
+        match self.client.list_objects(&request).sync() {
+            Ok(object_list) => match object_list {
+                object_key => {
```
raskchanky: I'm confused.  If I'm reading the docs correctly, `object_list` is of type `rusoto_s3::ListObjectsOutput`.  Then you match against that, comparing it to `object_key`, which is of type `&str`.  How does this ever succeed?

components/builder-depot/src/backend/s3.rs
```diff
@@ -128,13 +146,19 @@ impl S3Handler {
...
         let file = File::open(hart)?;
-        let path_attr = &hart.clone().into_os_string().into_string().unwrap();
+
         info!("S3Handler::upload request started for s3_key: {}", key);
+
         let size = file.metadata().unwrap().len() as usize;
         if size < MINLIMIT {
-            self.single_upload(key, file, &path_attr)
+            self.single_upload(&key, file, hart)
+//                .map_err(|e| Error::PackageUpload)
```
raskchanky: Probably want to uncomment the error handling?

components/builder-depot/src/backend/s3.rs
chefsalim: This match seems a bit redundant - is it needed (does not seem like there is a case for object key not in the list)..?

components/builder-depot/src/backend/s3.rs
chefsalim: Remove..

components/builder-depot/src/backend/s3.rs
```diff
@@ -128,13 +146,19 @@ impl S3Handler {
...
+                .and_then(move |_| {
+                    self.object_exists(&key)
```
chefsalim: I wonder if this check should be feature flagged so we can turn these checks on and off as we want... (pulling in the usage of the feature  crate into builder depot will also be helpful for other future work btw)

components/builder-depot/src/backend/s3.rs
```diff
@@ -128,13 +146,19 @@ impl S3Handler {
...
         let key = s3_key(ident, target)?;
         let file = File::open(hart)?;
-        let path_attr = &hart.clone().into_os_string().into_string().unwrap();
+
         info!("S3Handler::upload request started for s3_key: {}", key);
+
         let size = file.metadata().unwrap().len() as usize;
         if size < MINLIMIT {
-            self.single_upload(key, file, &path_attr)
+            self.single_upload(&key, file, hart)
```
chefsalim: Might be nice to also rename 'hart' to 'hart_path' or something like that..

components/builder-depot/src/backend/s3.rs
```diff
@@ -169,45 +193,53 @@ impl S3Handler {
...
+                        fqpi,
+                        start_time.to(end_time)
+                    );
+                    return Ok(());
+                }
+                Err(e) => {
+                    Counter::UploadFailures.increment();
+                    warn!("Upload failed for {}: ({:?})", fqpi, e);
+                    return Err(Error::PackageUpload(e));
+                }
```
raskchanky: It's possible I'm misreading this, but off the cuff, it looks like this might be introducing a subtle bug.  The rusoto docs say that `sync()` blocks the current thread until the future has resolved, which effectively gives you synchronous behavior.

I haven't read the docs for Rust futures, but in other languages, calling `then` on a future registers a callback that will be run whenever that future happens to complete, whether that's 5 seconds or 5 minutes from now.  It returns immediately, so the executing thread won't block.

So before, when this function returned, we were confident that all of this code had finished executing, but now, that's not the case.

Am I reading this right?

components/builder-depot/src/backend/s3.rs
```diff
@@ -169,45 +193,53 @@ impl S3Handler {
         }
     }
 
-    fn single_upload(&self, key: String, hart: File, path_attr: &str) -> Result<()> {
+    fn single_upload(&self, key: &str, hart: File, path_attr: &PathBuf) -> Result<()> {
```
chefsalim: Consider changing path_attr type to AsRef<PathBuf> so anything convertible to a Pathbuf will just work..

components/builder-depot/src/backend/s3.rs
```diff
@@ -169,45 +193,53 @@ impl S3Handler {
...
-                );
-                Ok(())
-            }
-            Err(e) => {
-                Counter::UploadFailures.increment();
-                warn!("Upload failed for {}: ({:?})", &path_attr, e);
-                Err(Error::PackageUpload(e))
+        let fqpi = path_attr.clone().into_os_string().into_string().unwrap();
+
+        let put_future = self.client.put_object(&request).then(move |result| {
```
chefsalim: Is the change to use futures required?

components/builder-depot/src/backend/s3.rs
eeyun: It's not actually matching against the object list, we actually pass object key into the request function as an argument which forces the search to search for that specific file. This is more or less an unwrap so that we can log the value once we've resolved the initial map successfully.

components/builder-depot/src/backend/s3.rs
eeyun: You are correct, the `then` call chains a computation on to the returned future - the `then` combinator will get executed whether the returned future is a failure or a success.

components/builder-depot/src/backend/s3.rs
eeyun: It's using futures either way, this just moves us closer to having our uploads handled asynchronously. If you'd prefer we keep `sync`ing it should be trivial to back this bit out.

elliott-davis: Since this is the only future we are waiting on when we run the reactor they are both doing essentially the same thing.

components/builder-depot/src/backend/s3.rs
```diff
@@ -103,6 +102,25 @@ impl S3Handler {
...
 
+    // This function checks whether an uploaded file
+    // exists in the configured s3 bucket. It should
+    // only get called from within an upload future.
+    fn object_exists(&self, object_key: &str) -> Result<()> {
+        let mut request = ListObjectsRequest::default();
+        request.bucket = self.bucket.clone();
+        request.prefix = Some(object_key.to_string());
+
+        match self.client.list_objects(&request).sync() {
```
elliott-davis: Would it be better to use [head_object](https://rusoto.github.io/rusoto/rusoto_s3/trait.S3.html#tymethod.head_object) here?

components/builder-depot/src/backend/s3.rs
eeyun: It's not the only thing we're waiting on, we've got a whole bunch of stuff in this module that uses sync this is just the first couple things I've removed the sync from. Do you think it's better to just phone it in and remove this change?

components/builder-depot/src/backend/s3.rs
raskchanky: Then I don't think this is going to be what we want.

components/builder-depot/src/backend/s3.rs
```diff
@@ -103,6 +102,24 @@ impl S3Handler {
...
+        request.bucket = self.bucket.clone();
+        request.key = object_key.to_string();
+
+        match self.client.head_object(&request).sync() {
+            Ok(object) => {
+                info!("Verified {} was written to minio!", object_key);
+                debug!("Head Object check returned: {:?}", object);
+                Ok(())
+            }
+            Err(e) => Err(Error::HeadObject(e)),
```
chefsalim: If object verification fails, we don't seem to be logging anything anywhere or failing the upload later on..?

chefsalim: ah ok, looks like the and_then result will get bubbled up.. probably would be good to log a warning message here as well

PR #561, opened 2018-07-31
Fix a minor issue with finding the files to run shellcheck against

baumanj:
The builder version of https://github.com/habitat-sh/habitat/pull/5396

Pretty much all the changes are just double-quotes to suppress unwanted expansion. There is one legit bug (occurring in several places) that I've called out with a comment.

### DIFF
--- a/.studiorc
+++ b/.studiorc
@@ -235,7 +235,7 @@ generate_bldr_keys() {
   keys=( $(find /hab/cache/keys -name "bldr-*.pub") )
 
   if [ "${#keys[@]}" -gt 0 ]; then
-    KEY_NAME=$(echo $keys[0] | grep -Po "bldr-\d+")
+    KEY_NAME=$(echo "${keys[0]}" | grep -Po "bldr-\d+")
     echo "Re-using existing builder key: $KEY_NAME"
   else
     KEY_NAME=$(hab user key generate bldr | grep -Po "bldr-\d+")
--- a/components/builder-api/habitat/hooks/run
+++ b/components/builder-api/habitat/hooks/run
@@ -1,17 +1,15 @@
-#!/bin/sh
-# shellcheck disable=1083
+#!/bin/bash
 
-
-export HOME={{pkg.svc_data_path}}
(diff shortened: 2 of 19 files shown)


thesentinels:

Thank you for contributing!

.studiorc
```diff
@@ -235,7 +235,7 @@ generate_bldr_keys() {
   keys=( $(find /hab/cache/keys -name "bldr-*.pub") )
 
   if [ "${#keys[@]}" -gt 0 ]; then
-    KEY_NAME=$(echo $keys[0] | grep -Po "bldr-\d+")
+    KEY_NAME=$(echo "${keys[0]}" | grep -Po "bldr-\d+")
```
baumanj: This syntax was just wrong. You always need braces to access an array variable:
```bash
$ arr=(one two three)
$ echo $arr[0]
one[0]
$ echo ${arr[0]}
one
```
The `-o` flag to `grep` just happened to save us.

components/builder-api/habitat/hooks/run
```diff
@@ -1,15 +1,17 @@
...
+# shellcheck disable=1083
+
 
 export HOME={{pkg.svc_data_path}}
 export RUST_LOG={{cfg.log_level}}
 export RUST_BACKTRACE=1
 pkg_svc_run="bldr-api start -c {{pkg.svc_config_path}}/config.toml"
 
-if [ "\$(whoami)" = "root" ]; then
+if [ "$(whoami)" = "root" ]; then
```
baumanj: This is a bug. It also occurs several other places:
```
$ whoami
jbauman
$ if [ "$(whoami)" = "jbauman" ]; then echo "you're jbauman"; else echo "you're not jbuman"; fi
you're jbauman
$ if [ "\$(whoami)" = "jbauman" ]; then echo "you're jbauman"; else echo "you're not jbuman"; fi
you're not jbuman
```

baumanj:
Looks like there are some issues here. I'm sorting them out.

Issue #562, opened 2018-07-31
Upstream not storing packages in minio


chrisortman:
What I think is happening is that if a package search is triggered as a result of an upload the upstream package does not make it into minio.

I'm looking at https://github.com/habitat-sh/builder/blob/master/components/builder-depot/src/server.rs#L2496 and I don't see where S3 is getting called.

The way I test this is to delete `/hab/{pkgs,cache,sup,studios}` and `/hab/svc/builder-*`
Then rerun the provision script setting log_level to debug and upstream_depot to public builder

Then I create a _core_ origin and from a separate machine do

`hab pkg install -u https://my.builder.url core/attr` 

If I wait a little bit the package will show up in the UI (and I can see the database insert in the logs) but not in minio and it will not be installable.

I can see in the logs that it looks like it is trying to move the .hart file to the old storage location

```
Jul 31 09:41:10 emrb-appdev02.icts.uiowa.edu hab[31366]: builder-api.default(O): DEBUG 2018-07-31T14:41:10Z: habitat_depot_client: Writing to /hab/svc/builder-api/data/pkgs/core-attr-2.4.47-20180608091914-x86_64-linux.hart.tmp-KIF4JhWh
Jul 31 09:41:10 emrb-appdev02.icts.uiowa.edu hab[31366]: builder-api.default(O): DEBUG 2018-07-31T14:41:10Z: habitat_depot_client: Moving /hab/svc/builder-api/data/pkgs/core-attr-2.4.47-20180608091914-x86_64-linux.hart.tmp-KIF4JhWh to /hab/svc/builder-api/data/pkgs/core-attr-2.4.47-20180608091914-x86_64-linux.hart
```

chefsalim:
Hmm, it does look like the S3 upload is outside of the function we are using to process the upstream downloads, so this looks like it may indeed be a missed case in the S3 migration.

PR #564, opened 2018-08-01
Fix missing minio upload for upstream downloads

chefsalim:
This change fixes a missing call to minio upload when the upstream syncs down a package. This causes the package metadata to get created, but the package artifact itself is missing from mino.  Resolves https://github.com/habitat-sh/builder/issues/562

tenor-114648496


thesentinels:

Thank you for contributing!

raskchanky:
Great catch.

PR #566, opened 2018-08-01
Fix JWT token for github app

chefsalim:
Fixing a regression in JWT usage for github apps, which expects integer values for the issued and expiration timestamps, where we were encoding as strings.

tenor-225873848


thesentinels:

Thank you for contributing!

PR #573, opened 2018-08-03
Migrate jobsrv to rusoto

chefsalim:
Mo mo mo more red in the cargo.lock!  This change ports jobsrv to use rusoto instead of aws-rust-sdk, which removes a bunch of deps, as well as consolidates S3 usage to the same set of crates across jobsrv and builder depot, and streamlines the code. This also removes the only known usage of a pre-1.0 serde as a bonus :)

tenor-5729668

### DIFF
--- a/components/builder-jobsrv/src/server/log_archiver/s3.rs
+++ b/components/builder-jobsrv/src/server/log_archiver/s3.rs
@@ -27,81 +27,50 @@
 
 use std::fs::OpenOptions;
 use std::io::Read;
-use std::panic::{self, AssertUnwindSafe};
 use std::path::PathBuf;
 use std::str::FromStr;
 
-use aws_sdk_rust::aws::common::credentials::{DefaultCredentialsProvider, ParametersProvider};
-use aws_sdk_rust::aws::common::region::Region;
-use aws_sdk_rust::aws::s3::endpoint::{Endpoint, Signature};
-use aws_sdk_rust::aws::s3::object::{GetObjectRequest, PutObjectRequest};
-use aws_sdk_rust::aws::s3::s3client::S3Client;
-use extern_url;
-use hyper::client::Client as HyperClient;
+use futures::{Future, Stream};
+use rusoto::{credential::StaticProvider, reactor::RequestDispatcher, Region};
+use rusoto_s3::{GetObjectRequest, PutObjectRequest, S3, S3Client};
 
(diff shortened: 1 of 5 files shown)


components/builder-jobsrv/src/server/log_archiver/s3.rs
```diff
@@ -115,40 +84,19 @@ impl S3Archiver {
...
-            Ok(Ok(_)) => Ok(()), // normal result
-            Ok(Err(e)) => {
-                // This is a "normal", non-panicking error, e.g.,
-                // they're configured with a non-existent bucket.
-                Err(Error::JobLogArchive(job_id, e))
-            }
+        request.body = Some(buffer.as_slice().to_vec());
+
+        match self.client.put_object(&request).sync() {
+            Ok(_) => Ok(()),
```
eeyun: Is it worth adding some debug logging here for the success case?

components/builder-jobsrv/src/server/log_archiver/s3.rs
```diff
@@ -158,30 +106,17 @@ impl LogArchiver for S3Archiver {
...
-        let body = match result {
-            Ok(Ok(response)) => response.body, // normal result
-            Ok(Err(e)) => {
-                // This is a "normal", non-panicking error, e.g.,
-                // they're configured with a non-existent bucket.
-                return Err(Error::JobLogRetrieval(job_id, e));
-            }
+        let payload = self.client.get_object(&request).sync();
+        let stream = match payload {
+            Ok(response) => response.body.expect("Downloaded object is not empty"),
```
eeyun: Does `.expect()`  cause a panic when it fires ?Also would the issue here be that the body in the response _is_ empty rather than _not_ empty?

components/builder-jobsrv/src/server/log_archiver/s3.rs
chefsalim: I just left the logging mostly as I found it for now..

components/builder-jobsrv/src/server/log_archiver/s3.rs
chefsalim: Yeah it will panic - I followed basically the pattern from your builder-depot S3 backend here.  GetObjectOutput should return an optional StreamingBody, which we want to be non-empty.

PR #574, opened 2018-08-03
Update Cargo.lock to latest

chefsalim:
The grand finale (sort of) of the great Builder deps cleanup journey.. somewhat boring PR as it is mostly updated versions in the Cargo.lock, plus a couple of minor cleanups, which is exactly what we wanted :)

### DIFF
--- a/components/oauth-client/Cargo.toml
+++ b/components/oauth-client/Cargo.toml
@@ -8,9 +8,9 @@ workspace = "../../"
 [dependencies]
 log = "*"
 reqwest = "*"
-serde = ">= 1.0.0"
-serde_derive = ">= 1.0.0"
-serde_json = ">= 1.0.0"
+serde = "*"
+serde_derive = "*"
+serde_json = "*"
 
 # For metrics library
 [dependencies.builder_core]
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

components/oauth-client/Cargo.toml
```diff
@@ -8,9 +8,9 @@ workspace = "../../"
 [dependencies]
 log = "*"
 reqwest = "*"
-serde = ">= 1.0.0"
-serde_derive = ">= 1.0.0"
-serde_json = ">= 1.0.0"
+serde = "*"
+serde_derive = "*"
+serde_json = "*"
```
raskchanky: I still disagree with changing these constraints, as there will _never_ be a time that we want a version of serde that's `< 1.0`, but I'm not going to fight it.

chefsalim: You're right, its mostly just a matter of keeping things consistent across the codebase where we are using "*" as the default unless there is a very specific reason (like for hyper) where we must have a special spec.

PR #575, opened 2018-08-06
Merge builder-depot into builder-api

chefsalim:
The Builder crate killing spree continues! The Voight-Kampff test has conclusively proven that builder-depot is a replicant - it is time for it to die. This change merges in the depot crate into builder-api. The old "v1/depot" routes are available both in their original location, as well as directly in the "v1" root.  This will allow us to consolidate the complete API consistently under "v1" going forward, and deprecate the "v1/depot" paths over time.  It will also allow us to consolidate the two API client crates into one. It will also make it easier to further simplify the server code, as well as introduce new frameworks in the future. More to come...

tenor-36827581


christophermaier:
Thus begins years of speculation as to whether or not @chefsalim is unwittingly an API himself....

PR #582, opened 2018-08-08
Pass --auth to integrations for private packages

### LINKED ISSUE #353: Unable to export private package to Docker image
I have a private Habitat package that I've built and promoted to the stable channel. I'm able to `hab pkg install` it but when I try to export to a Docker image, it says "Package not found". I have exported the correct HAB_AUTH_TOKEN and the origin but no dice.

eeyun:
This should resolve #353. That being said - I'm not 100% certain if this is a safe behavior or not. I would assume that it is since we never flush this token to disk or logs or the like. But, it's probably worth being aware that the token we generate for the workspace is the token we pass through to the export command.

### DIFF
--- a/components/builder-worker/src/runner/mod.rs
+++ b/components/builder-worker/src/runner/mod.rs
@@ -461,6 +461,7 @@ impl Runner {
                     util::docker_exporter_spec(&self.workspace),
                     &self.workspace,
                     &self.config.bldr_url,
+                    &self.bldr_token,
                 ).export(streamer)?;
 
                 if !status.success() {
--- a/components/builder-worker/src/runner/docker.rs
+++ b/components/builder-worker/src/runner/docker.rs
@@ -59,15 +59,22 @@ pub struct DockerExporter<'a> {
     spec: DockerExporterSpec,
     workspace: &'a Workspace,
     bldr_url: &'a str,
+    auth_token: &'a String,
 }
 
 impl<'a> DockerExporter<'a> {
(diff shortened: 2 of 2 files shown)


thesentinels:

Thank you for contributing!

components/builder-worker/src/runner/mod.rs
```diff
@@ -461,6 +461,7 @@ impl Runner {
                     util::docker_exporter_spec(&self.workspace),
                     &self.workspace,
                     &self.config.bldr_url,
+                    &self.bldr_token,
```
elliott-davis: This is _probably_ dangerous. The builder token is a admin token that has permissions to do anything, including getting private packages from repos you're not a member of. You probably want to fetch a token scoped to the user rather than use the builder token.

eeyun: Builds themselves aren't scoped to a user so I'm not certain how we'd do that. Also, this token is stored in memory and used in the workspace anyways today - this is just using it for one more operation.

eeyun: Perhaps we need a new token generation for our workers - something that generates a throwaway token for the build from the origin the package belongs to?

chefsalim: This should be fine (for now) - it's no different than what the runner itself is doing in the other steps. The builder tokens themselves are ephemeral and expire in 2 hours.

PR #584, opened 2018-08-08
Remove sharding from the database code.

raskchanky:
This introduces several new functions to replace functions that were previously iterating over all of the shards. It removes sharding from the database queries that run, but leaves it in the config for each
service, so that routing continues to work. It also introduces a new bash script to collapse all the existing shards into the public schema.

Note that the script that I wrote only migrates data, not schema changes.  The reason for this is the first time the services start up after this code has been merged, the migrations will run against the public schema and all of the tables, functions, views, etc will all get created there automatically.

Also note that we will want to be very intentional about when we merge this PR.  If this merges, and the acceptance supervisors update themselves with this code, and then someone logs into acceptance, an account record will get created for that person in the public schema and the migrate script will error out with a constraint violation.

Please give this PR some _extra_ scrutiny.  I've successfully run this on a builder studio in my local environment without problems, but let's err on the side of caution here.

ke3od3j


thesentinels:

Thank you for contributing!

components/builder-db/src/migration.rs
```diff
@@ -17,22 +17,15 @@ use diesel::query_dsl::RunQueryDsl;
...
 
-pub fn shard_setup(conn: &PgConnection, shard_id: u32) -> Result<()> {
-    debug!("Migrating shard_{:?}", shard_id);
-    sql_query(format!("CREATE SCHEMA IF NOT EXISTS shard_{}", shard_id))
-        .execute(conn)
-        .unwrap();
-    sql_query(format!("SET SEARCH_PATH TO shard_{}", shard_id))
-        .execute(conn)
-        .unwrap();
+pub fn shard_setup(conn: &PgConnection) -> Result<()> {
```
chefsalim: Maybe rename this to just setup or setup_ids ?

components/builder-db/src/migration.rs
```diff
@@ -17,22 +17,15 @@ use diesel::query_dsl::RunQueryDsl;
...
     // TED: We have to do this here because we rely on the shard id which isn't available in raw sql
-    sql_query(format!(
+    sql_query(
         r#"CREATE OR REPLACE FUNCTION next_id_v1(sequence_id regclass, OUT result bigint) AS $$
                 DECLARE
                     our_epoch bigint := 1409266191000;
                     seq_id bigint;
                     now_millis bigint;
-                    shard_id int := {};
+                    shard_id int := 0;
```
chefsalim: Remove this line entirely, and remove the use of the shard_id in the result below?

components/builder-db/src/pool.rs
```diff
@@ -79,12 +79,7 @@ impl Pool {
         shard_id: u32,
     ) -> Result<r2d2::PooledConnection<r2d2_postgres::PostgresConnectionManager>> {
         let conn = self.inner.get().map_err(Error::ConnectionTimeout)?;
-        debug!("Switching to shard {}", shard_id);
-
-        let schema_name = format!("shard_{}", shard_id);
-        let sql_search_path = format!("SET search_path TO {}", schema_name);
-        conn.execute(&sql_search_path, &[])
-            .map_err(Error::SchemaSwitch)?;
+        debug!("Ignoring shard {} on purpose", shard_id);
```
chefsalim: Cool.. it would also be nice to not see any more shard messages in debug mode.. maybe make this a trace message?

components/builder-db/src/migration.rs
raskchanky: Will do

components/builder-db/src/pool.rs
chefsalim: One other comment - it might be good to remove the shards field from the Pool structure entirely.  This will help catch other changes - for example, there is cross-DB sync code (eg, sync packages) that is iterating over all shards that should be adjusted to not iterate.

components/builder-db/src/migration.rs
raskchanky: Ah, of course, because anything ORed with 0 is just the thing.  Will do.

components/builder-db/src/pool.rs
raskchanky: Sure

raskchanky: Ah, good point.  Will do.

chefsalim:
Overall looking very good!  One thing that I think we should tackle is the case where we can get into a bad state if this merges and the services are up.  This will be particularly important to ensure we get right for the on-prem migration case.  So the datastores should not come up for business unless there is a "migration_ok" flag somewhere in the DB.  The migration script should set that flag after successful migration.

components/builder-db/src/pool.rs
```diff
@@ -98,13 +82,7 @@ impl Pool {
 
         let shard_id = match optional_shard_id {
             Some(id) => (id % SHARD_COUNT as u64) as u32,
-            None => {
-                let mut rng = rand::thread_rng();
-                match rng.choose(&self.shards) {
-                    Some(shard) => *shard,
-                    None => 0,
-                }
-            }
```
chefsalim: Since get_shard is basically a no-op now, the hash computation and most of this function can actually go away as well... so the entire function is "self.get_shard(0)".   Since that's the case, it looks like we can safely get rid of the get_shard and get_raw functions altogether, and replace them with the Pool get function that can now just return the inner connection..

raskchanky: Will do

components/builder-db/src/migration.rs
```diff
@@ -15,34 +15,55 @@
...
-        .execute(conn)
-        .unwrap();
-    sql_query(format!("SET SEARCH_PATH TO shard_{}", shard_id))
-        .execute(conn)
-        .unwrap();
+use error::{Error, Result};
+use pool::Pool;
+
+pub fn setup_ids(conn: &PgConnection) -> Result<()> {
     // TED: We have to do this here because we rely on the shard id which isn't available in raw sql
```
chefsalim: Well, hmm.. now that shard_id is gone, we may as  well move the next_id_v1 now into the main SQL, and kill this function completely.

tools/collapse_schemas/collapse-schemas.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+# The purpose of this script is a one time data migration from all of our 128
+# schemas into the public schema.
+
+set -euo pipefail
+
+usage() {
+  echo "Usage: collapse-schemas.sh <SERVER> [migrate]"
```
chefsalim: Really minor - would you mind renaming this script to "merge-schemas.sh" or "migrate-schemas.sh" instead of "collapse-schemas.sh"?

tools/collapse_schemas/collapse-schemas.sh
```diff
@@ -0,0 +1,104 @@
...
+      elif [ "${count_map[$table]}" -ne "$public_count" ]; then
+        echo "*** PANIC ***. The accumulated record total for the $table table across all shards is ${count_map[$table]}. That does not match the total record count in the public schema, which is $public_count."
+        errors=$((errors+1))
+      else
+        echo "The accumulated record count for the $table table matches the total record count in the public schema. All good."
+      fi
+    done
+
+    if [ "$errors" -eq 0 ]; then
+      echo "The entire shard migration completed without errors. Updating db flag."
```
chefsalim: Minor - reword output to say to "schema migration" instead of "shard migration"?

components/builder-db/src/migration.rs
chefsalim: Yep, just add to the migration...

tools/collapse_schemas/collapse-schemas.sh
raskchanky: Part of the challenge here is the term "schema" is overloaded.  It can be used to refer to normal database schemas, e.g. the data definition of a table.  It can also be used to refer to a postgres specific entity that provides a logical separation between tables and functions, which is of course how our sharding works.  It feels important to me that we are careful to disambiguate between the two.  To a person that's not familiar with the intimate details of our sharding implementation, our regular database migrations are definitely "schema migrations".  I'd prefer to keep the language in this script specific to shards and not schemas, to eliminate any ambiguity.

tools/collapse_schemas/collapse-schemas.sh
raskchanky: I'll rename it `merge-shards.sh`, per my comment below.

chefsalim: Cool, that works!

tools/collapse_schemas/collapse-schemas.sh
chefsalim: 👍

tools/collapse_schemas/collapse-schemas.sh
```diff
@@ -0,0 +1,104 @@
...
+fi
+
+execute_sql() {
+  local sql="$1"
+  local server="$2"
+
+  hab pkg exec core/postgresql psql -t -U hab -h 127.0.0.1 -p 5433 -c "$sql" "builder_$server"
+}
+
+process_shards() {
```
chefsalim: One more thing to think about - since this script is not idempotent, maybe check here to see if the migration has already been run (check the flag) and print a nice message and exit immediately if so.

components/builder-db/src/migration.rs
raskchanky: Will do

tools/collapse_schemas/collapse-schemas.sh
raskchanky: Yeah, good call

components/builder-db/src/migration.rs
raskchanky: Turns out this doesn't work.  Moving this function into the main SQL migrations causes all the other migrations to fail.

raskchanky: So I'm going to leave it where it is

chefsalim: Ok 👍

components/builder-db/src/migration.rs
```diff
@@ -15,34 +15,54 @@
...
+    ).execute(conn)
         .unwrap();
     Ok(())
 }
+
+pub fn validate_shard_migration(pool: &Pool) -> Result<()> {
+    let conn = pool.get()?;
+    match conn.query("SELECT shard_migration_complete FROM flags;", &[]) {
+        Ok(rows) => {
+            debug!("**** DEBUG HAHA. rows = {:?}", &rows);
```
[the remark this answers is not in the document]
raskchanky: :joy: No, I will remove those.

tools/merge-shards/merge-shards.sh
```diff
@@ -0,0 +1,126 @@
...
+
+  # before we do anything, let's check to see if this has already been done
+  echo "Checking to see if the shard migration has already happened"
+  if flag=$(execute_sql "SELECT shard_migration_complete FROM flags;" "$server" | xargs); then
+    if [ -n "$flag" ]; then
+      flag_record_present="1"
+    fi
+
+    if [ "$flag" == "t" ]; then
+      echo "This shard migration has already completed. Running it a second time is a bad idea. Aborting."
```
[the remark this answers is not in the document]
raskchanky: Will do

chefsalim:
Looks like tests may need fixing... as soon as Travis is green, you should be good to merge!  :)

raskchanky:
@chefsalim I believe I've addressed all your feedback at this point.  I'm doing one final test run in my local environment.

tools/merge-shards/merge-shards.sh
```diff
@@ -0,0 +1,140 @@
...
+
+if [ -z "${PGPASSWORD:-}" ]; then
+  echo "Please ensure that PGPASSWORD is exported in your environment and run this script again."
+  exit 1
+fi
+
+echo "This script performs delicate surgery on your database."
+echo "It will take all of the data that's currently spread across 127 shards and consolidate all of that down into the public Postgres schema."
+echo "It performs some rudimentary error checking to ensure that all of the data was migrated successfully, but it's not exhaustive."
+echo "PLEASE ensure that you have completed a full database backup before running this script. If something goes horribly wrong, restoring from that backup is your only escape hatch."
```
chefsalim: This verbiage is new?  I don't know if I missed this earlier.  The comments about delicate surgery and things going horribly wrong should to be reworded as that is probably not the type of text we should expose to end users.  We can suggest that users take a backup (and should then provide some instructions on how to do in the on-prem README separately).

raskchanky: It's new.  I toned it down 🙂

.travis.yml
```diff
@@ -27,7 +27,9 @@ matrix:
 #
 # Job for linting Habitat shell program components
 #
-    - sudo: false
+    - sudo: true
+      before_script:
+        - ./support/ci/install_shellcheck.sh
```
baumanj: Personally, I'd _rather_ us just use whatever version is installed by default on travis than add extra logic to manage the install (which we also have to keep in sync across our several repos which run shellcheck). I understand the downside is that when travis upgrades, we'll have shellcheck failures at an unexpected moment, but when that happens, we can either:
- Make the fixes if there aren't too many
- Add top-level suppressions in test/shellcheck.sh while we make the fixes
- Temporarily disable shellcheck while we make the fixes

Since shellcheck doesn't update all that often, this may not even come up before we're fully off travis. Additionally, since travis doesn't seem to be aggressive about tracking the latest version of shellcheck (0.46 is from Mar 26, 2017), we could probably put in fixes for newer versions *before* travis upgrades and never need to see any disruption at all. I'd be happy to drive that work.

I worry that if we go the route of having a pinned version, we'll never touch it and never get any benefits of newer versions.

tools/merge-shards/merge-shards.sh
```diff
@@ -0,0 +1,138 @@
...
+fi
+
+echo "This script will take all of the data that's currently spread across 127 shards and consolidate all of that down into the public Postgres schema."
+echo "Please ensure that you have completed a full database backup before running this script."
+echo
+
+read -rp "Do you want to proceed? [y/n] " answer
+clean_answer=$(echo "$answer" | xargs | tr '[:upper:]' '[:lower:]')
+
+if [ "${clean_answer::1}" != "y" ]; then
```
baumanj: I'm not sure what `${clean_answer::1}` does and couldn't find it in any [documentation](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02). Is that a typo?

You could also, do this and obviate the `clean_answer` line:
```bash
if [ "${answer}" != [yY] ]; then
```
Or if you wanted to accept anything that *starts* with y:
```bash
if [ "${answer}" != [yY]* ]; then
```

tools/merge-shards/merge-shards.sh
```diff
@@ -0,0 +1,138 @@
...
+    echo "The flags table is not present in the database, which is required for this script to run."
+    echo "It's possible something is amiss with your Builder database migrations."
+    exit 1
+  fi
+
+  declare -A count_map
+
+  mapfile -t shards < <(seq 0 "$upper_bound")
+
+  for shard in "${shards[@]}"
```
baumanj: I think this is a bit simpler/more idiomatic:
```bash
  for shard in $(seq 0 "$upper_bound")
```

tools/merge-shards/merge-shards.sh
raskchanky: `${clean_answer::1}` is substring extraction.  That was my way of getting the first character of `$clean_answer`.

.travis.yml
raskchanky: I added this in because I installed the latest shellcheck locally, 0.5.0, and then got shellcheck failures when running in Travis because it was using a different version.  Like our other ci scripts for installing zmq, libarchive, libsodium, it felt better to me to install a known version, rather than just trusting whatever version happened to be installed.

tools/merge-shards/merge-shards.sh
baumanj: Ah, ok. I get it now, thank you. I prefer `[ "${answer}" != [yY]* ]` to do the same thing, but that's a matter of preference, so go with whatever you prefer.

raskchanky:
@baumanj I removed the travis shellcheck changes but left my script as is.  I'm a little paranoid of changing it at this stage of the game, given how much testing I did with it in its current state.

.travis.yml
baumanj: That's interesting to me that there were things that were failures in 0.4.6 (which runs on travis), but *not* failures in 0.5.0. My general experience had been that newer versions added more checks, so unless there were spurious things in 0.4.6, I'd have expected your local results to be a superset of what the travis run got.

PR #585, opened 2018-08-08
Adding instructions for building from live -> local

mgamini:
Signed-off-by: Garrett Amini <kadjar@gmail.com>


thesentinels:

Thank you for contributing!

DEVELOPING.md
```diff
@@ -125,7 +125,7 @@ load_package /hab/cache/artifacts/core-elasticsearch*.hart
 
 *NOTE*: By default, an automated fetch from the https://bldr.habitat.sh upstream is configured so that 'core' and 'habitat' origin packages are automatically retrieved from the upstream. Make sure you have those origins created. The upstream fetch happens in the background. To trigger a fetch for a non-existent package, you can do a local 'hab pkg search' for that package - it will initially fail with a 'Not Found', but will trigger a background fetch. If you run into any issues, check the upstream log file in the '/hab/svc/builder-api/var' directory.
```
chefsalim: Not that this is part of your PR, but this section on the upstream isn't correct - we don't have the upstream automatically configured, as it can cause subtle changes that become confusing when deving.  So it needs to be explicitly turned on if you want/need it.

DEVELOPING.md
```diff
@@ -149,3 +149,39 @@ This should create a build job, and then dispatch it to the build worker.
...
+
+```
+» Installing core/node
+☁ Determining latest version of core/node in the 'stable' channel
+✗✗✗
+✗✗✗ Package not found. If you are attempting to install from a local depot with an upstream configured, try again in a few seconds.
+✗✗✗
+```
+
+One way to resolve this dependency is by swapping your Builder URL, building locally, and then uploading to your local depot.
```
chefsalim: This seems pretty painful - why would we do that instead of just leveraging the upstream already built packages?

mgamini:
@chefsalim ready for 👀

DEVELOPING.md
```diff
@@ -149,3 +142,39 @@ This should create a build job, and then dispatch it to the build worker.
...
 Note: you will need to upload additional packages to the core origin for the `core/nginx` build to succeed. Follow the same procedure as for `core/hab-backline`. Currently `core/gcc` and `core/libedit` are required.
+
+### Setting up a package upstream
+
+It is possible to configure the on-premise builder to point to the hosted Builder site as an 'upstream'. This allows new packages from the upstream to get created in the on-premise instance automatically.
+
+In order to do so, create a file called `upstream.toml` with the following content:
+```
+[api]
+features_enabled = "upstream"
```
chefsalim: The features_enabled should be: "jobsrv, upstream"

DEVELOPING.md
```diff
@@ -149,3 +142,39 @@ This should create a build job, and then dispatch it to the build worker.
...
+[upstream]
+endpoint = "https://bldr.habitat.sh"
+```
+
+Then, issue the following command:
+```
+hab config apply builder-api.default $(date +%s) upstream.toml
+```
+After the config is successfully applied, the services should be configured to use the upstream.
+
```
chefsalim: May be worth also adding a note: Alternatively, the config can be directly added to the builder-api user.toml by modifying the support/config.sh file.

DEVELOPING.md
```diff
@@ -149,3 +142,39 @@ This should create a build job, and then dispatch it to the build worker.
...
+```
+hab pkg install -u http://localhost -z <auth-token> <package>
+```
+
+Initially, you will get a `Package Not Found` error.  Wait for a bit (the package will get synchronized in the background) and try again - this time the install should succeed!
+
+*NOTE*: It is important to understand how the upstream cache is working. Packages that are requested (either via a `hab pkg install`, or even searching or browsing packages in the Web UI) in the local on-premise depot that have newer (or existing) versions in the upstream in the *stable* channel are marked for retrieval in the background. It is only after the background retrieval of the package succeeds that the package then becomes available in the local instance. If there is any failure to retrieve or submit the package, the next retrieval attempt will be triggered only by another request for that package. This functionality is new, and will be refined over time.
+
+Once you have installed a package from upstream, you can load the package into your local depot as follows:
```
chefsalim: These following lines are confusing in this context and should probably be removed..

mgamini:
@chefsalim feedback incorporated. Also added `upstream.toml` to `.gitignore`.

DEVELOPING.md
```diff
@@ -149,3 +142,35 @@ This should create a build job, and then dispatch it to the build worker.
 You can view the build progress in the web UI or by viewing `/hab/svc/builder-worker/data/876066265100378112/log_pipe-876066265100378112.log`. Replace `876066265100378112` with the group ID output by the `start` command.
 
 Note: you will need to upload additional packages to the core origin for the `core/nginx` build to succeed. Follow the same procedure as for `core/hab-backline`. Currently `core/gcc` and `core/libedit` are required.
+
+### Setting up a package upstream
+
+It is possible to configure the on-premise builder to point to the hosted Builder site as an 'upstream'. This allows new packages from the upstream to get created in the on-premise instance automatically.
```
raskchanky: Technically your on-prem builder can point to any other builder instance it wants to as the upstream.  There's nothing that says it _has_ to point to The Hosted Builder™, though that's likely the most useful for most people.

DEVELOPING.md
```diff
@@ -149,3 +142,35 @@ This should create a build job, and then dispatch it to the build worker.
...
+
+In order to do so, create a file called `upstream.toml` with the following content:
+```
+[api]
+features_enabled = "jobsrc, upstream"
```
raskchanky: I believe this should be `jobsrv`, not `jobsrc`.

PR #589, opened 2018-08-14
#534: representing job groups in the Builder UI

mgamini:
See: https://github.com/habitat-sh/builder/issues/534 / https://app.zenhub.com/workspace/o/habitat-sh/builder/issues/534


thesentinels:

Thank you for contributing!

components/builder-web/index.html
```diff
@@ -28,12 +28,12 @@
   <body>
 
     <!-- Segment -->
-    <script>
+    <!-- <script>
```
cnunciato: Let's leave this out of the PR so we can continue to use it in dev.

chefsalim:
LGTM! 

tenor-55087389

cnunciato:
image

The font color is too light for these; they look disabled.

cnunciato:
image

The checkmarks here really strongly suggest these groups succeeded, but the builds behind all three of them actually failed. We should consider simply indicating `DONE`, or something less concrete, if we can't bubble up the overall success of the run yet.

cnunciato:
When I navigate into an origin I belong to and click Build Jobs, I should see the message indicating there aren't any when none exist. (You can see it flash there briefly, but it disappears.)

cnunciato:
Page titles are `undefined` here.

mgamini:
TODO:

-     Mobile view for job groups table
-     Mobile view for job groups detail
-     Radio toggle for job groups table
-     Remove labels for which we have no data on job groups detail

components/builder-web/app/origin/origin-page/origin-job-detail/origin-job-detail.component.html
```diff
@@ -0,0 +1,91 @@
+<div class="origin-job-detail-component body">
+  <div class="content">
+    <h3>
+      <a (click)="backToGroups()">
+        Build Job Groups
+      </a>&nbsp;/ Group {{ this.id }}
+    </h3>
+    <section class="jobgroup-overview">
+      <table class="jobgroup-table">
```
cnunciato: I'd rather this not be `table`-based. Let's handle this mobile-first with stacked `div`s and scale up on the desktop with Flexbox. Liz is working on some designs, I believe.

components/builder-web/app/origin/origin-page/origin-job-detail/origin-job-detail.component.html
```diff
@@ -0,0 +1,91 @@
...
+              <hab-icon symbol="chevron-right"></hab-icon>
+            </span>
+            <span class="column">
+            </span>
+          </a>
+        </li>
+      </ol>
+    </section>
+  </div>
+</div>
```
cnunciato: it's good practice to put a newline at the end of every file. You should be able to tweak your editor to apply  one on every file-save. Be good to fix up the few of them in this PR.

components/builder-web/app/origin/origin-page/origin-job-detail/origin-job-detail.component.ts
```diff
@@ -0,0 +1,123 @@
...
+  get thisGroup() {
+    return this.store.getState().jobGroups.selected;
+  }
+
+  projectStateCount(param) {
+    const stateList = this.thisGroup.projects_by_state[param];
+    return stateList ? stateList.length : '0';
+  }
+
+  dateFor(timestamp) {
```
cnunciato: There's a `releaseToDate` function exported from `util.ts` that we should probably use for this (rather than haul references to `moment` around everywhere), and maybe also fix up the other ones to use it.

components/builder-web/app/origin/origin-page/origin-jobs-tab/_origin-jobs-tab.component.scss
```diff
@@ -0,0 +1,8 @@
+.origin-jobs-tab-component {
+  display: flex;
+  flex-flow: column;
```
cnunciato: Why use flexbox here and not just `display: block`, since these children are just stacked and there's no call for column wrapping?

components/builder-web/app/origin/origin-page/origin-jobs-tab/_origin-jobs-tab.component.scss
```diff
@@ -0,0 +1,8 @@
...
+
+  > .jobs-list {
+    width: 100%;
```
cnunciato: This has no effect, as the element is currently `display: inline`. We can just remove it.

components/builder-web/app/origin/origin-page/origin-jobs-tab/jobs-list/jobs-list.component.ts
```diff
@@ -0,0 +1,39 @@
...
+})
+export class OriginJobsListComponent {
+  @Input() jobGroups = List();
+  @Output() select = new EventEmitter();
+
+  onClick(job) {
+    this.select.emit(job);
+  }
+
+  dateFor(timestamp) {
```
cnunciato: Replace with `util`'s `releaseToDate`.

components/builder-web/app/reducers/jobGroups.ts
```diff
@@ -0,0 +1,43 @@
...
+    case actionTypes.POPULATE_JOB_GROUP:
+      return state.setIn(['selected'], keyProjectsByState(action.payload));
+
+    default:
+      return state;
+  }
+}
+
+function keyProjectsByState(payload) {
+  if (!payload.projects) return payload;
```
cnunciato: When does this happen? How is `payload` useful if there aren't any projects on it? is this an error case?

components/builder-web/app/reducers/jobGroups.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  payload.projects_by_state = payload.projects.reduce((obj, project) => {
+    const state = project['state'].toLowerCase();
+    obj[state] = obj[state] ? obj[state].concat([project]) : [project];
```
cnunciato: This works, but you could simplify it a bit more with a spread operator and a default:

```
payload.projectsByState = payload.projects.reduce((acc, project) => {
  const state = project.state.toLowerCase();
  acc.state = [...acc.state || [], project];
  return acc;
}, {});
```

components/builder-web/stylesheets/base/_colors.scss
```diff
@@ -47,11 +47,14 @@ $disabled: $medium-gray;
 
 $status-colors: (
   complete: $success,
+  completed: $success,
```
cnunciato: 😬 I'd rather not add things like 'completed` and `failed` when we have `complete`, `failure`. Is this just because the API is returning different strings in different situations? If so, we should fix this at the API and in the meantime, translate somehow, rather than propagate the inconsistency.

components/builder-web/app/util.ts
```diff
@@ -119,13 +119,16 @@ export function targetToPlatform(target: string = ''): string {
 }
 
 // Return a build state's proper icon symbol
-export function iconForBuildState(state) {
+export function iconForJobState(state) {
   return {
     complete: 'check',
+    completed: 'check',
```
cnunciato: Same comment here as in `_colors.scss`.

components/builder-web/stylesheets/base/_colors.scss
mgamini: That's exactly it. The API here returns "completed" and "failed". I could try to take a look and fix this in the API?

cnunciato: IMO it'd be fine to file an issue on Builder with the request to change it, and in the meantime patch it on the front end somewhere close to where it's deserialized. Maybe in a reducer?

components/builder-web/app/origin/origin-page/origin-job-detail/origin-job-detail.component.html
mgamini: That could be done, but this is a literal table of data, even if it only has one row. Seems a perfectly appropriate place for a table?

cnunciato: I disagree; there are no rows of data, just labels and values. The mobile view makes this a bit clearer.

components/builder-web/app/origin/origin-page/origin-job-detail/origin-job-detail.component.ts
mgamini: releaseToDate doesn't include `hh:mm:ss` in what it returns. We could just create a passthrough function to isolate where we import `moment` so different components can get differently formatted returned date strings, but doesn't that defeat the purpose of the compiler's import efficiency?

components/builder-web/app/origin/origin-page/origin-jobs-tab/_origin-jobs-tab.component.scss
mgamini: `display: block` didn't work :)

components/builder-web/app/origin/origin-page/origin-job-detail/origin-job-detail.component.html
```diff
@@ -0,0 +1,91 @@
+<div class="origin-job-detail-component body">
+  <div class="content">
+    <h3>
+      <a (click)="backToGroups()">
+        Build Job Groups
+      </a>&nbsp;/ Group {{ this.id }}
```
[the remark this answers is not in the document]
mgamini: ✔️

components/builder-web/app/origin/origin-page/origin-jobs-tab/_origin-jobs-tab.component.scss
mgamini: ✔️

components/builder-web/app/reducers/jobGroups.ts
mgamini: It is an error case, yes. I could test for error, but this seemed to include more error states, or incomplete returns from the API, which happen sometimes.

components/builder-web/app/origin/origin-page/origin-job-detail/origin-job-detail.component.ts
cnunciato: Oh, fair enough -- yeah, then passing in a format string might be nice, but we could also just file an issue and handle that as a separate PR. 

> doesn't that defeat the purpose of the compiler's import efficiency?

No. 😉 Either way, though, it's not a compilation-speed thing &mdash; it's just to reduce duplication.

components/builder-web/app/reducers/jobGroups.ts
mgamini: Huh, I didn't know the spread operator and a default could be used like that. Cool!

components/builder-web/app/reducers/jobGroups.ts
cnunciato: Are these error cases handled anywhere?

components/builder-web/app/origin/origin-page/origin-jobs-tab/_origin-jobs-tab.component.scss
cnunciato: Happy to pair with you on this, then; I can't think of a reason good ol' blocks wouldn't work here. If they don't, something else must be going on.

mgamini:
@cnunciato mobile views done. Ready for a bit of feedback. Will squash before merging!

components/builder-web/app/origin/origin-page/origin-jobs-tab/_origin-jobs-tab.component.scss
cnunciato: So let's remove this file altogether, then, since the flexbox isn't needed, and use a `section` to contain the list.

components/builder-web/app/origin/origin-page/origin-job-detail/_origin-job-detail.component.scss
```diff
@@ -0,0 +1,78 @@
+.origin-job-detail-component {
+  .jobgroup-overview {
+    @include shadow-panel;
+
+    section.jobgroup-details {
+      margin-top: 0px;
```
cnunciato: Minor, but `0px` (or `0` pretty-much-anything) should always just be `0` -- the linter should be catching these, though. 😕

components/builder-web/app/origin/origin-page/origin-job-detail/origin-job-detail.component.ts
```diff
@@ -0,0 +1,122 @@
...
+    return this.thisGroup.projects;
+  }
+
+  get thisGroup() {
+    return this.store.getState().jobGroups.selected;
+  }
+
+  projectStateCount(param) {
+    const stateList = this.thisGroup.projects_by_state[param];
+    return stateList ? stateList.length : '0';
```
cnunciato: This should probably just be `0` rather than `'0'` (since you're returning `.length`, a number, when there is one).

components/builder-web/app/origin/origin-page/origin-job-detail/origin-job-detail.component.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  dateFor(timestamp) {
+    return parseDate(timestamp, 'YYYY-MM-DD HH:mm:ss');
+  }
+
+  onSelectJob(name, job) {
+    this.router.navigate(
+      ['pkgs']
```
cnunciato: Totally fine, but you could also do it in one step:

```
this.router.navigate(['pkgs', ...name.split('/'), 'jobs', job]);
```

cnunciato:
Looks good! I'd prefer you fixed up the function returning the string `'0'` instead of a number, but otherwise, 👍 !

mgamini:
@cnunciato 👍

PR #594, opened 2018-08-20
Remove the "op" tool since we're on 1 schema now

raskchanky:


thesentinels:

Thank you for contributing!

raskchanky:
For those that don't know, the `op` tool was something I wrote to calculate what shard to look for, given either a database ID or a text string representing an origin name.  Now that we've merged all the shards into the public schema, this tool is useless.

chefsalim:
The op tool has the base64 session token decode capability as well that is still valid use case.. also we may add more session token related tooling in the future, so might be better to just leave it as is for now... it might help debug shard issues for on-prem installations that have not migrated yet.

raskchanky:
Ok, fair enough

PR #602, opened 2018-08-23
Adds channel support with actix-web

eeyun:

### DIFF
--- a/components/builder-api/src/server/resources/channels.rs
+++ b/components/builder-api/src/server/resources/channels.rs
@@ -14,10 +14,12 @@
 
 use actix_web::http::{self, StatusCode};
 use actix_web::FromRequest;
-use actix_web::{App, HttpRequest, HttpResponse, Json, Path};
+use actix_web::{App, HttpRequest, HttpResponse, Json, Path, Query};
 use protocol::originsrv::*;
+use serde_json;
 
 use hab_core::package::ident;
+use hab_net::{ErrCode, NetError, NetOk};
 
 use server::error::{Error, Result};
 use server::framework::headers;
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/builder-api/src/server/resources/channels.rs
```diff
@@ -31,354 +33,314 @@ impl Channels {
     //
     // Internal - these functions should return Result<..>
     //
+    fn do_promote_package(
+        req: &HttpRequest<AppState>,
+        origin: String,
+        channel: String,
+    ) -> Result<NetOk> {
```
chefsalim: You can just have this be Result<()>

components/builder-api/src/server/resources/channels.rs
```diff
@@ -31,354 +33,314 @@ impl Channels {
...
+        let ident = OriginPackageIdent::new();
+
+        if !helpers::check_origin_access(req, &origin).is_err() {
```
chefsalim: You can change this to
```
helpers::check_origin_access(req, &origin)?;
```

components/builder-api/src/server/resources/channels.rs
```diff
@@ -31,354 +33,314 @@ impl Channels {
...
+        origin_get.set_name(ident.get_origin().to_string());
 
-    //
-    // Route registration
-    //
-    pub fn register(app: App<AppState>) -> App<AppState> {
-        app
-    }
-}
+        let origin_id = match route_message::<OriginGet, Origin>(req, &origin_get) {
```
chefsalim: Mjght be a good time to look whether we need all these Channel.. errors - I think we can just remove channel specific errors as they are not getting used upstream ever... so this paragraph essentially becomes:
```
let origin_id = route_message::<OriginGet, Origin>(req, &origin_get)?;
```

components/builder-api/src/server/resources/channels.rs
```diff
@@ -31,354 +33,314 @@ impl Channels {
...
+            }
+        };
+
+        let mut promote = OriginPackagePromote::new();
+        promote.set_channel_id(origin_channel.get_id());
+        promote.set_package_id(package.get_id());
+        promote.set_ident(ident.clone());
+
+        match route_message::<OriginPackagePromote, NetOk>(req, &promote) {
+            Ok(_) => match route_message::<OriginPackagePromote, NetOk>(req, &promote) {
```
chefsalim: Why are there two OriginPackagePromote calls?

components/builder-api/src/server/resources/channels.rs
```diff
@@ -31,354 +33,314 @@ impl Channels {
...
+                Ok(_) => match Self::audit_package_rank_change(
+                    req,
+                    package.get_id(),
+                    origin_channel.get_id(),
+                    PackageChannelOperation::Promote,
+                    origin_id.get_id(),
+                ) {
+                    Ok(rank) => return Ok(rank),
```
chefsalim: not clear we need to return anything from the audit_package call...

components/builder-api/src/server/resources/channels.rs
```diff
@@ -31,354 +33,314 @@ impl Channels {
...
-                .collect();
-            let body = serde_json::to_string(&list).unwrap();
-            let mut response = Response::with((status::Ok, body));
-            dont_cache_response(&mut response);
-            Ok(response)
+    fn do_demote_package(
+        req: &HttpRequest<AppState>,
+        origin: String,
+        channel: String,
+    ) -> Result<NetOk> {
```
chefsalim: Same, return Result<()>..

components/builder-api/src/server/resources/channels.rs
```diff
@@ -31,354 +33,314 @@ impl Channels {
...
+        route_message::<PackageChannelAudit, NetOk>(req, &audit)
+    }
 
-    let mut request = OriginPackageGet::new();
-    request.set_ident(ident.clone());
-    request.set_visibilities(all_visibilities());
+    //
+    // Route handlers - these functions should return HttpResponse
+    //
+    pub fn list_channels(req: &HttpRequest<AppState>) -> HttpResponse {
```
chefsalim: change "list_channels" to "get_channels" to reflect the more consistent naming strategy called out in the doc..

components/builder-api/src/server/resources/channels.rs
eeyun: I'm not sure on that, this is how it is in master right now.

components/builder-api/src/server/resources/channels.rs
```diff
@@ -31,354 +33,314 @@ impl Channels {
...
-        Some(channel) => channel,
-        None => return Ok(Response::with(status::BadRequest)),
-    };
-
-    // you can't demote from "unstable"
-    if channel == "unstable" {
-        return Ok(Response::with(status::Forbidden));
+        // Pass ?sandbox=true to this endpoint to include sandbox channels in the list. They are not
+        // there by default.
+        let sandbox = Query::<(String)>::extract(&req).unwrap().into_inner();
```
chefsalim: You can put this query param handling in the function prototype, and have it automatically parse into a struct... see the code sample from one of the other modules (eg pkgs.rs).. lot cleaner

components/builder-api/src/server/resources/channels.rs
chefsalim: I think it's a copy-pasta error - we can just call promote once..

components/builder-api/src/server/resources/channels.rs
```diff
@@ -31,354 +33,314 @@ impl Channels {
...
+    }
 
-    let mut request = OriginPackageGet::new();
-    request.set_ident(ident.clone());
-    request.set_visibilities(all_visibilities());
+    //
+    // Route handlers - these functions should return HttpResponse
+    //
+    pub fn list_channels(req: &HttpRequest<AppState>) -> HttpResponse {
+        let origin = Path::<(String)>::extract(&req).unwrap().into_inner(); //Unwrap apparently Ok
```
chefsalim: The unwrap here is definitely ok.. the handler will never get called if there is no string to match in the route

components/builder-api/src/server/resources/channels.rs
```diff
@@ -14,371 +14,326 @@
...
+
+        let mut channel_req = OriginChannelGet::new();
+        channel_req.set_origin_name(ident.get_origin().to_string());
+        channel_req.set_name(channel.to_string());
+
+        let origin_channel =
+            match route_message::<OriginChannelGet, OriginPackage>(req, &channel_req) {
+                Ok(c) => c,
+                Err(e) => {
+                    error!("Error retrieving channel: {:?}", &e);
```
chefsalim: minor - switch error! to warn! .. maybe even debug! if these can happen under 'normal' operation

components/builder-api/src/server/resources/channels.rs
```diff
@@ -14,371 +14,326 @@
...
 
-    let package = match route_message::<OriginPackageGet, OriginPackage>(req, &request) {
-        Ok(p) => p,
-        Err(e) => {
-            error!("Error retrieving package: {:?}", &e);
-            return Ok(render_net_error(&e));
-        }
-    };
+        let mut request = OriginChannelListRequest::new();
+        request.set_include_sandbox_channels(false);
```
chefsalim: use sandbox.is_set in the param?

components/builder-api/src/server/resources/channels.rs
```diff
@@ -14,371 +14,326 @@
...
+                    .map(|channel| {
+                        let mut ident = OriginChannelIdent::new();
+                        ident.set_name(channel.get_name().to_string());
+                        ident
+                    })
+                    .collect();
+                let body = serde_json::to_string(&list).unwrap();
+                let mut response = HttpResponse::Ok();
+                response
+                    .header(http::header::CACHE_CONTROL, "private, no-cache, no-store")
```
chefsalim: use const from middleware/headers.rs for the cache control string..

components/builder-api/src/server/resources/channels.rs
```diff
@@ -14,371 +14,326 @@
...
+            Err(err) => err.into(),
+        }
+    }
 
-    route_message::<PackageChannelAudit, NetOk>(req, &audit)
+    //
+    // Route registration
+    //
+    pub fn register(app: App<AppState>) -> App<AppState> {
+        app.resource("/channels/:origin", |r| {
```
chefsalim: change the route registration params .. eg from ":origin" to "{origin}" (for example)
I missed that when wiring up some of the modules :)

PR #603, opened 2018-08-23
Add memcached support for builder

elliott-davis:
When a package gets uploaded, promoted or marked as public the latest version of that package gets added to memcached.
When a packages gets demoted or marked as private, it is removed from the cache.

Any calls to `/latest` endpoints will no check the cache to see if the package exists. If it does, it's returned, if not, we round trip to the db for a result and update the cache.


thesentinels:

Thank you for contributing!

components/builder-api/habitat/config/config.toml
```diff
@@ -14,11 +14,11 @@ app_private_key = "{{pkg.svc_files_path}}/builder-github-app.pem"
 [oauth]
 {{toToml cfg.oauth}}
 
-{{~#eachAlive bind.router.members as |member|}}
+{{~#each bind.router.members as |member|}}
```
raskchanky: Was this intentional?

components/builder-api/habitat/config/config.toml
```diff
@@ -31,3 +31,9 @@ root = "{{pkg.svc_static_path}}"
 
 [upstream]
 {{toToml cfg.upstream}}
+
+{{~#each bind.memcached.members as |member|}}
+[[memcached]]
+host = "{{member.sys.ip}}"
```
raskchanky: Am I reading this right, that the `host` portion of your `MemcacheCfg` struct is not used in the config file?

components/builder-api/src/depot/server.rs
```diff
@@ -786,6 +786,15 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
...
     );
 
+    // This stinks but is necessary since memcached Client returns mutable reference
+    // and iron chain.link returns everything wrapped in Arc
+    // The tradeoff here is that we have to initialize a new client on every request until
+    // we can switch to a framework that's not Iron
+    let memcache_host_strings: Vec<String> =
+        depot.memcached.iter().map(|h| format!("{}", h)).collect();
+    let memcache_hosts: Vec<&str> = memcache_host_strings.iter().map(AsRef::as_ref).collect();
+    let mut client = memcache::Client::new(memcache_hosts).unwrap();
```
chefsalim: Ugh... yeah we can live with it, we're close to landing the new framework..

components/builder-api/habitat/config/config.toml
elliott-davis: It was. @christophermaier added a feature that made using `eachAlive` unnecessary.

components/builder-api/src/depot/server.rs
```diff
@@ -1426,6 +1444,16 @@ fn package_privacy_toggle(req: &mut Request) -> IronResult<Response> {
...
 
+    // This stinks but is necessary since memcached Client returns mutable reference
+    // and iron chain.link returns everything wrapped in Arc
+    // The tradeoff here is that we have to initialize a new client on every request until
+    // we can switch to a framework that's not Iron
+    let depot = req.get::<persistent::Read<Config>>().unwrap();
+    let memcache_host_strings: Vec<String> =
+        depot.memcached.iter().map(|h| format!("{}", h)).collect();
+    let memcache_hosts: Vec<&str> = memcache_host_strings.iter().map(AsRef::as_ref).collect();
+    let mut client = memcache::Client::new(memcache_hosts).unwrap();
```
raskchanky: This ugliness is repeated a few times.  You could make a helper function that takes an iron `req` and returns a `memcache::Client`, if you want.

elliott-davis:
One of the tests in here will always fail until we build and promote the builder-memcached package contained in this PR

components/builder-api/src/depot/server.rs
```diff
@@ -1703,26 +1757,74 @@ fn show_package(req: &mut Request) -> IronResult<Response> {
...
```
raskchanky: Same comment as above re: helper function.

components/builder-api/src/depot/server.rs
chefsalim: Alternatively, just have the _memcache_package functions create the client internally so the callers don't have to do it.

components/builder-api/src/depot/server.rs
```diff
@@ -1937,6 +2062,16 @@ fn promote_package(req: &mut Request) -> IronResult<Response> {
...
```
raskchanky: That helper function is looking better and better 😂

components/builder-api/src/depot/server.rs
```diff
@@ -2020,6 +2168,16 @@ fn demote_package(req: &mut Request) -> IronResult<Response> {
...
```
raskchanky: That helper function is pretty much worth its weight in gold by this point in the file.

components/builder-api/src/depot/server.rs
```diff
@@ -965,7 +973,17 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
                         &filename, e
                     ),
                 }
-
+                match set_memcache_package(
+                    &mut client,
+                    ident.get_origin().to_string(),
```
chefsalim: Just pass in the full ident struct as a reference?

components/builder-api/src/depot/server.rs
```diff
@@ -2654,6 +2827,94 @@ fn do_cache_response(response: &mut Response) {
     )));
 }
 
+fn set_memcache_package(
+    client: &mut memcache::Client,
+    origin: String,
+    name: String,
+    version: Option<String>,
+    channel: Option<String>,
+    key: &str,
```
raskchanky: Is this actually the value?  From the memcache crate examples, it looks like the first arg is the key and the second is the value.  If that's the case, probably worth renaming this.

components/builder-api/src/depot/server.rs
elliott-davis: oh that's a good idea since PackageIdent versions and revisions are also optional!

components/builder-api/src/config.rs
```diff
@@ -153,6 +156,28 @@ impl Default for ApiCfg {
     }
 }
 
+#[derive(Debug, Clone, Deserialize)]
+#[serde(default)]
+pub struct MemcacheCfg {
+    pub host: String,
+    pub port: i32,
```
baumanj: In the spirit of ultra type-safety, should this be a u16?

components/builder-api/src/depot/server.rs
raskchanky: That's actually a better idea 👍

components/builder-api/src/depot/server.rs
```diff
@@ -2654,6 +2827,94 @@ fn do_cache_response(response: &mut Response) {
     )));
 }
 
+fn set_memcache_package(
```
chefsalim: It would be great if you could pull the memcache functions to a new file instead of leaving them in server.rs.  Also create a 'struct MemCache' and put the functions into the struct.  This will make it much easier to integrate into the new framework.  Then you could just do something like:

```
MemCache::new(config).set_package(&ident);
```

chefsalim: Also, it would be great if you could hook up a couple of metrics in the MemCache struct so we can monitor the caching get/sets on our dashboard.

components/builder-api/src/depot/server.rs
```diff
@@ -945,12 +954,11 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
             builder_flag,
         ) {
             Ok(_) => {
-                let mut response = Response::with((
-                    status::Created,
-                    format!("/pkgs/{}/download", package.get_ident()),
-                ));
+                let ident = package.get_ident();
```
chefsalim: Would be nice not to change any of this code - just use package.get_ident() .. it will make life easier in rebase

components/builder-api/src/depot/server.rs
```diff
@@ -2654,6 +2827,94 @@ fn do_cache_response(response: &mut Response) {
...
 }
 
+fn set_memcache_package(
+    client: &mut memcache::Client,
+    origin: String,
+    name: String,
+    version: Option<String>,
+    channel: Option<String>,
+    key: &str,
+) -> Result<()> {
```
chefsalim: Probably cleaner to omit Result - the callers should not have to check / react to failure to set or delete cache - we don't want to fail the caller ever on set/delete.

components/builder-api/src/depot/server.rs
```diff
@@ -2654,6 +2827,94 @@ fn do_cache_response(response: &mut Response) {
...
+    };
+    Ok(())
+}
+
+fn get_memcache_package(
+    client: &mut memcache::Client,
+    origin: &str,
+    name: &str,
+    version: Option<&str>,
+    channel: Option<&str>,
```
chefsalim: I'm not really liking the optional channel here... the vast majority of the calls we're concerned with have a channel attached... maybe better to always require a channel.. see what happens and what the shape of the code looks like.

components/builder-api/src/depot/server.rs
```diff
@@ -2654,6 +2827,94 @@ fn do_cache_response(response: &mut Response) {
...
+        },
+    };
+    return res;
+}
+
+fn delete_memcache_package(
+    client: &mut memcache::Client,
+    origin: String,
+    name: String,
+    channel: Option<String>,
```
chefsalim: same feel wrt optional channel

components/builder-api/src/depot/server.rs
```diff
@@ -1462,7 +1490,33 @@ fn package_privacy_toggle(req: &mut Request) -> IronResult<Response> {
...
+                            Ok(_) => {}
+                            Err(e) => error!("Failed to cache artifact {}", e),
+                        };
+                    } else {
+                        match delete_memcache_package(
+                            &mut client,
+                            ident.get_origin().to_string(),
+                            ident.get_name().to_string(),
+                            Some(ident.get_version().to_string()),
+                            None,
```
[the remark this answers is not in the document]
elliott-davis: I think I've covered all the cases in the set and delete function. We might consider adding a latest check for `origin/name/version` without a channel?

components/builder-api/src/depot/memcached.rs
```diff
@@ -0,0 +1,139 @@
...
+            cli: memcache::Client::new(memcache_hosts).unwrap(),
+        }
+    }
+
+    pub fn set_package(&mut self, package: PackageIdent, channel: &str) {
+        if package.version.is_none() || package.release.is_none() {
+            debug!("Can't insert non-fully qualified package ident into the cache");
+            return;
+        }
+        match self.cli.set(
```
chefsalim: I think you could just do this in a loop...

components/builder-api/src/depot/memcached.rs
```diff
@@ -0,0 +1,139 @@
...
+            package.to_string(),
+            0,
+        ) {
+            Ok(_) => debug!("Saved {:?} to memcached", package.to_string()),
+            Err(_) => error!("Failed to save {:?} to memcached", package.to_string()),
+        };
+        match self.cli.set(
+            &format!("{}/{}/{}", channel, package.origin, package.name),
+            package.to_string(),
+            0,
```
chefsalim: Just to be safe, might be good to have an expiration on the sets.. maybe 15 mins?

components/builder-api/src/depot/memcached.rs
```diff
@@ -0,0 +1,139 @@
...
+            }
+        }
+    }
+
+    pub fn delete_package(&mut self, package: PackageIdent, channel: &str) {
+        if package.version.is_none() || package.release.is_none() {
+            debug!("Can't insert non-fully qualified package ident into the cache");
+            return;
+        }
+        match self
```
chefsalim: can probably do this cleaner as a loop..

components/builder-api/src/depot/server.rs
```diff
@@ -1462,7 +1462,14 @@ fn package_privacy_toggle(req: &mut Request) -> IronResult<Response> {
             opu.set_pkg(package);
 
             match route_message::<OriginPackageUpdate, NetOk>(req, &opu) {
-                Ok(_) => Ok(Response::with(status::Ok)),
+                Ok(_) => {
```
chefsalim: thinking about this more - 1) I'm still not sure this logic is correct; and 2) I think we can actually eliminate this.  if we set a short expiration like 15 mins, things will just get naturally expired out, and we don't need to go hunting for all the channels that this package was in and manually try to delete them.  So net net, suggest we remove this section changes.

components/builder-api/src/depot/server.rs
```diff
@@ -1702,27 +1709,50 @@ fn show_package(req: &mut Request) -> IronResult<Response> {
...
-            request.set_target(target.to_string());
-            request.set_visibilities(visibility_for_optional_session(
-                req,
-                session_id,
-                &ident.get_origin(),
-            ));
-            request.set_ident(ident.clone());
-
-            match route_message::<OriginChannelPackageLatestGet, OriginPackageIdent>(req, &request)
+            match MemCache::new(config.clone()).get_package(ident.clone().into(), channel.as_str())
```
chefsalim: if you pull this up to the top of the function.. 
```
if let Some(id) = MemCache::new(..).get_package(...) {
  return the Ident
}
```
 that should cause fewer changes in the rest of the function below (other than a couple of lines to write out to the cache)

components/builder-api/src/depot/server.rs
```diff
@@ -1702,27 +1709,50 @@ fn show_package(req: &mut Request) -> IronResult<Response> {
...
+                        session_id,
+                        &ident.get_origin(),
+                    ));
+                    request.set_ident(ident.clone());
+
+                    match route_message::<OriginChannelPackageLatestGet, OriginPackageIdent>(
+                        req, &request,
+                    ) {
+                        Ok(id) => {
+                            ident = {
```
chefsalim: not clear why the memcache call is inside the ident scope.. borrowing issue?

components/builder-api/src/depot/server.rs
```diff
@@ -1981,16 +2024,19 @@ fn promote_package(req: &mut Request) -> IronResult<Response> {
...
-                package.get_id(),
-                origin_channel.get_id(),
-                PackageChannelOperation::Promote,
-                origin_id,
-            ) {
-                Ok(_) => Ok(Response::with(status::Ok)),
-                Err(err) => return Ok(render_net_error(&err)),
-            },
+            Ok(_) => {
+                MemCache::new(config).set_package(ident.into(), &channel);
```
chefsalim: putting this in the ok arm of audit_package_rank_change might result in fewer overall lines changed (== less rebase pain)

components/builder-api/src/depot/memcached.rs
```diff
@@ -0,0 +1,139 @@
...
+    pub fn new(config: Arc<Config>) -> Self {
+        let memcache_host_strings: Vec<String> =
+            config.memcached.iter().map(|h| format!("{}", h)).collect();
+        let memcache_hosts: Vec<&str> = memcache_host_strings.iter().map(AsRef::as_ref).collect();
+        MemCache {
+            cli: memcache::Client::new(memcache_hosts).unwrap(),
+        }
+    }
+
+    pub fn set_package(&mut self, package: PackageIdent, channel: &str) {
```
chefsalim: also metrics would be good to add in the functions..see metrics.rs ..

components/builder-api/src/depot/memcached.rs
```diff
@@ -0,0 +1,139 @@
...
+        if package.version.is_none() || package.release.is_none() {
+            debug!("Can't insert non-fully qualified package ident into the cache");
+            return;
+        }
+        match self.cli.set(
+            &format!("{}/{}", package.origin, package.name),
```
chefsalim: I think there may be another logic issue here in terms of an assumption regarding the latest for a non-qualified channel "origin/name" package - we are setting it in the cache based on a specific channel's latest ident, but those two are not necessarily the same...

components/builder-api/src/depot/memcached.rs
```diff
@@ -0,0 +1,139 @@
...
+            package.to_string(),
+            0,
+        ) {
+            Ok(_) => debug!("Saved {:?} to memcached", package.to_string()),
+            Err(_) => error!("Failed to save {:?} to memcached", package.to_string()),
```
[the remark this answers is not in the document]
elliott-davis: why are we yelling? :trollface:

components/builder-api/src/depot/memcached.rs
elliott-davis: That is 100% true and a great source for a potential bug. Good catch!

components/builder-api/src/depot/server.rs
baumanj: A reasonable question to ask there is how the cache needs to behave in order to be effective in its purpose. If we can easily tolerate this query once every 15 minutes, perhaps the expiration time should be even lower.

Another thing to think about is whether many things will enter the cache right around the same time, and therefore leave the cache all at the same time, potentially leading to traffic spikes. If that's a concern, randomizing the TTL over an acceptable lifetime may be prudent.

PR #604, opened 2018-08-24
Upgrade Angular and other dependencies

cnunciato:
This change upgrades Builder UI to the latest version of Angular (6) and brings a number of  runtime and development dependencies to their latest versions as well, along with the necessary refactorings where appropriate. 

It also applies the Node version specified in `.nvmrc` to ensure `builder-api-proxy` packages get built using the same version of Node that we use in development.

It stops short of updating Karma- and Webpack-related dependencies, however, in order to save those for a follow-up PR, which will certainly be oh so much fun, indeed.


thesentinels:

Thank you for contributing!

PR #606, opened 2018-08-24
Merge databases for originsrv, sessionsrv and jobsrv into a single builder database

elliott-davis:
Merge databases for originsrv, sessionsrv and jobsrv into a single builder database

This change comes with a migration script that will need to be run after services are upgraded in tools/merge-databases.

This is the first pass and collapsing our services into a more condensed service that will be easier to	manage and deploy.
By merging the databases I have reduced the amount of data duplication across disparate	databases and will make	it easier to reason about service domains going forward.
I also took the	liberty of removing old migrations since we are starting with a pristine database and included a snapshot of the tables and functions we are currently using.

Finally, this removed all of the async code required for package and origin syncing.


thesentinels:

Thank you for contributing!

components/builder-db/src/error.rs
```diff
@@ -41,7 +41,6 @@ pub enum Error {
     SchemaDrop(postgres::error::Error),
     SchemaSwitch(postgres::error::Error),
     SetSearchPath(postgres::error::Error),
-    ShardMigrationIncomplete,
```
chefsalim: I'm not sure we can remove this since we have potential on-prem instances that might have un-migrated data.

components/builder-originsrv/src/data_store.rs
```diff
@@ -2035,43 +1982,6 @@ impl DataStore {
...
-                    );
-                    result = EventOutcome::Retry;
-                }
-            }
-        }
-    }
-    Ok(result)
-}
-
 fn sync_packages(pool: Pool, mut route_conn: RouteClient) -> DbResult<EventOutcome> {
```
chefsalim: When we merge jobsrv then you can remove sync_packages, AND all of the underlying sync code in builder-db as well!

components/builder-originsrv/src/migrations/2018-03-05-224023_origin_secrets/down.sql
```diff
@@ -1 +0,0 @@
--- This file should undo anything in `up.sql`
```
chefsalim: You might want to add a couple lines in the Migrations doc that we should leave the down.sqls blank (that will also close another issue we currently have open :)

components/builder-originsrv/src/migrations/2018-08-24-184520_initial_originsrv_migration/up.sql
```diff
@@ -0,0 +1,1149 @@
...
+CREATE TABLE origins (
+    id bigint DEFAULT next_id_v1('origin_id_seq') PRIMARY KEY NOT NULL,
+    name text UNIQUE,
+    owner_id bigint,
+    created_at timestamp with time zone DEFAULT now(),
+    updated_at timestamp with time zone DEFAULT now(),
+    default_package_visibility text DEFAULT 'public'::text NOT NULL
+);
+
+CREATE TABLE audit (
```
chefsalim: Do we still use the audit table?  I recall that we deprecated it in lieu of audit_packges/audit_package_groups.  If so, we can remove it (I don't believe it made it into prod, but we should check)

components/builder-originsrv/src/migrations/2018-08-24-184520_initial_originsrv_migration/up.sql
```diff
@@ -0,0 +1,1149 @@
...
+      UPDATE origin_invitations SET ignored = true, updated_at = now() WHERE id = oi_invite_id;
+    ELSE
+      SELECT origin_id, origin_name, account_id, account_name INTO oi_origin_id, oi_origin_name, oi_account_id, oi_account_name FROM origin_invitations WHERE id = oi_invite_id;
+      PERFORM insert_origin_member_v1(oi_origin_id, oi_origin_name, oi_account_id, oi_account_name);
+      DELETE FROM origin_invitations WHERE id = oi_invite_id;
+    END IF;
+  END
+$$;
+
+CREATE FUNCTION add_audit_entry_v1(p_origin_id bigint, p_package_id bigint, p_channel_id bigint, p_operation smallint, p_trigger smallint, p_requester_id bigint, p_requester_name text) RETURNS SETOF audit
```
chefsalim: we can also remove this..

note: jobsrv has audit table/functions with same name (audit/add_audit_entry_v1), so removing these here will make it easier to migrate jobsrv over if we want to keep the same schema

support/builder/config.sh
```diff
@@ -78,7 +78,6 @@ log_level = "debug"
 
 [datastore]
 password = "$PGPASSWORD"
-database = "builder_jobsrv"
```
chefsalim: How does jobsrv connect if this is removed?

components/builder-originsrv/src/migrations/2018-08-24-184520_initial_originsrv_migration/up.sql
poopydoodles: I was hoping we'd be able to remove it but I haven't had time to double back to see if it's used

support/builder/config.sh
poopydoodles: it's the default in `default.toml`

components/builder-originsrv/src/data_store.rs
elliott-davis: do you want me to do that as part of this PR? I'm neck deep in it anyway

components/builder-db/src/error.rs
chucklehunk: If we leave this we will have db table collision between the services how big of an upgrade do we need to support? Maybe we could create a touch file instead?

chefsalim: Where is the db table collision happening?  I don't see the flags table in the merge script..

elliott-davis: That's because I don't migrate the flags table to avoid the collision

tools/merge-databases/merge-databases.sh
```diff
@@ -0,0 +1,79 @@
...
+echo "Checking to see if the shard migration has already happened"
+if ! execute_sql "builder_originsrv" "SELECT shard_migration_complete FROM flags;" | grep -q 't'; then
+echo "The flags table is not present in the database, which is required for this script to run."
+echo "It's possible something is amiss with your Builder database migrations."
+exit 1
+fi
+
+# before we do anything part 2, let's check to see if this has already been done
+echo "Checking to see if the db migration has already happened"
+# shellcheck disable=SC2046
```
baumanj: Why not
```bash
if [ "$(execute_sql "builder" "SELECT count(*) from origins;")" -gt 0 ]; then
```
?

elliott-davis: I'm pretty sure that breaks the inner quoting and I can't escape the inner quotes because then it translates into literal sql

baumanj: It doesn't break the inner quoting. Shell handles this fairly differently from other languages in that you can nest quotes and expansions to an arbitrary depth. For example:
```bash
jbauman@ubuntu:~/builder$ touch "foo bar"
jbauman@ubuntu:~/builder$ echo "$(ls -l "foo bar")"
-rw-rw-r-- 1 jbauman jbauman 0 Aug 29 08:14 foo bar
```

DEVELOPING.md
```diff
@@ -85,42 +85,6 @@ If all goes well, you'll see some messages about things being all good, and
 your services should behave normally. If they don't, exit the studio and
 re-enter.
 
-If you try to run a build and get an error message in the logs about
```
chefsalim: Also should add instruction here on how to do the migration for a dev env...

components/builder-db/src/error.rs
chefsalim: As discussed if the migration script checks for the flags table and requires a migration before proceeding then we should be fine not migrating the flags table.  But we still must leave this check in.  If an on-prem user uninstalls/reinstalls, and gets the latest services, the services will start normally if the check is not there.  That will mess up their env.  So the shard migration check needs to be kept in the code

DEVELOPING.md
chefsalim: Also can you please open an issue to update key docs - eg, the 'on-call' docs which deal with the DB instructions, and follow up on that once this merges.

components/builder-db/src/error.rs
elliott-davis: The way the tables are designed it's not possible to leave them in safely. They all have the same table name and the same column name to check if a migration is complete. If you migrate one of the flags tables and leave this code in it doesn't mean that the other two databases have their shards migrated. What if we write a script that creates a touch file in the services mutable directory based on the table in the db? Then we can put part of this code back and rely on the touch-file to see if the migration is complete

components/builder-jobsrv/src/migrations/2018-08-29-173340_initial_jobsrv_migration/up.sql
```diff
@@ -0,0 +1,393 @@
+CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;
+CREATE SEQUENCE IF NOT EXISTS groups_id_seq;
+CREATE SEQUENCE IF NOT EXISTS job_id_seq;
+CREATE SEQUENCE IF NOT EXISTS group_projects_id_seq;
+
+CREATE TABLE audit (
```
chefsalim: Might be good to rename this table to audit_jobs or audit_job_groups..

elliott-davis:
```[16][default:/src:130]# PORT=5433 PGPASSWORD=$(cat /hab/svc/builder-datastore/config/pwfile) tools/merge-databases/merge-databases.sh 
This script will take all of the data that's currently in builder_originsrv and builder_sessionsrv and merge them into builder.
Please ensure that you have completed a full database backup before running this script.

Do you want to proceed? [y/n] y
Checking to see if the shard migration has already happened
Checking to see if the db migration has already happened
COPY 0
COPY 0
Copying accounts from builder_sessionsrv to builder
COPY 0
Copying account_tokens from builder_sessionsrv to builder
COPY 0
Copying audit_package from builder_originsrv to builder
COPY 0
Copying audit_package_group from builder_originsrv to builder
COPY 0
Copying origin_channels from builder_originsrv to builder
COPY 0
Copying origin_integrations from builder_originsrv to builder
COPY 0
Copying origin_members from builder_originsrv to builder
COPY 0
Copying origin_packages from builder_originsrv to builder
COPY 0
Copying origin_channel_packages from builder_originsrv to builder
COPY 0
Copying origin_private_encryption_keys from builder_originsrv to builder
COPY 0
Copying origin_projects from builder_originsrv to builder
COPY 0
Copying origin_project_integrations from builder_originsrv to builder
COPY 0
Copying origin_public_encryption_keys from builder_originsrv to builder
COPY 0
Copying origin_public_keys from builder_originsrv to builder
COPY 0
Copying origin_secret_keys from builder_originsrv to builder
COPY 0
Copying origin_secrets from builder_originsrv to builder
COPY 0
UPDATE 0
Copying audit from builder_jobsrv to builder
COPY 4815
Copying groups from builder_jobsrv to builder
COPY 12316
Copying group_projects from builder_jobsrv to builder
COPY 158982
Copying jobs from builder_jobsrv to builder
COPY 36587
Copying busy_workers from builder_jobsrv to builder
COPY 0
```

tools/merge-databases/merge-databases.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+# The purpose of this script is a one time data migration from
+# builder_originsrv and builder_sessionsrv -> builder
```
chefsalim: also jobsrv

we'll need this to also handle on-prem (no jobsrv) or create a different variant for on-prem. 

could you also open an issue to follow up on the on-prem work after this merges..

tools/merge-databases/merge-databases.sh
```diff
@@ -0,0 +1,87 @@
...
+  echo "Usage: merge-databases.sh"
+  exit 1
+}
+
+if [ -z "${PGPASSWORD:-}" ]; then
+  echo "Please ensure that PGPASSWORD is exported in your environment and run this script again."
+  exit 1
+fi
+
+echo "This script will take all of the data that's currently in builder_originsrv and builder_sessionsrv and merge them into builder."
```
chefsalim: also jobsrv

tools/merge-databases/merge-databases.sh
```diff
@@ -0,0 +1,87 @@
...
+
+for table in "${originsrv_tables[@]}"; do
+    echo "Copying ${table} from builder_originsrv to builder"
+    execute_sql builder_originsrv "\\copy ${table} to stdout" | execute_sql builder "\\copy ${table} from stdin"
+done
+
+execute_sql builder "UPDATE origin_packages SET ident_array=regexp_split_to_array(ident, '/');"
+
+for table in "${jobsrv_tables[@]}"; do
+    echo "Copying ${table} from builder_originsrv to builder"
```
chefsalim: jobsrv

components/builder-db/src/error.rs
chefsalim: I think the touch file approach is flawed in key ways - we add state to the services directory that needs to be carried around (I'm against doing this - thats what the datastore is for), and we create yet another way to give the services an "all clear" to start.  The scenario you are concerned about (a partial shard migration) is not a real world concern - let's say an on-prem admin runs shard-migrate.sh and it fails (partial migration).  Well, they are not going to be able to run your merge-databases.sh script (since you're going to check for proper flags for all services).  So at that point, what an admin will need to do is 1) run shard migration, then 2) run db migration.  Then, services will come up properly.  These 2 steps are exactly what we will document in the on-prem instructions.

chefsalim: The one thing that is tricky is that we added the check for no rows to specifically let brand new DBs start without the shard error.  So we have to think about how that is going to play out as well.  If we keep that as is, then there's no way to stop the on-prem install from bringing up new services and immediately starting with a fresh DB after this goes in :(

chefsalim: I can't think of any great solutions to this.  We may have to just put in a visible warning in the on-prem readme to shut down services immediately after re-install and do the db merge.  The db merge script will then have to check for the scenario of non-empty DB and and not assume that merge  db was run - instead it should halt the merge.

components/builder-jobsrv/src/server/handlers.rs
```diff
@@ -615,47 +615,6 @@ pub fn job_group_get(
     Ok(())
 }
 
-pub fn job_graph_package_create(
```
chefsalim: I think we still need this in some fashion, otherwise the graph is not going to get extended when new packages are added (uploaded)..

chefsalim: The simplest thing could be to just take the sync_packages function and run it from within the main loop of the scheduler thread.. you can then put the graph extension code in there

elliott-davis: yeah you're right, let me think about this some more

DEVELOPING.md
elliott-davis: https://github.com/habitat-sh/builder/issues/620

components/builder-core/src/file_walker.rs
```diff
@@ -13,20 +13,14 @@
 // limitations under the License.
 
 use hab_core::package::{FromArchive, PackageArchive};
```
elliott-davis: consider removing this file entirely?

components/builder-api/src/depot/server.rs
```diff
@@ -2425,8 +2418,18 @@ fn process_upload_for_package_archive(
...
-            return Err(err);
+        match conn.route::<OriginPackageCreate, OriginPackage>(&package) {
+            Ok(pkg) => {
+                let mut job_graph_package = JobGraphPackageCreate::new();
+                job_graph_package.set_package(pkg);
+
+                if let Err(err) =
+                    conn.route::<JobGraphPackageCreate, OriginPackage>(&job_graph_package)
+                {
+                    return Err(err);
```
raskchanky: If this particular branch is reached, it means that the `OriginPackageCreate` succeeded, but the `JobGraphPackageCreate` failed.  So effectively, we've created a package that doesn't exist in our package graph.  Handling that kind of error gracefully feels pretty tricky, and likely something we can punt on for this particular PR (since it will become irrelevant as the services are consolidated), but at the very least, we should log an error with `warn!`, so if this ends up biting us, we can track it down easier.

components/builder-core/src/file_walker.rs
raskchanky: From what I can tell, this file isn't used at all.  Delete it.

components/builder-jobsrv/src/migrations/2018-08-29-173340_initial_jobsrv_migration/up.sql
```diff
@@ -0,0 +1,393 @@
...
+
+CREATE FUNCTION check_active_group_v1(pname text) RETURNS SETOF groups
+    LANGUAGE sql
+    AS $$
+  SELECT * FROM groups
+  WHERE project_name = pname
+  AND group_state IN ('Pending', 'Dispatching')
+$$;
+
+CREATE FUNCTION count_group_projects_v2(origin text) RETURNS bigint
```
raskchanky: `CREATE FUNCTION IF NOT EXISTS`

components/builder-jobsrv/src/migrations/2018-08-29-173340_initial_jobsrv_migration/up.sql
```diff
@@ -0,0 +1,393 @@
...
+    LANGUAGE plpgsql STABLE
+    AS $$
+  BEGIN
+    RETURN COUNT(*) FROM group_projects WHERE project_ident LIKE (origin || '/%');
+  END
+$$;
+
+-- TED these two functions can be removed when the stats endpoint is gone
+CREATE FUNCTION count_origin_packages_v1(origin text) RETURNS bigint
```
raskchanky: `CREATE FUNCTION IF NOT EXISTS`

components/builder-jobsrv/src/migrations/2018-08-29-173340_initial_jobsrv_migration/up.sql
```diff
@@ -0,0 +1,393 @@
...
+    LANGUAGE plpgsql STABLE
+    AS $$
+BEGIN
+  RETURN COUNT(*) FROM origin_packages WHERE ident_array[1]=origin;
+END
+$$;
+
+CREATE FUNCTION count_unique_origin_packages_v1(op_origin text) RETURNS bigint
```
raskchanky: `CREATE FUNCTION IF NOT EXISTS`

components/builder-jobsrv/src/migrations/2018-08-29-173340_initial_jobsrv_migration/up.sql
```diff
@@ -0,0 +1,393 @@
...
+    LANGUAGE sql STABLE
+    AS $$
+  SELECT COUNT(DISTINCT ident_array[2]) AS total
+  FROM origin_packages
+  WHERE ident_array[1] = op_origin
+$$;
+
+CREATE FUNCTION delete_busy_worker_v1(in_ident text, in_job_id bigint) RETURNS void
```
raskchanky: `CREATE FUNCTION IF NOT EXISTS`

components/builder-jobsrv/src/data_store.rs
```diff
@@ -429,18 +407,21 @@ impl DataStore {
...
 
-    pub fn get_job_graph_package(&self, ident: &str) -> Result<jobsrv::JobGraphPackage> {
+    pub fn get_job_graph_package(&self, ident: &str) -> Result<originsrv::OriginPackage> {
         let conn = self.pool.get()?;
 
         let rows = &conn
-            .query("SELECT * FROM get_graph_package_v1($1)", &[&ident])
+            .query(
+                "SELECT * FROM get_origin_package_v4($1, $2)",
+                &[&ident, &String::from("public,private")],
```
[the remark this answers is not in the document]
elliott-davis: good catch

components/builder-jobsrv/src/migrations/2018-08-29-173340_initial_jobsrv_migration/up.sql
```diff
@@ -0,0 +1,393 @@
...
+    LANGUAGE sql
+    AS $$
+  DELETE FROM busy_workers
+  WHERE ident = in_ident AND job_id = in_job_id
+$$;
+
+CREATE FUNCTION find_group_project_v1(gid bigint, name text) RETURNS SETOF group_projects
```
raskchanky: `CREATE FUNCTION IF NOT EXISTS`

components/builder-jobsrv/src/migrations/2018-08-29-173340_initial_jobsrv_migration/up.sql
```diff
@@ -0,0 +1,393 @@
...
+    LANGUAGE plpgsql STABLE
+    AS $$
+BEGIN
+  RETURN QUERY SELECT * FROM group_projects WHERE owner_id = gid AND project_name = name;
+  RETURN;
+END
+$$;
+
+CREATE FUNCTION get_busy_workers_v1() RETURNS SETOF busy_workers
```
raskchanky: `CREATE FUNCTION IF NOT EXISTS`

components/builder-jobsrv/src/migrations/2018-08-29-173340_initial_jobsrv_migration/up.sql
```diff
@@ -0,0 +1,393 @@
...
+    job_state text,
+    project_id bigint,
+    project_name text,
+    project_owner_id bigint,
+    project_plan_path text,
+    vcs text,
+    vcs_arguments text[],
+    net_error_code integer,
+    net_error_msg text,
+    scheduler_sync boolean DEFAULT false,
```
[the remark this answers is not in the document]
elliott-davis: I think it is still needed and is used by the scheduler for building the graph. @chefsalim may know more

components/builder-jobsrv/src/migrations/2018-08-29-173340_initial_jobsrv_migration/up.sql
raskchanky: As well as the rest of the `CREATE FUNCTION` statements in this file 😁

components/builder-jobsrv/src/migrations/2018-08-29-173340_initial_jobsrv_migration/up.sql
```diff
@@ -0,0 +1,393 @@
...
+$$;
+
+CREATE FUNCTION get_busy_workers_v1() RETURNS SETOF busy_workers
+    LANGUAGE sql STABLE
+    AS $$
+  SELECT * FROM busy_workers
+$$;
+
+CREATE FUNCTION get_cancel_pending_jobs_v1() RETURNS SETOF jobs
+    LANGUAGE sql
```
raskchanky: `STABLE`

components/builder-originsrv/src/migrations/2018-08-24-184520_initial_originsrv_migration/up.sql
```diff
@@ -0,0 +1,1133 @@
...
+CREATE SEQUENCE IF NOT EXISTS origin_invitations_id_seq;
+CREATE SEQUENCE IF NOT EXISTS origin_private_encryption_key_id_seq;
+CREATE SEQUENCE IF NOT EXISTS origin_project_integration_id_seq;
+CREATE SEQUENCE IF NOT EXISTS origin_project_id_seq;
+CREATE SEQUENCE IF NOT EXISTS origin_public_key_id_seq;
+CREATE SEQUENCE IF NOT EXISTS origin_secret_key_id_seq;
+CREATE SEQUENCE IF NOT EXISTS origin_id_seq;
+CREATE SEQUENCE IF NOT EXISTS origin_public_encryption_key_id_seq;
+
+CREATE TABLE origins (
```
raskchanky: `CREATE TABLE IF NOT EXISTS`, for this and the rest

components/builder-originsrv/src/migrations/2018-08-24-184520_initial_originsrv_migration/up.sql
```diff
@@ -0,0 +1,1133 @@
...
+    origin_id bigint NOT NULL REFERENCES origins(id),
+    origin_name text,
+    account_id bigint NOT NULL,
+    account_name text,
+    created_at timestamp with time zone DEFAULT now(),
+    updated_at timestamp with time zone DEFAULT now(),
+    PRIMARY KEY (origin_id, account_id)
+);
+
+CREATE FUNCTION accept_origin_invitation_v1(oi_invite_id bigint, oi_ignore boolean) RETURNS void
```
raskchanky: `CREATE FUNCTION IF NOT EXISTS` for this and the rest

components/builder-originsrv/src/migrations/2018-08-24-184520_initial_originsrv_migration/up.sql
```diff
@@ -0,0 +1,1133 @@
...
+CREATE FUNCTION validate_origin_invitation_v1(oi_invite_id bigint, oi_account_id bigint) RETURNS TABLE(is_valid boolean)
+    LANGUAGE plpgsql
+    AS $$
+  BEGIN
+    RETURN QUERY SELECT true FROM origin_invitations WHERE id = oi_invite_id AND account_id = oi_account_id;
+    RETURN;
+  END
+  $$;
+
+CREATE VIEW origins_with_private_encryption_key_full_name_v1 AS
```
raskchanky: `CREATE VIEW IF NOT EXISTS` for this and the rest

components/builder-originsrv/src/migrations/2018-08-24-184520_initial_originsrv_migration/up.sql
```diff
@@ -0,0 +1,1133 @@
...
+ SELECT origins.id,
+    origins.name,
+    origins.owner_id,
+    origin_secret_keys.full_name AS private_key_name,
+    origins.default_package_visibility
+   FROM (origins
+     LEFT JOIN origin_secret_keys ON ((origins.id = origin_secret_keys.origin_id)))
+  ORDER BY origins.id, origin_secret_keys.full_name DESC;
+
+CREATE INDEX origin_packages_ident_array ON origin_packages(ident_array);
```
raskchanky: `CREATE INDEX IF NOT EXISTS`

components/builder-originsrv/src/server/mod.rs
```diff
@@ -310,12 +313,9 @@ impl Dispatcher for OriginSrv {
 
     fn app_init(
         config: Self::Config,
-        router_pipe: Arc<String>,
+        _router_pipe: Arc<String>,
```
raskchanky: Why not just remove this arg completely?

components/builder-sessionsrv/src/migrations/2018-08-24-193407_initial_sessionsrv_migration/up.sql
```diff
@@ -0,0 +1,84 @@
+CREATE SEQUENCE IF NOT EXISTS accounts_id_seq; 
+CREATE SEQUENCE IF NOT EXISTS account_tokens_id_seq;
+
+CREATE TABLE accounts (
```
raskchanky: `CREATE TABLE IF NOT EXISTS` for this and the rest

components/builder-sessionsrv/src/migrations/2018-08-24-193407_initial_sessionsrv_migration/up.sql
```diff
@@ -0,0 +1,84 @@
...
+);
+
+CREATE TABLE account_tokens (
+    id bigint DEFAULT next_id_v1('account_tokens_id_seq') PRIMARY KEY NOT NULL,
+    account_id bigint,
+    token text UNIQUE,
+    created_at timestamp with time zone DEFAULT now()
+);
+
+CREATE FUNCTION get_account_by_id_v1(account_id bigint) RETURNS SETOF accounts
```
raskchanky: `CREATE FUNCTION IF NOT EXISTS` for this and the rest

test/builder-api/bin/cleanup-integration-tests.sh
```diff
@@ -77,7 +77,6 @@ do
   sql=$(cat <<EOF
 SET SEARCH_PATH TO shard_0;
 DELETE FROM busy_workers WHERE job_id IN (SELECT id FROM jobs WHERE project_name LIKE '$origin%');
-DELETE FROM graph_packages WHERE ident LIKE '$origin%';
```
raskchanky: Why was this deleted?

test/builder-api/test.sh
```diff
@@ -60,9 +57,7 @@ clean_test_artifacts() {
   sql=
 
   for origin in "${origins[@]}"; do
-    sql+="SET SEARCH_PATH TO shard_0;"
     sql+="DELETE FROM busy_workers WHERE job_id IN (SELECT id FROM jobs WHERE project_name LIKE '$origin%');"
-    sql+="DELETE FROM graph_packages WHERE ident LIKE '$origin%';"
```
raskchanky: Why was this line removed?

tools/merge-databases/merge-databases.sh
```diff
@@ -0,0 +1,91 @@
...
+  echo "Usage: merge-databases.sh"
+  exit 1
+}
+
+if [ -z "${PGPASSWORD:-}" ]; then
+  echo "Please ensure that PGPASSWORD is exported in your environment and run this script again."
+  exit 1
+fi
+
+echo "This script will take all of the data that's currently in builder_originsrv, builder_sessionsrv and builder_jobsrv and merge them into builder."
```
raskchanky: It might be worth saying `merge them into a brand new database, called "builder"`, or something like that.  As it reads right now, it could be confusing.

components/builder-originsrv/src/server/mod.rs
elliott-davis: `app_init` is abstracted from the net crate so you have to accept 2 args, even if you don't use them. The other services still need the router pipe so it can't be refactored completely out.

raskchanky:
I left several dozen comments - apologies for that. Towards the end, I got a bit smarter and just indicated to follow the pattern, which you were already smart enough to do, but I tend to err on the side of being overly verbose. The next 🥃 is on me.

Overall, looks good.

components/builder-originsrv/src/server/mod.rs
raskchanky: Gotcha, makes sense.

test/builder-api/bin/cleanup-integration-tests.sh
elliott-davis: we deleted the table it was referencing https://github.com/habitat-sh/builder/pull/606/files#diff-5e7bf13e8c37dbad84f4d6524185aacbL332

test/builder-api/test.sh
elliott-davis: we deleted the table it was referencing https://github.com/habitat-sh/builder/pull/606/files#diff-5e7bf13e8c37dbad84f4d6524185aacbL332

test/builder-api/bin/cleanup-integration-tests.sh
raskchanky: Ah, of course.  I missed that.

Issue #608, opened 2018-08-25
Can't start sessionsrv or originsrv in an exported docker container


jamesc:
The run script fails for both builder-sessionsrv and builder-originsrv when running them in a container created using `hab pkg export docker ...`:

```
builder-originsrv.default(E): /hab/svc/builder-originsrv/run: line 6: syntax error: unexpected "("
hab-launch(SV): Child for service 'builder-originsrv.default' with PID 684 exited with code exit code: 2
```
This is because the `run` hook is using a bash array and the bash that is supplied with busybox isn't acutally bash, it's `ash` and that doesn't support arrays.

jamesc:
Fixed by #614

PR #609, opened 2018-08-26
Don't use arrays in a run hook

jamesc:
This allows a real bash to be used when exported to a docker
container rather than relying on busybox (or system bash if run
directly on a node)

Fix for #608 

### DIFF
--- a/components/builder-api/habitat/hooks/run
+++ b/components/builder-api/habitat/hooks/run
@@ -3,13 +3,12 @@
 export HOME="{{pkg.svc_data_path}}"
 export RUST_LOG="{{cfg.log_level}}"
 export RUST_BACKTRACE=1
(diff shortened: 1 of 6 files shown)


thesentinels:

Thank you for contributing!

chefsalim:
This seems reasonable @jamesc .  Will thumbs it after tests pass.  Can you re-base off master to grab a Travis update I did to see if we can get the tests to pass.

jamesc:
@chefsalim I did the rebase but the same error

chucklehunk:
@jamesc instead of adding a new dep on core/bash I think it'd be a bit better to "re-hydrate" the hooks and remove the bash-isms. You can basically remove ` pkg_svc_run=(bldr-router start -c "{{pkg.svc_config_path}}/config.toml")` and sub in the contents of the array for the interpolated values further down in the file.

jamesc:
closing to re-open within habitat origin so travis can run

PR #614, opened 2018-08-28
Don't use arrays in run hooks

jamesc:
When exported to a docker container `/bin/bash` is actually ash which
 doesn't support arrays.

Fix for #608


thesentinels:

Thank you for contributing!

PR #621, opened 2018-08-31
Migrate builder-api from Iron to Actix-web

chefsalim:
This change migrates builder-api from Iron to the Actix-web framework.  In addition, there are significant improvements and refactoring to make the code base easier to manage and develop that are described below.  The switchover preserves backward compatibility for users of the builder API.  All (167 currently) e2e test cases (including a significant number of tests that were previously skipped/pending that have now been re-enabled) pass.

Why?
* Move away from older-gen mostly non-maintained framework 
* Move to a newer-gen, better supported, easier to develop for infra
* Have cleaner direct deps (= remove net of 14 crates from Cargo.toml)
* Much cleaner / smaller boilerplate code
* Leverage new framework features/patterns (as per below)
* Opportunity to leverage async for better performance for long-running/blocking APIs
* Opportunity to clean up code structure
* Opportunity to review and clean up API routes
* Have a better foundation for future work to come

Key changes:
* The API codebase structure is now logically split up into resources (break the monoliths)
* Each resource is responsible for configuring it's paths - no monolithic path list
* A Logger middleware is added to the application in order to get automatic call logging 
* We no longer insert middleware for things like OAuth, Github client, etc - they are easily retrieved via the AppState from all requests
* Config is simplified and available via the AppState everywhere - the DepotUtil trait is removed
* Extractors are used to reduce or eliminate boilerplate code for Path params
* Stateless Authentication middleware makes the route construction simpler - there is no more need to add auth middleware on a case-by-case basis when constructing routes
* Authorization is better encapsulated into 'authorize_session' that combines the ability to check for authenticated user, that is a part of an origin (optional)
* Monolith handlers like list_packages are split up  - in general handler functions are not re-used across routes any more
* The helpers.rs module that had become a big grab bag of things is now much slimmed down
* New metrics have been added to track a couple of highest traffic routes (more to come)
* Error handling is better consolidated into error.rs, and is much more simpler to convert from error to a HttpResponse

Net net there's quite a bit of change here. In order to perhaps make the review simpler, for now, all commits have been left as is, but they will be squashed into a few logical commits before merge.

tenor-87889944


thesentinels:

Thank you for contributing!

components/builder-originsrv/src/migrations/2018-08-23-153639_insert_origin_package_v4/up.sql
```diff
@@ -0,0 +1,31 @@
+CREATE OR REPLACE FUNCTION insert_origin_package_v4 (
```
elliott-davis: No... :joy:

components/builder-api/src/server/services/github.rs
```diff
@@ -45,7 +46,7 @@ pub enum GitHubEvent {
 impl FromStr for GitHubEvent {
     type Err = Error;
 
-    fn from_str(event: &str) -> Result<Self, Self::Err> {
+    fn from_str(event: &str) -> error::Result<Self, Error> {
```
eeyun: I think this can just be `fn from_str(event: &str) -> Result<Self, Error>` since we are importing the result from the error.rs

components/builder-api/src/server/resources/channels.rs
```diff
@@ -14,371 +14,326 @@
...
-    let package = match route_message::<OriginPackageGet, OriginPackage>(req, &request) {
-        Ok(p) => p,
-        Err(e) => {
-            error!("Error retrieving package: {:?}", &e);
-            return Ok(render_net_error(&e));
+    //
+    // Route handlers - these functions should return HttpResponse
+    //
+    pub fn get_channels(
+        (req, sandbox): (HttpRequest<AppState>, Query<SandboxBool>),
```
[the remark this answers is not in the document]
eeyun: Actually looking back through the old code it _does_ appear that this value is an option so this MUST get resolved before merging.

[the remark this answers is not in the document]
eeyun: For future work cleaning this up: https://github.com/actix/actix-web/issues/223 might be good to use either a HashMap or the .query method (maybe in an expression on the route registration or maybe not - more tinkering needed).

elliott-davis:
`/v1/depot/origins/core` fails with a panic

components/builder-api/src/server/services/github.rs
chefsalim: No, it's specifically not our Result.. it takes 2 arguments..

components/builder-originsrv/src/migrations/2018-08-23-153639_insert_origin_package_v4/up.sql
chefsalim: Sad but needed to make the force upload idempotent..

PR #627, opened 2018-09-07
Remove/replace set_{permissions,owner} calls which aren't supported on windows

baumanj:
(NB: This will be easier to review with whitespace changes hidden)

These are misleading since the core implementation doesn't actually do what the
function names imply, so replace with more illustrative code on a per-platform
basis. Long-term, we should implement the appropriate windows FFI calls, but
the current situation is dangerously misleading.

See https://github.com/habitat-sh/core/blob/master/components/core/src/os/filesystem/windows.rs#L28-L34

### DIFF
--- a/components/builder-worker/src/runner/mod.rs
+++ b/components/builder-worker/src/runner/mod.rs
@@ -401,7 +401,11 @@ impl Runner {
                 let dst = res.unwrap();
                 debug!("Imported origin secret key, dst={:?}.", dst);
                 if self.config.airlock_enabled {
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

raskchanky:
@eeyun I'm guessing it's the builder version of https://github.com/habitat-sh/habitat/pull/5591

baumanj:
What @raskchanky said. It's not required on builder, but as long as I was doing the `habitat` work, it seemed worthwhile to fix code which is misleading.

Issue #612, opened 2018-08-28
Builder shouldn't assume a target of `x86_64-linux` when unable to parse UserAgent


smacfarlane:
When running a query against the /latest api endpoint,  if Builder is unable to parse the target from the user agent, it assumes a target of `x86_64-linux`. [code](https://github.com/habitat-sh/builder/blob/master/components/builder-api/src/depot/server.rs#L2319-L2322)

This assumption breaks `hab pkg install` when developing a new `PackageTarget` as builder is unaware of the target and will returns results for `x86_64-linux` which the hab cli will then download and attempt to install, breaking the build process. 

In addition, if I was a Windows user hitting the latest endpoint in my browser, I would expect any results it returns to be Windows packages.  

I think that we should return a helpful error message when encountering an unknown user agent, letting the user know about the `target` query parameter, rather than make an assumption about what the user wants.

PR #632, opened 2018-09-12
Dont default target when not present

chefsalim:
We no longer assume that the package target is "x86_64-linux" if it is not specified in by a valid target param, or in a parsable User Agent header on the request.  This resolves https://github.com/habitat-sh/builder/issues/612

tenor-26585737


thesentinels:

Thank you for contributing!

PR #639, opened 2018-09-12
Make latest calls work again

chefsalim:
Since we removed default fallback to x86_64-linux for the package latest calls, we need to add explicit target to the caller. Until we have a more permanent solution that is being worked on, this tweak adds a default target of x86_64-linux to the web UI so that things can continue to function at parity.


thesentinels:

Thank you for contributing!

PR #643, opened 2018-09-14
Sa/perf tuning

chefsalim:
Add some more perf tuning options to nginx config, and enable a keep-alive config for the builder-api.

tenor-33570378

### DIFF
--- a/components/builder-api-proxy/habitat/config/nginx.conf
+++ b/components/builder-api-proxy/habitat/config/nginx.conf
@@ -125,7 +125,9 @@ http {
       rewrite ^(.*)$ https://$host$1 permanent;
     }
 
+    {{~#if cfg.server.enable_access_log}}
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

raskchanky:
The first thing that went through my head after reviewing this PR was wondering if the keep-alive config is going to exacerbate the thread-bound problems we've experienced.

chefsalim:
It's not clear, the idea is to have a tunable knob for keep alive, which according to nginx perf tuning docs is a pretty key config to enable.  We are currently using http 1.0 to the upstream without keep alives. So if we can improve the upstream performance, we improve the rate at which we respond back to clients which only seems like a good thing..

raskchanky:
I'm guessing we're using at least HTTP 1.1, otherwise CORS wouldn't work.  But I guess as long as it's tunable, it should be ok.

PR #651, opened 2018-09-15
Check for bldr account id for session name

chefsalim:
We need to check for builder account id when getting session name in order to remove un-needed backend call, and to remove spurious warning.


thesentinels:

Thank you for contributing!

PR #661, opened 2018-09-18
Remove CORS support

cnunciato:
This change removes CORS support from builder-api and removes the configurability of the `habitat_api_url` setting. It also adds support for proxying Builder API services in development in order to work around this new restriction.

### DIFF
--- a/components/builder-web/bs-config.js
+++ b/components/builder-web/bs-config.js
@@ -1,0 +1,13 @@
+const proxyMiddleware = require('http-proxy-middleware');
+
+module.exports = {
+  server: {
+    middleware: [
+      proxyMiddleware('/v1', {
+        target: 'http://localhost:9636',
+        logLevel: 'debug'
+      })
+    ]
+  },
+  startPath: '/#/pkgs/core'
+};
--- a/components/builder-web/package-lock.json
+++ b/components/builder-web/package-lock.json
@@ -522,6 +522,12 @@
       "integrity": "sha1-104bh+ev/A24qttwIfP+SBAasjQ=",
       "dev": true
     },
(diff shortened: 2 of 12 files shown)


thesentinels:

Thank you for contributing!

components/builder-web/package-lock.json
```diff
@@ -4302,6 +4307,256 @@
         "extend": "3.0.1"
       }
     },
+    "http-proxy-middleware": {
```
chefsalim: Are we including this in the prod package as well?  Not clear whether this is dev only..

components/builder-web/bs-config.js
```diff
@@ -0,0 +1,13 @@
+const proxyMiddleware = require('http-proxy-middleware');
```
chefsalim: Are there any specific steps for devs to take advantage of this?

components/builder-web/package-lock.json
cnunciato: Nope, it's a `devDependency`, so it's only needed in dev.

components/builder-web/bs-config.js
cnunciato: This makes me wonder whether the API URL even needs to be configurable anymore. I don't think it does, so I'll update this PR to remove that as well.

cnunciato:
All right @chefsalim, this should take care of what's needed for `builder-api-proxy`. I'm looking at removing the CORS support in `builder-api` now.

elliott-davis:
We need to add cors back into the API otherwise GitHub auth is going to break _i think_

chefsalim:
Both webhooks and auth are ok on acceptance

PR #665, opened 2018-09-20
Cache session tokens in memcache rather than a hashmap in originsrv.

elliott-davis:
This change moves session token validation to the edge rather than routing messages all the way back to originsrv.
Performance city - here we come - CHOO CHOO

### DIFF
--- a/components/builder-api/src/server/framework/middleware.rs
+++ b/components/builder-api/src/server/framework/middleware.rs
@@ -23,13 +23,14 @@ use protobuf;
 use bldr_core;
 use bldr_core::metrics::CounterMetric;
 use hab_net::conn::RouteClient;
-use hab_net::{ErrCode, NetError, NetOk};
+use hab_net::{ErrCode, NetError};
 use oauth_client::types::OAuth2User;
 use protocol;
 use protocol::originsrv::*;
 use protocol::Routable;
 
 use server::error;
+use server::resources::profile::do_get_access_tokens;
 use server::services::metrics::Counter;
 use server::services::route_broker::RouteBroker;
 use server::AppState;
@@ -98,12 +99,60 @@ fn authenticate(req: &HttpRequest<AppState>, token: &str) -> error::Result<Sessi
 
     // Check for a valid personal access token
     if bldr_core::access_token::is_access_token(token) {
-        let session =
-            bldr_core::access_token::validate_access_token(&req.state().config.api.key_path, token)
-                .map_err(|_| NetError::new(ErrCode::BAD_TOKEN, "net:auth:bad-token"))?;
-        revocation_check(req, session.get_id(), token)
-            .map_err(|_| NetError::new(ErrCode::BAD_TOKEN, "net:auth:revoked-token"))?;
-        return Ok(session);
+        let mut memcache = req.state().memcache.borrow_mut();
+        match memcache.get_session(token) {
+            Some(session) => {
+                trace!("Session {} Cache Hit!", token);
+                return Ok(session);
+            }
+            None => {
+                trace!("Session {} Cache Miss!", token);
+                // Pull the session out of the current token provided so we can validate
+                // it against the db's tokens
+                let session = bldr_core::access_token::validate_access_token(
+                    &req.state().config.api.key_path,
+                    token,
+                ).map_err(|_| NetError::new(ErrCode::BAD_TOKEN, "net:auth:bad-token"))?;
+
+                // If we can't find a token in the cache, we need to round-trip to the
+                // db to see if we have a valid session token.
+                match do_get_access_tokens(&req, session.get_id()) {
+                    Ok(access_tokens) => {
+                        assert!(access_tokens.get_tokens().len() <= 1); // Can only have max of 1 for now
+                        match access_tokens.get_tokens().first() {
+                            Some(access_token) => {
+                                let new_token = access_token.get_token();
+                                if token.trim_right_matches('=')
+                                    != new_token.trim_right_matches('=')
+                                {
+                                    // Token is valid but revoked or otherwise expired
+                                    return Err(error::Error::NetError(NetError::new(
+                                        ErrCode::BAD_TOKEN,
+                                        "net:auth:revoked-token",
+                                    )));
+                                }
+                                memcache.set_session(new_token, &session);
+                                return Ok(session);
+                            }
+                            None => {
+                                // We have no tokens in the database for this user
+                                return Err(error::Error::NetError(NetError::new(
+                                    ErrCode::BAD_TOKEN,
+                                    "net:auth:revoked-token",
+                                )));
+                            }
+                        }
+                    }
+                    Err(_) => {
+                        // Failed to fetch tokens from the database for this user
+                        return Err(error::Error::NetError(NetError::new(
+                            ErrCode::BAD_TOKEN,
+                            "net:auth:revoked-token",
+                        )));
+                    }
+                }
+            }
+        };
     };
 
     // Check for internal sessionsrv token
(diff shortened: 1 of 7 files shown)


thesentinels:

Thank you for contributing!

components/builder-api/src/server/framework/middleware.rs
```diff
@@ -98,12 +99,62 @@ fn authenticate(req: &HttpRequest<AppState>, token: &str) -> error::Result<Sessi
...
+            }
+            None => {
+                trace!("Session {} Cache Miss!", token);
+                // Pull the session out of the current token provided so we can validate
+                // it against the db's tokens
+                let mut session = bldr_core::access_token::validate_access_token(
+                    &req.state().config.api.key_path,
+                    token,
+                ).map_err(|_| NetError::new(ErrCode::BAD_TOKEN, "net:auth:bad-token"))?;
+                // If we can't find a token in the cache, we need to round-trip to the
```
chefsalim: whitesp

components/builder-api/src/server/framework/middleware.rs
```diff
@@ -98,12 +99,62 @@ fn authenticate(req: &HttpRequest<AppState>, token: &str) -> error::Result<Sessi
...
+                // db to see if we have a valid session token. The first time this is used,
+                // it is only to get the account_id so we can do the lookup
+                match do_get_access_tokens(&req, session.get_id()) {
+                    Ok(access_tokens) => match access_tokens.get_tokens().first() {
```
chefsalim: assert access_tokens length == 1

components/builder-api/src/server/framework/middleware.rs
```diff
@@ -98,12 +99,62 @@ fn authenticate(req: &HttpRequest<AppState>, token: &str) -> error::Result<Sessi
...
+                        Some(access_token) => {
+                            let new_token = access_token.get_token();
+                            if token.trim_right_matches('=') != new_token.trim_right_matches('=') {
+                                // Token is valid but revoked or otherwise expired
+                                return Err(error::Error::NetError(NetError::new(
+                                    ErrCode::BAD_TOKEN,
+                                    "net:auth:revoked-token",
+                                )));
+                            }
+                            // Unfortuantely, we need to pull the session out of the
```
chefsalim: Not clear why we need to do a second round trip.. if the token pulled from DB matches the token we are passed in, we are good - just add it to the cache

components/builder-api/src/server/framework/middleware.rs
```diff
@@ -130,21 +181,9 @@ fn authenticate(req: &HttpRequest<AppState>, token: &str) -> error::Result<Sessi
...
-    request.set_account_id(account_id);
-    request.set_token(token.to_owned());
-    route_message::<AccountTokenValidate, NetOk>(req, &request)?;
-    Ok(())
-}
-
 fn session_validate(req: &HttpRequest<AppState>, token: SessionToken) -> error::Result<Session> {
     let mut request = SessionGet::new();
-    request.set_token(token);
+    request.set_token(token.clone());
```
chefsalim: we need this clone?

components/builder-api/src/server/framework/middleware.rs
elliott-davis: What does that net us if we are matching on some or none for `.first()`

chefsalim: It's not a huge deal - but it would be a false negative if there were multiple tokens and the second token was the matching one.

components/builder-api/src/server/framework/middleware.rs
```diff
@@ -98,12 +99,56 @@ fn authenticate(req: &HttpRequest<AppState>, token: &str) -> error::Result<Sessi
...
+                // Pull the session out of the current token provided so we can validate
+                // it against the db's tokens
+                let session = bldr_core::access_token::validate_access_token(
+                    &req.state().config.api.key_path,
+                    token,
+                ).map_err(|_| NetError::new(ErrCode::BAD_TOKEN, "net:auth:bad-token"))?;
+
+                // If we can't find a token in the cache, we need to round-trip to the
+                // db to see if we have a valid session token. The first time this is used,
+                // it is only to get the account_id so we can do the lookup
```
chefsalim: This comment isn't clear - the account_id is already available as the session id..

components/builder-api/src/server/resources/profile.rs
```diff
@@ -97,6 +103,18 @@ fn generate_access_token(req: HttpRequest<AppState>) -> HttpResponse {
         session.get_flags()
     };
 
+    // Memcache supports multiple tokens but to preserve legacy behavior
+    // we must purge any existing tokens before generating new ones
+    match do_get_access_tokens(&req, account_id) {
+        Ok(access_tokens) => {
+            let mut memcache = req.state().memcache.borrow_mut();
+            for token in access_tokens.get_tokens() {
+                memcache.delete_key(token.get_token())
```
chefsalim: I wonder if there is a race condition here.. the time between we delete from cache and we delete from DB.  We may want to move this AFTER we delete from DB.

components/builder-api/src/server/resources/profile.rs
```diff
@@ -128,9 +146,24 @@ fn revoke_access_token(req: HttpRequest<AppState>) -> HttpResponse {
...
+        Err(err) => return err.into(),
+    };
+
     let mut request = AccountTokenRevoke::new();
     request.set_id(token_id);
 
+    match do_get_access_tokens(&req, account_id) {
+        Ok(access_tokens) => {
+            let mut memcache = req.state().memcache.borrow_mut();
+            for token in access_tokens.get_tokens() {
```
chefsalim: Same potential race condition

components/builder-originsrv/src/server/mod.rs
```diff
@@ -333,7 +329,6 @@ lazy_static! {
 pub struct ServerState {
     datastore: DataStore,
     sessions: Arc<Box<RwLock<HashSet<Session>>>>,
```
chefsalim: Do we need the sessions set here either?  We may be able to get rid of this later (different PR)

elliott-davis: I wasn't sure - I figured I'd keep my scope of impact smaller on this PR and look into moving more session stuff over after this lands.

components/builder-api/src/server/framework/middleware.rs
chefsalim: well it may be even better to change do_get_access_tokens to do_get_access_token that only returns one .. we probably put in more infra for multiple tokens than we need right now

components/builder-originsrv/src/server/mod.rs
chefsalim: yep agreed

components/builder-api/src/server/resources/profile.rs
```diff
@@ -116,7 +122,20 @@ fn generate_access_token(req: HttpRequest<AppState>) -> HttpResponse {
...
     match route_message::<AccountTokenCreate, AccountToken>(&req, &request) {
-        Ok(account_token) => HttpResponse::Ok().json(account_token),
+        Ok(account_token) => {
+            // Memcache supports multiple tokens but to preserve legacy behavior
+            // we must purge any existing tokens AFTER generating new ones
+            match do_get_access_tokens(&req, account_id) {
+                Ok(access_tokens) => {
+                    let mut memcache = req.state().memcache.borrow_mut();
+                    for token in access_tokens.get_tokens() {
+                        memcache.delete_key(token.get_token())
```
chefsalim: still not correct.. we need to delete the old token, not the new one... so we have to get the access token prior to the DB call... then delete the old token from cache after the DB call.

components/builder-api/src/server/resources/profile.rs
```diff
@@ -128,11 +147,27 @@ fn revoke_access_token(req: HttpRequest<AppState>) -> HttpResponse {
...
+        Err(err) => return err.into(),
+    };
+
     let mut request = AccountTokenRevoke::new();
     request.set_id(token_id);
 
     match route_message::<AccountTokenRevoke, NetOk>(&req, &request) {
-        Ok(_) => HttpResponse::Ok().finish(),
+        Ok(_) => {
+            match do_get_access_tokens(&req, account_id) {
```
chefsalim: same as above - get and save the old token, delete it from cache after the DB call

components/builder-api/src/server/resources/profile.rs
elliott-davis: oh I see now - LOL that last change would have been a fun failure.

PR #667, opened 2018-09-20
Update caching strategy for channel packages

chefsalim:
This change updates the cache strategy for channel packages.  Previously we were caching only the origin package object, however that still left some backend calls uncovered. This change caches the full json response.

tenor-94015203


thesentinels:

Thank you for contributing!

PR #671, opened 2018-09-21
Use mouseup for Integration buttons

cnunciato:
I can’t say I’m completely sure _why_ Angular Material is canceling the `click` event in Chrome for these buttons, but it is &mdash; so this fix has them use the `mouseup` event instead.


thesentinels:

Thank you for contributing!

PR #684, opened 2018-09-26
Elliott/package channels

elliott-davis:


thesentinels:

Thank you for contributing!

components/builder-api/src/server/resources/package_channels.rs
```diff
@@ -77,16 +78,24 @@ impl PackageChannels {
...
     let (origin, channel, pkg, version, release) =
         Path::<(String, String, String, String, String)>::extract(&req)
             .unwrap()
             .into_inner();
 
-    match do_promote_package(&req, origin, channel, pkg, version, release) {
-        Ok(_) => HttpResponse::new(StatusCode::OK),
-        Err(err) => err.into(),
-    }
+    let ident = format!("{}/{}/{}/{}", origin, pkg, version, release);
```
elliott-davis: might be a better way to handle this - if there is, I'd love to know about it

christophermaier:
```rust
let i = PkgIdent::new(origin, pkg, Some(version), Some(release));
let i_as_string = format!("{}", i);
```
Bonus poitns if `PromotePackage` could take a `PkgIdent` rather than a string.

components/builder-api/src/server/resources/channels.rs
```diff
@@ -58,38 +45,6 @@ impl Channels {
                 "/depot/channels/{origin}/{channel}",
                 Method::DELETE,
                 delete_channel,
-            ).route(
```
chefsalim: I suggest we leave these routes here and not break up channels without breaking up the route prefix as well - each resource should have all routes at the same prefix in one place

components/builder-api/src/server/resources/package_channels.rs
chefsalim: I think we could leverage PackageIdent for this

PR #688, opened 2018-09-27
Fix id serialization for channel ID's coming from the database

elliott-davis:
Fun fact: your browser can't handle 64 bit integers even though they are completely valid in json.
2nd fun fact: all of our db IDs are 64 bit integers.

Previously we implemented the serialize trait for each message struct that would get sent over the wire that converted u64 to string.
I hvae written a serialize function that you can use as a decorator in your structs to tell serde to convert an i64 to string.

Ex:

```
use server::model::db_id_format;

struct Foo {
  #[serde(with = "db_id_format")]
  id: i64,
  name: String,
  #[serde(with = "db_id_format")]
  owner_id: i64,
}

//initialize foo

json!(foo) -> renders quoted i64
```

### DIFF
--- a/components/builder-api/src/server/models/channel.rs
+++ b/components/builder-api/src/server/models/channel.rs
@@ -1,3 +1,4 @@
+use super::db_id_format;
 use actix_web::{actix::Message, Error};
 use chrono::NaiveDateTime;
 use diesel;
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

fnichol:
Is there a reason we're using signed integers (i.e. `i64`) vs. unsigned integers (i.e. `u64`)? As far as I can tell identifiers here shouldn't ever be negative and we lose half the capacity of the type when signed.

raskchanky:
@fnichol It's an artifact of PG.  PG's BigInt type is a signed 64bit integer.

fnichol:
Ah, that makes a bit more sense, thanks!

PR #698, opened 2018-10-01
Clarify some details for setting up Builder dev

cnunciato:
This adds a few clarifying details, along with a section on how to prepare an empty depot to run local builds.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -53,12 +53,18 @@ If you are on a Mac, you will need brew, direnv, habitat, and Docker for Mac
 
 ### Studio requirements
 
-1. Ensure you have run `hab setup` in your environment that will be executing the studio
+Ensure you have run `hab setup` in the environment that will be executing the studio, or exported environment variables of `HAB_ORIGIN` and `HAB_AUTO_TOKEN` whose values correspond with the public Builder service. For example:
+
+```
+export HAB_ORIGIN=habitat
+export HAB_AUTH_TOKEN=<YOUR_PERSONAL_ACCESS_TOKEN>
+```
 
 ### Starting the services
 
 From either your VM or Mac:
 
+* `sudo -i # Or your preferred method of running as root`
 * `cd <source path>`
 * `direnv allow`
 * `hab studio enter`
@@ -140,7 +146,7 @@ You may use the `load_package` helper to specify a package to upload. Ex:
 load_package /hab/cache/artifacts/core-*.hart
 ```
 
-If you are missing packages, see [Setting up an upstream source](#setting-up-an-upstream-source) below.
+If you are missing packages, see [Setting up an upstream source](#setting-up-a-package-upstream) below.
 
 #### Option A: From the Web UI
 * Navigate to http://${APP_HOSTNAME}/#/pkgs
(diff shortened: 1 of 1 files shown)


DEVELOPING.md
```diff
@@ -53,12 +53,18 @@ If you are on a Mac, you will need brew, direnv, habitat, and Docker for Mac
 
 ### Studio requirements
 
-1. Ensure you have run `hab setup` in your environment that will be executing the studio
+Ensure you have run `hab setup` in the environment that will be executing the studio, or exported environment variables of `HAB_ORIGIN` and `HAB_AUTO_TOKEN` whose values correspond with the public Builder service. For example:
+
+    ```
+    export HAB_ORIGIN=<YOUR_PERSONAL_ORIGIN>
```
chefsalim: The origin may need to be 'habitat' - the package builds within dev studio doesn't seem to work otherwise.

DEVELOPING.md
```diff
@@ -140,7 +146,7 @@ You may use the `load_package` helper to specify a package to upload. Ex:
 load_package /hab/cache/artifacts/core-*.hart
```
chefsalim: we should just be able to use the load_package helper for backline, no?

DEVELOPING.md
```diff
@@ -198,3 +206,51 @@ hab pkg install -u http://localhost -z <auth-token> <package>
...
+      hab origin key download core --secret
+      ```
+
+  1. Upload your locally installed version of `hab-backline` &mdash; for example, for Habitat 0.64.1:
+
+      ```
+      hab pkg upload /hab/cache/artifacts/core-hab-backline-0.64.1-20180928012546-x86_64-linux.hart
+      ```
+
+  1. Upload the public key with which that version of `hab-backline` was built:
```
chefsalim: this should happen automatically via the hab client

DEVELOPING.md
```diff
@@ -198,3 +206,51 @@ hab pkg install -u http://localhost -z <auth-token> <package>
...
+
+      ```
+      hab origin key upload --pubfile /hab/cache/keys/core-20160810182414.pub
+      ```
+
+  1. And then finally, in your local instance of Builder UI, promote that version of `core/hab-backline` to the stable channel.
```
chefsalim: promote can be done via upload by using the -c stable flag, no need for a separate step

cnunciato: I didn't have to do it that way (presumably because the deps are fully qualified), but it does seem better to promote all of them at the same time, yeah.

DEVELOPING.md
cnunciato: I was able to `bb` everything with `HAB_ORIGIN` set as my personal, but setting it as `habitat` makes more sense, anyway. I'll update.

DEVELOPING.md
```diff
@@ -198,3 +206,51 @@ hab pkg install -u http://localhost -z <auth-token> <package>
...
+    hab origin key download core --secret
+    ```
+
+  1. Browse to your local instance of Builder UI (e.g., http://localhost), sign in, navigate to your Profile and generate a new personal access token (if you haven't already).
+
+  1. Since you no longer need to interact with public Builder, export a set of environment variables to point the `hab` client to your new, local instance of Builder:
+
+      ```
+      export HAB_ORIGIN=<YOUR_PERSONAL_ORIGIN> # If you've created one
+      export HAB_ORIGIN=<YOUR_PERSONAL_ACCESS_TOKEN>
```
chefsalim: Couple of comments here: 1) the export presumably should be HAB_AUTH_TOKEN instead of HAB_ORIGIN  2) we should be clear not to do this within the studio, as it would then start pinging prod depot with the a bad auth token

cnunciato:
@chefsalim All right, should be all good now. Thanks for the guidance on this.

PR #701, opened 2018-10-02
Add datadog service startup event

chefsalim:
Adding a service startup event datadog event to builder-api.  This will be expanded to the other services eventually after we try this out in acceptance and prod.

tenor-96861068

### DIFF
--- a/components/builder-api/habitat/hooks/run
+++ b/components/builder-api/habitat/hooks/run
@@ -4,6 +4,10 @@ export HOME="{{pkg.svc_data_path}}"
 export RUST_LOG="{{cfg.log_level}}"
 export RUST_BACKTRACE=1
 
+title="Service Start"
+text="builder-api starting"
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/builder-api/habitat/hooks/run
```diff
@@ -4,6 +4,10 @@ export HOME="{{pkg.svc_data_path}}"
 export RUST_LOG="{{cfg.log_level}}"
 export RUST_BACKTRACE=1
 
+title="Service Start"
```
elliott-davis: Should we also add a post-run hook to indicate that builder-api has started?

chefsalim: I don't think we strictly need it right now.. eg, if the service is thrashing, we'll see a spike of service starts.  But we can certainly consider adding more events in hooks once we're happy with the shape of this one.

PR #706, opened 2018-10-05
 Sessionsrv migration to api

eeyun:
Woo doggies. This was a massive miserable bag of spaghetti snakes. This is a major simplification and @elliott-davis wrote like 99% of this PR.

### DIFF
--- a/components/builder-api/src/server/error.rs
+++ b/components/builder-api/src/server/error.rs
@@ -48,7 +48,7 @@ pub enum Error {
     Connection(conn::ConnErr),
     BadRequest(String),
     DieselError(diesel::result::Error),
-    GetDbPool(r2d2::Error),
+    DbPoolError(r2d2::Error),
     Github(HubError),
     InnerError(io::IntoInnerError<io::BufWriter<fs::File>>),
     Protocol(protocol::ProtocolError),
@@ -88,7 +88,7 @@ impl fmt::Display for Error {
             }
             Error::Connection(ref e) => format!("{}", e),
             Error::DieselError(ref e) => format!("{}", e),
-            Error::GetDbPool(ref e) => format!("{}", e),
+            Error::DbPoolError(ref e) => format!("{}", e),
             Error::Github(ref e) => format!("{}", e),
             Error::InnerError(ref e) => format!("{}", e.error()),
             Error::ParseIntError(ref e) => format!("{}", e),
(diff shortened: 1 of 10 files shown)


thesentinels:

Thank you for contributing!

components/builder-api/src/server/error.rs
```diff
@@ -45,6 +47,8 @@ pub enum Error {
     CircularDependency(String),
     Connection(conn::ConnErr),
     BadRequest(String),
+    DieselError(diesel::result::Error),
+    GetDbPool(r2d2::Error),
```
chefsalim: Minor - maybe make this error more generic - eg, DbError(r2d2::Error)

components/builder-api/src/server/framework/middleware.rs
```diff
@@ -100,7 +104,7 @@ fn authenticate(req: &HttpRequest<AppState>, token: &str) -> error::Result<Sessi
     // Check for a valid personal access token
     if bldr_core::access_token::is_access_token(token) {
         let mut memcache = req.state().memcache.borrow_mut();
-        match memcache.get_session(token) {
+        match memcache.get_access_token(token) {
```
chefsalim: This is excellent.. I think we can make it even simpler.  Just have a single get_session/set_session memcache API.  If the cache hits, great.  If the cache misses, check whether the token is a personal access token, and if so, do the validation according to the None branch below.  If the token is not a personal access token, then return session expired.

chefsalim:
Overall this is some really great work - it's such an improvement over the current sessionsrv.

raskchanky:
This is a total nitpick but I might suggest changing the commit message to remove "WIP" and maybe add some more details about what was done.

components/builder-api/src/server/services/memcache.rs
```diff
@@ -112,10 +112,10 @@ impl MemcacheClient {
         self.reset_namespace(&channel_ns_key(origin, channel));
     }
 
-    pub fn get_session(&mut self, user: &str) -> Option<Session> {
-        trace!("Getting session for user {} from memcached", user);
+    pub fn get_token(&mut self, token: &str) -> Option<Session> {
```
chefsalim: Nitpicking a bit, but the "get_session" naming seems more accurate, and more consistent with the other function naming here (eg, get/set package)..

PR #710, opened 2018-10-08
Clarify the wording around job cancellation

cnunciato:
The wording we’re using now is a bit inconsistent; this uses the word “remaining” to include jobs that have either not yet begun or are in progress, and changes the Cancel button to use something other than “cancel”, because the thing you’re confirming happens to be a cancellation.

Also ensures we animate the icon in the jobs list when applicable.

### DIFF
--- a/components/builder-web/app/shared/dialog/job-cancel/job-cancel.dialog.html
+++ b/components/builder-web/app/shared/dialog/job-cancel/job-cancel.dialog.html
@@ -5,16 +5,16 @@
   </section>
   <section class="body">
     <p>
-      Completed build jobs will remain in the <b>unstable</b> channel. Pending jobs will not be processed.
+      Completed packages will remain in the <b>unstable</b> channel. Remaining jobs will be canceled.
     </p>
     <p>
-      Build jobs to be canceled: <b>{{ pendingCount }}</b>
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

components/builder-web/app/shared/dialog/job-cancel/job-cancel.dialog.html
```diff
@@ -5,16 +5,16 @@ <h1>Confirm Job Cancellation</h1>
   </section>
   <section class="body">
     <p>
-      Completed build jobs will remain in the <b>unstable</b> channel. Pending jobs will not be processed.
+      Completed packages will remain in the <b>unstable</b> channel. Remaining jobs will be canceled or terminated.
```
chefsalim: Not sure if there is value in calling out "cancelled or terminated"?  Maybe just say "Remaining jobs will be canceled" and leave it at that?

cnunciato: Ok. I just noticed that cancellation also seems to stop individual jobs that are currently running, so I figured it might be good (or not bad) to mention that also.

PR #719, opened 2018-10-10
Remove usage of the stats endpoint to get origin package counts

elliott-davis:
Since the only place we were using the stats endpoint was to list the package counts for an origin I have completely removed it from the API.

This will also allow on-prem users to see their package counts for their origins.


thesentinels:

Thank you for contributing!

eeyun:
Not that this is a blocker but I do have a couple of questions about this.
First, instead of grabbing package stats and publishing them to an API now you're effectively rendering just the package count somewhere in the UI?

Second, and possibly more importantly - are we removing this endpoint because it's not useful? Is it actually not useful, or is it just not being effectively utilized? I can see a reality where having an origin's stats endpoint could be super useful if the right pieces of data are being exposed. I'd hate to see this removed primarily for simplification of the API if there is future work that could be useful in here.

elliott-davis:
@eeyun to answer your first question, yes. Instead of sending back just the origin row from the origins table, I'm now sending back the row, plus the package count in the `/users/origins` endpoint.

This feature was added when we had an overall stats view in the UI. Upon looking at the metrics for page usage from users we decided to remove that view because it was so underutilized. We didn't remove the accompanying API because it was still in use to get the origin package counts for the "My Origins" view. The query that was removed was an extremely expensive call to the database and was called multiple times when a user clicked anywhere on an origin view. So, while this does reduce our overall API surface area it also reduces load on the database and enables on-prem users to see package totals (since the stats API was only exposed if you had jobsrv enabled)

components/builder-api/src/server/models/origins.rs
```diff
@@ -0,0 +1,43 @@
...
+    pub owner_id: i64,
+    pub name: String,
+    pub created_at: Option<NaiveDateTime>,
+    pub updated_at: Option<NaiveDateTime>,
+    pub default_package_visibility: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, QueryableByName)]
+#[table_name = "origins_with_stats"]
+pub struct OriginWithStats {
```
chefsalim: I'm not sure I like the pattern of creating a different OriginWithStats struct... can we not modify the Origin struct?

chefsalim:
Looks great, just one comment above

components/builder-api/src/server/models/origins.rs
elliott-davis: The struct has to match the return type of the function unfortunately. I think we could fix this if we used regular diesel functions for querying and joining but by using stored procedures the compiler tries to type-check based on the returned "table" data. When we return origin rows in the future without a package count, we'll need the regular origin struct. I made an attempt at a struct like: 

```
pub struct OriginsWithStats {
  pub origin: Origin,
  pub package_count: i64
} 
```
but I couldn't get the return types right from the sql function.

PR #724, opened 2018-10-12
Expose and use router-less HTTP-based RPC interface to Job server

chefsalim:
This change adds a new HTTP transport for protobuf RPC. The new RPC infrastructure is now leveraged by the job server and builder-api for all JobSrv Protocol RPC. This eliminates the need to route the jobsrv protocol through the router.  Note that originsrv is still using the ZMQ based RPC, so it will still use the router for now.

tenor-212205207

### DIFF
--- a/.studiorc
+++ b/.studiorc
@@ -205,7 +205,7 @@ start-cache() {
 }
 
 declare -A svc_params=(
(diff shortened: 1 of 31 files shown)


thesentinels:

Thank you for contributing!

Issue #500, opened 2018-06-22
Upstream depot tweaks


raskchanky:
We've gotten feedback from a few customers that our current implementation of upstream depot syncing is sub-optimal.  Some tweaks to improve it are:

- [ ] Allow the calling process to optionally block until the upstream sync is finished, rather than doing it asynchronously in the background.  This could be accomplished with a new command line flag to `hab pkg install`.
- [ ] Allow any version of any package to sync, not just the latest version.
- [ ] For the background async installation, rather than doing it in a 60 second loop, do it on demand.

chefsalim:
Tracking upstream issues and replacement plan in https://github.com/habitat-sh/on-prem-builder/issues/105

PR #725, opened 2018-10-12
Remove upstream package syncing functionality

elliott-davis:
The overall design of this is confusing to users. For now, we are removing it
and will replace it with a standalone script/service that lets users sync
packages synchronously


thesentinels:

Thank you for contributing!

raskchanky:
There's an issue that documents the confusing parts of this feature and improvements that could be made. https://github.com/habitat-sh/builder/issues/500

elliott-davis:
So just for clarification on the commit message. @chefsalim is actively working on a stand-alone feature to sync packages on demand that should cover everything from #500. We're removing this functionality from the API now because it unblocks more of the simplification of services work around uploading packages.

chefsalim:
I think we mis-designed this feature on a number of levels and we're going to be better off with a clean start with much clearer requirements and scenarios - both user facing as well as operational.  There are some fundamental issues that need to be resolved that will take some time.  In the meantime,  the lack of things like not syncing dependencies and keys means we can essentially leave depots in a bad state if a key package syncs but its deps are not present (until those are specifically searched for or attempted to be installed).  Adding more (potentially synchronous) code to do that sort of functionality as a background thread that lives in the UI nodes is not the model that I want to invest in.  Also the operational clarity for managing such a sync service needs to be better thought through than the logs we have now.  Net net, I think we are better off removing this code now, unblocking our simplification work, adding some scripts as a temporary bridge workaround until we get a solution in place that can be a delighter, and not worrying about sunk costs.

PR #728, opened 2018-10-16
Adds helper for executing functional tests in builder devenv

eeyun:

### DIFF
--- a/.studiorc
+++ b/.studiorc
@@ -273,6 +273,41 @@ origin() {
   fi
 }
 
+test-builder() {
+    hab sup status
+    if [ $? -eq 0 ]; then
+        cat <<DOCS
+
+*****************"
+Running supervisor detected
+*****************"
+ 
+*****************"
+Terminating running supervisor
+*****************"
+DOCS
+        hab sup term
+        sleep 8
+    fi
+
+    echo ""
+    echo "*****************"
+    echo "Restarting supervisor in test mode"
+    echo "*****************"
+    echo ""
+    HAB_FUNC_TEST=1 sup-run
+    sleep 10
+    echo "*****************"
+    echo "Executing functional tests"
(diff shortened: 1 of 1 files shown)


.studiorc
```diff
@@ -273,6 +273,35 @@ origin() {
   fi
 }
 
+test-builder() {
+    hab sup status
+    if [ $? -eq 0 ]; then
```
elliott-davis: Is this a reliable way to check if the sup is running? I see on line 91 of this file we check the text output. If it is a reliable way, both ways of checking for sup aliveness should match

.studiorc
```diff
@@ -273,6 +273,35 @@ origin() {
...
+        echo ""
```
elliott-davis: Consider using HereDoc for multiline output

.studiorc
```diff
@@ -273,6 +273,35 @@ origin() {
...
+    echo "Restarting supervisor in test mode"
+    echo "*****************"
+    echo ""
+    HAB_FUNC_TEST=1 sup-run
+    sleep 10
+    echo "*****************"
+    echo "Executing functional tests"
+    echo "*****************"
+    /src/test/builder-api/test.sh
+}
```
elliott-davis: Should we stop test mode after the tests have run?

.studiorc
eeyun: This is a reliable way. It checks the exit code of running the previous command. Since we're checking up/down-ness from an expected unclean state the intention of these two checks are slightly different. The up check on line 91 is intended for a clean state and validates both that the sup is up AND that there are services loaded. In the case of 278, we are only validating that sup is up.

.studiorc
eeyun: Good call. Added.

PR #729, opened 2018-10-16
Better error handling for job logs

chefsalim:
This change removes error JSON from the body of some errors (specifically, NetErrors) that were bubbling up via the API.  The UI code will be amended to remove the specific usage, which will no longer be necessary due to another change in this PR, which is to not have the job log return a failure if it has not been written yet. A number of job log tests are now re-enabled, and a test is added to check for the availability of a blank job log for newly created build jobs.

tenor-189611255


thesentinels:

Thank you for contributing!

components/builder-api/src/server/error.rs
```diff
@@ -191,7 +191,7 @@ impl Into<HttpResponse> for Error {
             Error::BadRequest(_) => HttpResponse::new(StatusCode::BAD_REQUEST),
             Error::Github(_) => HttpResponse::new(StatusCode::FORBIDDEN),
             Error::CircularDependency(_) => HttpResponse::new(StatusCode::FAILED_DEPENDENCY),
-            Error::NetError(ref e) => HttpResponse::build(net_err_to_http(&e)).json(&e),
+            Error::NetError(ref e) => HttpResponse::new(net_err_to_http(&e)),
```
elliott-davis: I think @cnunciato needed this to get the polling on log watch working. A 404 wasn't enough to tell him that the request was dead but to keep watching so he was using the `{"code": 4}` in the response body

chefsalim: Right, so there's more more 404 for missing job logs.  The only remaining case is if there's a legit failure.  That said, @cnunciato may have another commit to add to this.

PR #736, opened 2018-10-22
Update Cargo.lock

mpeck:
This updates the Cargo.lock file and makes some tweaks to accommodate updates to `hab_core` and the `users` crate.

The changes to `airlock` are due to some updates to types in the `users` crate. I used https://github.com/habitat-sh/core/commit/80b351d25b3b3c57bbce40a50822556b669b6284 as the example.

The changes to `builder-worker` come from some updates that were made to the `hab_core` api: https://github.com/habitat-sh/core/commit/c71807b825b36aebb9226eaccd628de0077e07d4#diff-acba3894e00237cdf440982b2ce1f5dfL15.


components/airlock/src/user.rs
```diff
@@ -45,11 +45,15 @@ pub fn home_dir_for_username(username: &str) -> Result<PathBuf> {
 }
 
 pub fn my_username() -> Result<String> {
-    users::get_effective_username().ok_or(Error::EffectiveUsernameNotFound)
+    users::get_effective_username()
```
baumanj: Since the code to convert from OsString → String [already exists in the core crate](https://github.com/habitat-sh/core/blob/master/components/core/src/os/users/linux.rs#L48), why not use it?

components/airlock/src/util.rs
```diff
@@ -98,7 +99,7 @@ pub fn check_user_group_membership(username: &str) -> Result<()> {
     for grp in vec!["tty"].iter() {
         let user_group =
             users::get_group_by_name(grp).ok_or(Error::GroupNotFound(String::from(*grp)))?;
-        if !user_group.members().contains(&user) {
+        if !user_group.members().contains(&OsString::from(&user)) {
```
baumanj: I think you can eliminate the need to construct a new `user` string and use just the `&str` parameter with the implicit conversion:
```suggestion
        if !user_group.members().contains(&username.into()) {
```
And with this, you may not need the `use std::ffi::OsString;` either.

thesentinels:

Thank you for contributing!

components/builder-worker/src/runner/mod.rs
```diff
@@ -34,7 +34,10 @@ use bldr_core::logger::Logger;
 use chrono::Utc;
 use hab_core::os::users;
 use hab_core::package::archive::PackageArchive;
-use hab_core::util::perm;
+#[cfg(not(windows))]
```
baumanj: I think this should be `#[cfg(unix)]` rather than `not(windows)` since the `posix_perm` module depends on unix libraries, not just the absence of windows.

chefsalim:
It might be nice to squash the commits.. otherwise looks good

components/builder-worker/src/runner/mod.rs
```diff
@@ -34,7 +34,10 @@ use bldr_core::logger::Logger;
...
+use hab_core::util::posix_perm as perm;
+#[cfg(windows)]
```
baumanj: What happens if you leave the `windows` include out? Since this isn't currently built for windows, I think it should work.

components/builder-worker/src/server.rs
```diff
@@ -18,7 +18,10 @@ use std::iter::FromIterator;
 use std::sync::Arc;
 
 use hab_core::users;
-use hab_core::util::perm;
+#[cfg(not(windows))]
```
baumanj: Same comments as above.

components/airlock/src/user.rs
christophermaier: I suspect there may have been a desire to not depend on any Habitat-specific code for `airlock`, though @fnichol would know that for sure.

components/airlock/src/util.rs
mpeck: That doesn't work. I'm not sure that I completely understand how the borrower works in rust yet, but I think that consumes the value so it's then not available to be used down on line 103.

components/builder-worker/src/runner/mod.rs
mpeck: I don't disagree, but the code that implements it uses `#[cfg(not(windows))]`. Shouldn't those match?

components/builder-worker/src/runner/mod.rs
mpeck: It builds and runs fine for me, but of course I'm running it on Linux so it should. I didn't know that this wasn't built for windows. I can totally remove it if it's not necessary.

components/builder-worker/src/runner/mod.rs
baumanj: I would argue that code should also be `#[cfg(unix)]`

components/airlock/src/util.rs
baumanj: Huh, I'm not sure without looking deeper what the issue is, but if it's not a simple change, the way you have it should work.

components/builder-worker/src/runner/mod.rs
mpeck: Yeah, that's a good argument :) There are several places that use it though. Would it be reasonable to do another PR changing all of those calls?

PR #739, opened 2018-10-23
Ignore target info when package ident is fully qualified

elliott-davis:
If we get a fully qualified ident we shouldn't use the latest function to query the db.
Instead, we should use the exact match function from the database.


thesentinels:

Thank you for contributing!

chefsalim:
tenor-71090656

PR #740, opened 2018-10-23
Convert origin projects to direct DB calls

chefsalim:
This change migrates the project-related APIs over to direct DB calls.

tenor-102190634

### DIFF
--- a/components/builder-api/src/server/error.rs
+++ b/components/builder-api/src/server/error.rs
@@ -176,6 +176,7 @@ impl ResponseError for Error {
             Error::OAuth(_) => HttpResponse::new(StatusCode::UNAUTHORIZED),
             Error::ParseIntError(_) => HttpResponse::new(StatusCode::BAD_REQUEST),
             Error::Protocol(_) => HttpResponse::new(StatusCode::INTERNAL_SERVER_ERROR),
+            Error::DieselError(ref e) => HttpResponse::new(diesel_err_to_http(&e)),
 
             // Default
             _ => HttpResponse::new(StatusCode::UNPROCESSABLE_ENTITY),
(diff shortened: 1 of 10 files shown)


thesentinels:

Thank you for contributing!

components/builder-api/src/server/error.rs
```diff
@@ -176,6 +176,7 @@ impl ResponseError for Error {
             Error::OAuth(_) => HttpResponse::new(StatusCode::UNAUTHORIZED),
             Error::ParseIntError(_) => HttpResponse::new(StatusCode::BAD_REQUEST),
             Error::Protocol(_) => HttpResponse::new(StatusCode::INTERNAL_SERVER_ERROR),
+            Error::DieselError(ref e) => HttpResponse::new(diesel_err_to_http(&e)),
```
elliott-davis: ooh you fancy huh?

components/builder-db/src/models/projects.rs
```diff
@@ -0,0 +1,111 @@
...
+    #[serde(with = "db_id_format")]
+    pub id: i64,
+    #[serde(with = "db_id_format")]
+    pub origin_id: i64,
+    #[serde(with = "db_id_format")]
+    pub owner_id: i64,
+    pub origin_name: Option<String>,
+    pub package_name: Option<String>,
+    pub name: Option<String>,
+    pub plan_path: Option<String>,
```
elliott-davis: I see in the create/update structs below a lot of the things that are optional here are required there. Does it make sense to convert these to not use `Options`?

components/builder-api/src/server/resources/projects.rs
```diff
@@ -115,33 +111,37 @@ fn create_project((req, body): (HttpRequest<AppState>, Json<ProjectCreateReq>))
         Err(err) => return err.into(),
     };
 
+    let conn = match req.state().db.get_conn() {
```
elliott-davis: This is not your fault but I'd really like to find a way to map a `Pool` error to an HttpResponse of Internal server error since we cargo cult this around all over the place

components/builder-api/src/server/resources/projects.rs
```diff
@@ -270,37 +269,46 @@ fn update_project((req, body): (HttpRequest<AppState>, Json<ProjectUpdateReq>))
...
             "Missing value for field: `plan_path`",
         );
     }
 
+    let conn = match req.state().db.get_conn() {
+        Ok(conn_ref) => conn_ref,
+        Err(_) => return HttpResponse::new(StatusCode::INTERNAL_SERVER_ERROR),
+    };
+
+    // TODO (SA): We should not need to fetch the origin project here.
```
elliott-davis: This TODO is why the migrations file is exploding. So many unnecessary calls to get ID's

components/builder-db/src/models/projects.rs
chefsalim: Yeah that would be better if it works - I've only wired these as Options to match the DB columns.

components/builder-api/src/server/resources/projects.rs
chefsalim: Sure I can take a look at doing that

PR #744, opened 2018-10-25
Make latest channel package call a direct-db call from the API

elliott-davis:

### DIFF
--- a/components/builder-api/src/server/resources/channels.rs
+++ b/components/builder-api/src/server/resources/channels.rs
@@ -36,9 +36,9 @@ use server::framework::middleware::route_message;
 use server::helpers::{self, Pagination, Target};
 // TED PROTOCLEANUP remove aliases when we get rid of all the protos
 use db::models::channel::{
-    Channel, CreateChannel, DeleteChannel, ListChannels, OriginChannelDemote as OCD,
-    OriginChannelPackage as OCPackage, OriginChannelPromote as OCP, PackageChannelAudit as PCA,
-    PackageChannelAudit, PackageChannelOperation as PCO,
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

components/builder-protocol/src/originsrv.rs
```diff
@@ -744,10 +744,14 @@ impl From<hab_core::package::PackageIdent> for OriginPackageIdent {
         ident.set_origin(value.origin);
         ident.set_name(value.name);
         if let Some(ver) = value.version {
-            ident.set_version(ver);
+            if ver != "" {
```
chefsalim: Is this check here primarily to work around for bad data in the db?

elliott-davis: exactly, since we could have potentially poisoned the db with the other bug fixed in here

components/builder-api/src/server/resources/channels.rs
```diff
@@ -527,31 +527,11 @@ fn do_get_channel_package(
...
-            match route_message::<OriginChannelPackageLatestGet, OriginPackageIdent>(req, &request)
-            {
-                Ok(id) => id.into(),
-                Err(err) => return Err(err),
-            };
-    }
-
-    let pkg = match Package::get(
-        GetPackage {
+    let pkg = match Channel::get_latest_package(
```
chefsalim: yes!

components/builder-protocol/src/originsrv.rs
chefsalim: would be good to also clean up the db after

PR #747, opened 2018-10-26
Fix regression; return private key name with origin

chefsalim:
This fixes a regression where we were not returning the private key name with the origin request, thereby causing the UI to not account for keys properly.  Also adds an e2e test for this scenario.

tenor-217203149


thesentinels:

Thank you for contributing!

PR #753, opened 2018-10-30
Sa/migrate account

chefsalim:
Migrate accounts and invitations to direct DB

tenor-255421298

### DIFF
--- a/components/builder-db/src/models/account.rs
+++ b/components/builder-db/src/models/account.rs
@@ -1,3 +1,5 @@
+use super::db_id_format;
+use chrono::NaiveDateTime;
 use diesel;
 use diesel::pg::PgConnection;
 use diesel::result::QueryResult;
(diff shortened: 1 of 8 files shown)


thesentinels:

Thank you for contributing!

components/builder-db/src/models/account.rs
```diff
@@ -1,77 +1,100 @@
...
         diesel::sql_query("SELECT update_account_v1($1, $2)")
-            .bind::<BigInt, _>(account.id)
-            .bind::<Text, _>(account.email)
+            .bind::<BigInt, _>(id as i64)
+            .bind::<Text, _>(email)
+            .execute(conn)
+    }
+}
+
+#[derive(Insertable)]
```
elliott-davis: Do you need to derive insertable here if we're not using the query dsl?

components/builder-db/src/models/invitations.rs
```diff
@@ -25,13 +25,63 @@ pub struct OriginInvitation {
     pub updated_at: Option<NaiveDateTime>,
 }
 
+#[derive(Insertable)]
```
elliott-davis: Same question as above

components/builder-db/src/models/account.rs
chefsalim: I'm leaving those in mostly as a placeholder for experimenting with direct inserts

PR #768, opened 2018-11-01
More direct DB migration for Project and fix for group promote/demote

chefsalim:
Fix a breakage in group project promote/demote and convert Project visibility to an enum

tenor-164312402

### DIFF
--- a/components/builder-api/src/server/resources/projects.rs
+++ b/components/builder-api/src/server/resources/projects.rs
@@ -135,7 +135,7 @@ fn create_project((req, body): (HttpRequest<AppState>, Json<ProjectCreateReq>)) 
             vcs_type: "git",
             vcs_data: "https://github.com/habitat-sh/testapp.git",
             install_id: body.installation_id as i64,
-            visibility: &originsrv::OriginPackageVisibility::Public.to_string(),
+            visibility: &PackageVisibility::Public,
             auto_build: body.auto_build,
         };
 
@@ -205,7 +205,7 @@ fn create_project((req, body): (HttpRequest<AppState>, Json<ProjectCreateReq>)) 
         vcs_type: "git",
         vcs_data: &vcs_data,
         install_id: body.installation_id as i64,
-        visibility: &origin.default_package_visibility.to_string(),
+        visibility: &origin.default_package_visibility,
         auto_build: body.auto_build,
     };
 
@@ -304,7 +304,7 @@ fn update_project((req, body): (HttpRequest<AppState>, Json<ProjectUpdateReq>)) 
             vcs_type: "git",
             vcs_data: "https://github.com/habitat-sh/testapp.git",
             install_id: body.installation_id as i64,
-            visibility: &originsrv::OriginPackageVisibility::Public.to_string(),
+            visibility: &PackageVisibility::Public,
             auto_build: body.auto_build,
         };
 
@@ -617,7 +617,7 @@ fn toggle_privacy(req: HttpRequest<AppState>) -> HttpResponse {
         vcs_type: &project.vcs_type,
         vcs_data: &project.vcs_data,
         install_id: project.vcs_installation_id,
-        visibility: &pv.to_string(),
+        visibility: &pv,
         auto_build: project.auto_build,
     };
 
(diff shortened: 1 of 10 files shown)


thesentinels:

Thank you for contributing!

components/builder-db/src/schema/project.rs
```diff
@@ -11,7 +12,7 @@ table! {
         owner_id -> BigInt,
         created_at -> Nullable<Timestamptz>,
         updated_at -> Nullable<Timestamptz>,
-        visibility -> Text,
+        visibility -> PackageVisibilityMapping,
```
elliott-davis: I don't see a migration for this? I suspect it will fail?

components/builder-db/src/models/channel.rs
```diff
@@ -202,23 +202,23 @@ impl PackageChannelAudit {
...
-impl PackageGroupChannelAudit {
+impl<'a> PackageGroupChannelAudit<'a> {
     pub fn audit(req: PackageGroupChannelAudit, conn: &PgConnection) -> QueryResult<usize> {
         diesel::sql_query(
-            "select * from add_audit_package_group_entry_v1($1, $2, $3, $4, $5, $6, $7, $8)",
-        ).bind::<BigInt, _>(req.origin_id)
-        .bind::<BigInt, _>(req.channel_id)
+            "select * from add_audit_package_group_entry_v2($1, $2, $3, $4, $5, $6, $7, $8)",
+        ).bind::<Text, _>(req.origin)
+        .bind::<Text, _>(req.channel)
```
elliott-davis: YES

components/builder-api/src/server/resources/projects.rs
```diff
@@ -205,7 +205,7 @@ fn create_project((req, body): (HttpRequest<AppState>, Json<ProjectCreateReq>))
         vcs_type: "git",
         vcs_data: &vcs_data,
         install_id: body.installation_id as i64,
-        visibility: &origin.default_package_visibility.to_string(),
+        visibility: &origin.default_package_visibility.into(),
```
elliott-davis: is the into necessary here? The origin model should return PackageVisibility by default

components/builder-api/src/server/resources/projects.rs
```diff
@@ -135,7 +135,7 @@ fn create_project((req, body): (HttpRequest<AppState>, Json<ProjectCreateReq>))
             vcs_type: "git",
             vcs_data: "https://github.com/habitat-sh/testapp.git",
             install_id: body.installation_id as i64,
-            visibility: &originsrv::OriginPackageVisibility::Public.to_string(),
+            visibility: &originsrv::OriginPackageVisibility::Public.into(),
```
elliott-davis: maybe use PackageVisibility::Public here?

components/builder-db/src/schema/project.rs
chefsalim: This is copied over from packages, but I'm not seeing where it's defined either

components/builder-api/src/server/resources/projects.rs
chefsalim: Good catch

components/builder-api/src/server/resources/projects.rs
chefsalim: Yep

components/builder-db/src/schema/project.rs
elliott-davis: The DbEnum derives it. If you don't change the column type in the database to the `origin_package_visibility` enum, you're going to have a bad time

chefsalim: Right, so that's in the migration
```
ALTER TABLE origin_projects ALTER COLUMN visibility SET DEFAULT 'public'::origin_package_visibility;
```

elliott-davis: example here: https://github.com/habitat-sh/builder/blob/master/components/builder-originsrv/src/migrations/2018-10-10-198563_enum_visibility/up.sql#L5-L10

chefsalim: Ah ok, will update.. it definitely works fine without the type change.. it leaves the underlying type as text.

PR #770, opened 2018-11-01
Redo search from the ground up using full text search

elliott-davis:
Move search into a direct db diesel query. Unfortuantely there are some trait bound
issues that prevents us from merging search and search_distinct into the same function.

Ignore the duplication and you'll be just fineeee.


thesentinels:

Thank you for contributing!

PR #771, opened 2018-11-01
Move migrations from originsrv to builder-db

elliott-davis:
Delete originsrv, router, most of net and the op tool.


elliott-davis:
Because the bindings changed for jobsrv and api this will not pass integration tests until we promote those packages. I've verified that all the integration tests pass locally

PR #774, opened 2018-11-02
Migrate off postgres functions and on to the diesel ORM

elliott-davis:
Move the remaining API calls to postgres functions to leverage diesel directly
Leave the postgres functions in place for now so we don't break jobsrv


thesentinels:

Thank you for contributing!

components/builder-db/src/models/channel.rs
```diff
@@ -264,18 +287,47 @@ pub struct OriginChannelDemote {
...
     pub fn promote(package: OriginChannelPromote, conn: &PgConnection) -> QueryResult<usize> {
         Counter::DBCall.increment();
-        diesel::sql_query("select * from promote_origin_package_v3($1, $2, $3)")
-            .bind::<Text, _>(package.origin)
-            .bind::<Text, _>(package.ident.to_string())
-            .bind::<Text, _>(package.channel)
-            .execute(conn)
+        // If this looks bad, it is. To ensure we get values here or die we have to execute queries to get the IDs first.
+        // I can hear the groaning already, "Why can't we just do a sub-select and let the database barf on insert"
+        // Great question - Because the typechecking happens in Rust and you wanted a type-safe language
```
weiznich: As discussed in the diesel gitter, this catches the possibility that one of the subqueries return no value. If that's impossible because of certain database constrains it is possible to trick diesel into accepting that query:

To do this you need to extend diesel by some custom type magic:

```rust

struct AssertNotNull<T, ST>(T, ::std::marker::PhantomData<ST>);

fn unsafe_assert_not_null<T, ST>(t: T) -> AssertNotNull<T, ST>
where T: Expression<SqlType = Nullable<ST>>,
      ST: NotNull
{
    AssertNotNull(t, Default::default())
}

impl<T, ST> Expression for AssertNotNull<T, ST>
where T: Expression<SqlType = Nullable<ST>>,
      ST: NotNull,
{
    type SqlType = ST;
}

impl<DB, T, ST> QueryFragment<DB> for AssertNotNull<T, ST>
where T: QueryFragment<DB>,
      DB: Backend
{

    fn walk_ast(&self, pass: AstPass<DB>) -> Result<(), Error> {
        self.0.walk_ast(pass)
    }
}
```

This defines some type that changes the sql type of a nullable expression to not nullable. After that just call `unsafe_assert_not_null` on both queries and it will work.

components/builder-api/src/server/resources/channels.rs
```diff
@@ -106,14 +106,7 @@ fn get_channels((req, sandbox): (HttpRequest<AppState>, Query<SandboxBool>)) ->
...
 
-    match Channel::list(
-        ListChannels {
-            origin: origin,
-            include_sandbox_channels: sandbox.is_set,
-        },
-        &*conn,
-    ).map_err(Error::DieselError)
-    {
+    match Channel::list(origin, sandbox.is_set, &*conn).map_err(Error::DieselError) {
```
chefsalim: 👍 removing the extra structs

components/builder-db/src/migrations/2018-11-05-111111_diesel_query_improvements/up.sql
```diff
@@ -0,0 +1,93 @@
+DROP FUNCTION IF EXISTS get_account_by_name_v1(text);
```
chefsalim: Would be good to have a related PR for on-prem to ensure we don't break the DB merge migration with schema changes

components/builder-db/src/models/account.rs
```diff
@@ -38,35 +36,39 @@ pub struct NewAccount<'a> {
...
-            .bind::<Text, _>(name)
-            .bind::<Text, _>(email)
+    pub fn find_or_create(account: &NewAccount, conn: &PgConnection) -> QueryResult<Account> {
+        match accounts::table
+            .filter(accounts::name.eq(account.name))
+            .limit(1)
             .get_result(conn)
+        {
+            Ok(account) => Ok(account),
+            Err(_) => diesel::insert_into(accounts::table)
```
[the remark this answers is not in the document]
elliott-davis: It turns out when there is a conflict postgres doesn't return the row. I did flip the logic around to try and insert first, then select on error

components/builder-db/src/models/package.rs
```diff
@@ -93,24 +91,20 @@ pub const ALL_COLUMNS: AllColumns = (
...
+#[table_name = "origin_packages"]
 pub struct NewPackage {
-    #[serde(with = "db_id_format")]
-    pub origin_id: i64,
+    pub origin: String,
     #[serde(with = "db_id_format")]
     pub owner_id: i64,
     pub name: String,
     pub ident: BuilderPackageIdent,
+    pub ident_array: Vec<String>,
```
chefsalim: Having both ident and ident array seems redundant?

components/builder-db/src/models/package.rs
```diff
@@ -245,29 +239,40 @@ impl Package {
...
-        .get_result(conn)
+        let package = match diesel::insert_into(origin_packages::table)
+            .values(&package)
+            .returning(ALL_COLUMNS)
+            .on_conflict_do_nothing()
+            .get_result::<Package>(conn)
+        {
+            Ok(pkg) => pkg,
+            Err(e) => {
+                debug!("PACKAGE ERROR: {:?}", e);
```
chefsalim: Either remove debug or update text to be more descriptive - eg "Create package error: {:?}" ?

components/builder-db/src/models/package.rs
```diff
@@ -379,13 +383,13 @@ impl Package {
...
         Counter::DBCall.increment();
-        // So many regrets with this query
-        diesel::sql_query(
-            "select $1 as origin, $2 as name, version, release_count, latest, string_to_array(platforms, ',') as platforms from get_origin_package_versions_for_origin_v8($1, $2, $3)",
-        ).bind::<Text, _>(&ident.origin)
-        .bind::<Text, _>(&ident.name)
-        .bind::<Array<PackageVisibilityMapping>, _>(visibility)
-        .get_results(conn)
+
+        origin_package_versions::table
```
chefsalim: Nice!

components/builder-db/src/models/package.rs
elliott-davis: Previously, we were splitting on slashes at insertion time. Now, if we want to insert that array we have to split it outside. Also, if we want to use insertable we have to have all columns present in the struct that we want to set. 

So yes, it is redundant but necessary. We might be able to do an audit to see where we're using ident in the code now to see if we could drop that column in the future.

components/builder-db/src/models/project_integration.rs
```diff
@@ -70,20 +73,58 @@ impl ProjectIntegration {
         conn: &PgConnection,
     ) -> QueryResult<usize> {
         Counter::DBCall.increment();
-        diesel::sql_query("select * from delete_origin_project_integration_v1($1, $2, $3)")
-            .bind::<Text, _>(origin)
-            .bind::<Text, _>(name)
-            .bind::<Text, _>(integration)
-            .execute(conn)
+        //delete_origin_project_integration_v1
```
chefsalim: remove comment?

components/builder-db/src/models/secrets.rs
```diff
@@ -3,63 +3,62 @@ use chrono::NaiveDateTime;
...
 impl OriginSecret {
-    pub fn create(
-        origin_id: i64,
-        name: &str,
-        value: &str,
-        conn: &PgConnection,
-    ) -> QueryResult<usize> {
+    pub fn create(secret: &NewOriginSecret, conn: &PgConnection) -> QueryResult<usize> {
         Counter::DBCall.increment();
-        // TODO FIX: We're missing setting the account id here
```
chefsalim: Maybe we should drop the account_id column from the origin secrets table?

PR #775, opened 2018-11-03
Perf and monitoring improvements to builder-api

chefsalim:
This PR makes a few key improvements to open up perf for the direct DB builder API.  Since getting a DB connection/making a DB call are blocking operations, we need to ensure we avoid those in the high volume paths. With the changes here, we are able to serve all high volume calls (including authentication/authorization via personal access tokens) without hitting the DB.  Leveraging memcache for origin membership also allows us to make some other key simplifications. Finally, counter metrics are added for every DB API in the builder-db models.

tenor-268071921

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1015,6 +1015,7 @@ dependencies = [
 name = "habitat_builder_db"
 version = "0.0.0"
 dependencies = [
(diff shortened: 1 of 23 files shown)


thesentinels:

Thank you for contributing!

PR #782, opened 2018-11-05
Fix list secret regression

chefsalim:
This fixes a regression where origin secrets list was failing, due to a Diesel trying to map a null owner_id. Currently we don't write out the owner id for secrets, so that column is unused.

tenor-21168069


thesentinels:

Thank you for contributing!

PR #783, opened 2018-11-07
Better proxy support for builder-api

chefsalim:
This PR improves http/https proxy support in builder.  It uses the env_proxy crate to determine the proxy settings, which takes no_proxy into account. The use case is primarily for on-prem builder.

tenor-9234196


thesentinels:

Thank you for contributing!

PR #798, opened 2018-11-10
Fix target handling for package upload

chefsalim:
Bug fix and some minor cleanup to package upload target handling - previously we were falling back to a default Linux target if no target was specified by the client on upload - however that failed for Windows packages where the hab client is not explicitly specifying a target in the url.

tenor-84017365


raskchanky:
On the upside, if the rust syntax wasn't correct, it wouldn't compile 😁

PR #800, opened 2018-11-14
Remove fanout for channel package call

chefsalim:
This change improves the perf on the channel package call by removing the internal fanout of calls that were being made to retrieve channels and platforms. Instead, it leverages the new packages_with_channel_platform view to perform the same operation.

tenor-266416807


thesentinels:

Thank you for contributing!

PR #805, opened 2018-11-15
Update packages view to only show distinct rows

chefsalim:
This fixes an issue where we are showing multiple versions of the same package in the UI. The root cause is duplicate rows in the packages view, which is resolved by updating the view defn.  A new test is added to catch this scenario in the future.

tenor-244413307


thesentinels:

Thank you for contributing!

PR #806, opened 2018-11-16
More perf/cache related fixes

chefsalim:
This PR has a few more perf and cache related fixes.
1. This backs out the last change to make the Postgres package view return distinct rows as the performance penalty was too high - we now de-dup the list in code after it is returned
2. It adds caching to non-channel package get calls
3. It adds checking for private packages for cache calls
4. It clears the cache when package visibility is modified

tenor-226582615


thesentinels:

Thank you for contributing!

PR #807, opened 2018-11-16
More perf tuning

chefsalim:
With this PR we go back to using the regular package table instead of the combined package-channels-platform view for the latest package calls due to the perf penalty on the view.  This change also adds a new metric to start tracking DB call times.

tenor-56030929

### DIFF
--- a/components/builder-db/src/models/channel.rs
+++ b/components/builder-db/src/models/channel.rs
@@ -1,5 +1,7 @@
 use super::db_id_format;
 use chrono::NaiveDateTime;
+use time::PreciseTime;
+
 use diesel;
 use diesel::dsl::sql;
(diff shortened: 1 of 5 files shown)


components/builder-db/src/models/channel.rs
```diff
@@ -98,23 +101,31 @@ impl Channel {
...
             )).limit(1)
-            .get_result(conn)
+            .get_result(conn);
+
+        let end_time = PreciseTime::now();
+        trace!(
+            "DBCall channel::get_latest_package time: {} ms",
+            start_time.to(end_time).num_milliseconds()
+        );
+        Histogram::DbCallTime.set(start_time.to(end_time).num_milliseconds() as f64);
```
raskchanky: These metrics are terrific.

PR #867, opened 2019-01-04
Use the IP address of API service, not my sys.ip in backend block

jamesc:
This allows the builder api and api-proxy services to be different hosts (or containers/pods)

### DIFF
--- a/components/builder-api-proxy/habitat/config/nginx.conf
+++ b/components/builder-api-proxy/habitat/config/nginx.conf
@@ -86,9 +86,13 @@ http {
                    '"$http_referer" "$http_user_agent" - $http_x_forwarded_for';
 
   upstream backend {
+    {{~#if cfg.load_balanced}}
     {{~#each bind.http.members as |member|}}
     server {{member.sys.ip}}:{{member.cfg.port}};
     {{~/each}}
+    {{~else}}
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/builder-api-proxy/habitat/config/nginx.conf
```diff
@@ -86,7 +86,9 @@ http {
                    '"$http_referer" "$http_user_agent" - $http_x_forwarded_for';
 
   upstream backend {
-    server {{sys.ip}}:{{bind.http.first.cfg.port}};
+    {{~#each bind.http.members as |member|}}
```
chefsalim: Unfortunately this breaks our hosted service where we run them both together... we'll have to do a larger migration to a model where we run a single nginx reverse proxy running builder-api-proxy, and multiple backends running builder-api

jamesc: I can't get the two services running inside the same pod with our operator.

How about if I add a `backend-mode` option which defaults to the current logic (use `sys.ip`) but allows for a `load-balanced` mode which does this new logic.

chefsalim: That seems reasonable - go for it

PR #869, opened 2019-01-04
Allow setting of datastore superuser password via config

jamesc:
In a container based deployment scenario you need to set the password
externally so you can pass it to all containers that need it. This
enables that while maintaining the existing behaviour as the default.


thesentinels:

Thank you for contributing!

PR #881, opened 2019-01-18
Replace Strings and &strs with type-safe habitat_core::ChannelIdent type

baumanj:
See https://github.com/habitat-sh/core/pull/107

### DIFF
--- a/components/builder-api/src/server/resources/jobs.rs
+++ b/components/builder-api/src/server/resources/jobs.rs
@@ -24,8 +24,8 @@ use crate::protocol::jobsrv;
 use crate::protocol::net::NetOk;
 use crate::protocol::originsrv::OriginPackageIdent;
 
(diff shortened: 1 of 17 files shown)


components/builder-api/src/server/resources/jobs.rs
```diff
@@ -244,10 +246,10 @@ fn do_group_promotion_or_demotion(
     let channel = match Channel::get(origin, channel, &*conn) {
         Ok(channel) => channel,
         Err(NotFound) => {
-            if channel != STABLE_CHANNEL || channel != UNSTABLE_CHANNEL {
+            if channel != &ChannelIdent::stable() || channel != &ChannelIdent::unstable() {
```
chefsalim: Merge conflict here with the fix to change to "&&" that has gone in..

baumanj: Oops, I fixed it earlier, but hadn't force-pushed the rebase. Done now!

PR #885, opened 2019-01-24
Make builder-minio clustered

jsirex:
This change completely reworks minio:
* Fix ssl certs not always copied on change
* Fix ssl cert name private.crt -> public.crt
* Add standalone mode (defualt)
* Add automatic cluster mode (using habitat discovery)
* Add explicit cluster mode
* Smart reload/restart service on config changes

Added docker-compose.yml for test.

Signed-off-by: Yauhen Artsiukhou <jsirex@gmail.com>

### DIFF
--- a/components/builder-minio/habitat/hooks/post-run
+++ b/components/builder-minio/habitat/hooks/post-run
@@ -1,19 +1,15 @@
-#!/bin/sh
+#!/bin/bash
+# shellcheck disable=2154
 
 exec 2>&1
 
-export AWS_ACCESS_KEY_ID="{{cfg.key_id}}"
-export AWS_SECRET_ACCESS_KEY="{{cfg.secret_key}}"
+source "{{ pkg.svc_config_path }}/hook-helper.sh"
 
-{{ #if cfg.use_ssl }}
-MINIO_ENDPOINT="--endpoint-url https://localhost:9000 --no-verify-ssl"
-{{else ~}}
-MINIO_ENDPOINT="--endpoint-url http://localhost:9000"
-{{ /if }}
+minio_splay_delay
 
-if aws $MINIO_ENDPOINT s3api list-buckets | grep "{{cfg.bucket_name}}" > /dev/null; then
-  echo "Minio already configured"
-else
-  echo "Creating bucket in Minio"
-  aws  $MINIO_ENDPOINT s3api create-bucket --bucket "{{cfg.bucket_name}}"
+if [ "$standalone" != "true" ] && [ ! -f "$run_good_hook" ]; then
+    echo "Minio is not ready for bucket configuring. May be later"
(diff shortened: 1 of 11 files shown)


components/builder-minio/habitat/hooks/post-run
```diff
@@ -1,19 +1,15 @@
...
-export AWS_SECRET_ACCESS_KEY="{{cfg.secret_key}}"
+source {{ pkg.svc_config_path }}/hook-helper.sh
 
-{{ #if cfg.use_ssl }}
-MINIO_ENDPOINT="--endpoint-url https://localhost:9000 --no-verify-ssl"
-{{else ~}}
-MINIO_ENDPOINT="--endpoint-url http://localhost:9000"
-{{ /if }}
+# Give service some time to start
+sleep $((RANDOM % 20 + 10))
```
chefsalim: Do we need the random sleep for standalone mode?  We're going to hit this in a lot of cases.

components/builder-minio/habitat/hooks/reconfigure
```diff
@@ -0,0 +1,14 @@
+#!{{pkgPathFor "core/bash"}}/bin/bash
+
+exec 2>&1
+
+source {{ pkg.svc_config_path }}/hook-helper.sh
+
+# Wait 10..30 seconds for initialization
+sleep $((RANDOM % 20 + 10))
```
chefsalim: Same q on random sleep for non-clustered modes..

chefsalim:
Thanks for the contribution @jsirex !  Looks good overall, a couple of questions added.

components/builder-minio/habitat/hooks/post-run
jsirex: I agree with you that this part is bad. I'm also don't like random waits. I did it because of habitat bug: post-run hook will not re-run on failure (https://github.com/habitat-sh/habitat/issues/2364). It is hard to guess how much time minio will spend at first start so just wait. I added delay. Because there are might be number of nodes (in cluster) I added splay (random).
If bug with post-run will be resolved I believe this code must be deleted, but better to leave a tiny splay for cluster mode.

Also good point for standalone mode. I'll better change delay to 2 sec for standalone mode - more than enough. Will send update soon

chefsalim: Cool, that works - thanks!

chefsalim:
Oh, shellcheck failures also need fixing..

jsirex:
I see in test/shellcheck.sh files were disabled. For now I see only way to exclude "bad" files from shellcheck. Is it ok?

chefsalim:
May be better to refactor some of the code, eg, see https://github.com/habitat-sh/builder/pull/874

PR #889, opened 2019-01-25
Fix to account for different build paths

chefsalim:
Fix a missed case where the studio build results are in different locations on Windows vs. Linux builds

tenor-172656898


thesentinels:

Thank you for contributing!

PR #898, opened 2019-01-30
Adds initial backend functionality for origin deletion

eeyun:


thesentinels:

Thank you for contributing!

components/builder-api/src/server/authorize.rs
```diff
@@ -103,3 +104,16 @@ pub fn check_origin_member(
...
     }
 }
+pub fn check_origin_empty(
+    req: &HttpRequest<AppState>,
+    origin: &str,
+) -> Result<bool> {
+   let conn = req.state().db.get_conn().map_err(Error::DbError)?;
+   let count = Package::count_origin_packages(&origin, &*conn).map_err(Error::DieselError)?;
+
+    if count > 0 {
```
chefsalim: This can be simplified to just Ok(count == 0)

components/builder-db/src/models/origin.rs
```diff
@@ -119,6 +122,42 @@ impl Origin {
...
+            diesel::delete(
+                origin_secret_keys::table
+                    .filter(origin_secret_keys::origin.eq(origin)))
+                    .execute(conn)?;
+            diesel::delete(
+                origin_public_keys::table
+                    .filter(origin_public_keys::origin.eq(origin)))
+                    .execute(conn)?;
+            diesel::delete(
+                origin_members::table
```
chefsalim: I think we may want to also make sure that there are also no members other than the owner in the origin before we trigger the delete

chefsalim:
Looks great overall, just a couple of comments..

components/builder-db/src/models/origin.rs
eeyun: I can get this in tomorrow morning. Is this something we definitely want? This deletion will remove any membership associated with the origin being terminated.

I don't have a preference either way mostly curious your thoughts on it.

chefsalim: I'm mostly concerned about any undefined behavior if there are other members in the origin. If things look clean after the deletion then go for it. We can also add some warning message(s) in the cli/UI later.

eeyun: Yep, I validated this locally, there may be other undiscovered behavior here but WRT members, everything appears clean in the UI for the owner as well as the subordinate members.

Issue #819, opened 2018-11-30
The 'Cancel jobs' feature should be more discoverable in the UI


chefsalim:
We added a very useful feature to allow users to cancel their group jobs.

Unfortunately the UI for that is buried inside of the group page, and even on that page is not very visible.

We should look at making this more discoverable by adding a Cancel button to the higher level pages.

apriofrost:
Hi @chefsalim , I just tried this task flow and realized that the 'view output' link is incorrectly linked to the job details page under the package level, it should be the group details page under the Origin level, which contains the Cancel button (it is not shown immediately after the build request is submitted, however).  I'm going to work with @mgamini to fix this. Thanks for reporting the issue!

chefsalim:
Great, thanks @apriofrost .. also would be great if the Cancel button was a bit more prominent looking, like the Build button.

apriofrost:
@chefsalim The Cancel button was made looking less prominent because it was designed to sit next to the Promote to Stable button, which is the primary action on that page. It seems the promote functionality is not ready to be implemented yet. I'll work with @mgamini to see the current status of the Promote button and will adjust the styles for the Cancel button accordingly.

apriofrost:
To-do after discussion with @mgamini 

- [ ] Change the link of the 'view the output' to the group details page under Build Jobs tab in the Origin level (the page that contains the 'Cancel unfinished jobs' button)
![image](https://user-images.githubusercontent.com/9401459/49461631-f44c2f00-f7a8-11e8-886a-7b1ab8fb66cd.png)
- [ ] Investigate the status of 'Promote to stable' button, if it's not planned to implementation shortly, change the 'Cancel unfinished jobs' button to be the primary button on the page.
- [ ] Investigate and fix the issue when the 'Cancel unfinished jobs' button is not showing up on the page constantly between the time when the job has been initiated and is not completed/canceled, e.g., when the job is in the queue.

apriofrost:
@Shadae is taking a look at this to fix the first to-do item (fixing the link)

PR #907, opened 2019-02-01
Fix cancel job groups button to show up when jobs are in progress and pending

Shadae:
When fixing the link part of #819 and running build after build. I never saw the cancel button surface. Following the logic, I think we want the cancel button to show up when build jobs are in progress. Currently, the cancel button is set to surface when jobs are `Pending`. We could also make it surface for both `InProgress` and `Pending`, happy for some direction as the page also currently says `Cancel Remaining Jobs` and the mocks say `Cancel Unfinished Jobs`.

Signed-off-by: shadae <sholmes@chef.io>


thesentinels:

Thank you for contributing!

chefsalim:
Jobs should be cancellable at any stage (except Complete or Failed..), so it would be fine to expand the state check.

Shadae:
Job groups can now be canceled when jobs are `inprogress`, or `notstarted`. 
screen shot 2019-02-05 at 10 07 09 am

PR #910, opened 2019-02-04
Fix BoxKeyPair::encrypt return type

baumanj:
https://github.com/habitat-sh/core/pull/115 has merged, so this can merge once CI passes

### DIFF
--- a/components/builder-api/src/server/resources/origins.rs
+++ b/components/builder-api/src/server/resources/origins.rs
@@ -575,7 +575,7 @@ fn create_origin_secret(
     }
 
     // get metadata from secret payload
-    let secret_metadata = match BoxKeyPair::secret_metadata(body.value.as_bytes()) {
+    let secret_metadata = match BoxKeyPair::secret_metadata(&body.value) {
         Ok(res) => res,
         Err(err) => {
             debug!("{}", err);
--- a/components/builder-api/src/server/resources/origins.rs
+++ b/components/builder-api/src/server/resources/origins.rs
@@ -32,6 +32,7 @@ use futures::future::Future;
 use serde_json;
 
 use crate::bldr_core;
+use crate::hab_core::crypto::keys::box_key_pair::WrappedSealedBox;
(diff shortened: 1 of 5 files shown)


thesentinels:

Thank you for contributing!

components/builder-api/src/server/resources/origins.rs
```diff
@@ -575,8 +576,12 @@ fn create_origin_secret(
     }
 
     // get metadata from secret payload
-    let secret_metadata = match BoxKeyPair::secret_metadata(body.value.as_bytes()) {
-        Ok(res) => res,
+    let ciphertext = WrappedSealedBox::from(body.value.as_str());
+    let secret_metadata = match BoxKeyPair::secret_metadata(&ciphertext) {
+        Ok(res) => {
+            debug!("Secret Metadata: {:?}", res);
```
raskchanky: Is it ok to leave this `debug!` in?  Just wondering about sensitive info leaking into the logs.

baumanj: I think so, because if you look at [`secret_metadata`](https://github.com/habitat-sh/core/blob/master/components/core/src/crypto/keys/box_key_pair.rs#L332), you'll see it's returning is a `BoxSecret`:

```
pub struct BoxSecret<'a> {
    pub sender: &'a str,
    pub ciphertext: Vec<u8>,
    pub receiver: Option<&'a str>,
    pub nonce: Option<Nonce>,
}
```

The key here is that the key field is *cipher*text, not plaintext. In a functional crypto system, cipher text should be indistinguishable from random bytes. The only way to recover the plaintext would be to have the appropriate keys, which this struct lacks. It does contain the `nonce`, which I think would possibly be better to *not* share, but in the context of internal debug logs, I think its value in debugging outweighs the risk in exposing it. Then again, I'm not a security expert. In any case, this is not a net change, so perhaps we should file a separate issue to consider a non-derived implementation of `Debug` for `BoxSecret` which would expose fewer details.

See also https://en.wikipedia.org/wiki/Ciphertext-only_attack

PR #914, opened 2019-02-06
adds firewall rule to windows worker hardening docker network

mwrock:
This drops all TCP traffic from the docker NAT network to the local 10.* ethernet network as well as the NAT default gateway which resolves to the local host. It also blocks all non DNS UDP traffic.

Ideally I would have one rule that blocks everything between those endpoints and than an allow rule for UDP 53 between the docker NAT and its DNS servers but thats not possible with how the windows firewall works. It does not process rules in order but applies the deny rules first and if it finds a match it does not override any allow rules.


thesentinels:

Thank you for contributing!

mwrock:
I've made some minor changes and confirmed via terraform in acceptance that this creates the correct rule and that I cannot access the 10.* network or the nat gateway (which is the host) inside of a docker studio on the worker. I have also confirmed that normal external internet activity works.

PR #916, opened 2019-02-06
Check PackageTarget before attempting docker export

### LINKED ISSUE #913: Disable docker exporter when building x86_64-linux-kernel2 packages
When building packages for the x86_64-linux-kernel2 package target, we should disable the Docker exporter.  

Docker states the minimum kernel version for the docker engine as 3.10 ([source](https://docs.docker.com/install/linux/docker-ce/binaries/)).  Since x86_64-linux supports kernels 3.2 and up, any package intended to run in a Docker container will be able to do so with the x86_64-linux package target.  

Not disabling the exporter for x86_64-linux-kernel2 won't be harmful, but will generate excess work that will likely never be used.

smacfarlane:
This updates builder-worker to check the PackageTarget for supported platforms (x86_64-linux, x86_64-windows) before attempting a docker export of a package, if exports are enabled.  It also introduces a plan variant for x86_64-linux-kernel2 with the docker dependencies removed. 

#### Question
The `use` statements for the docker export code are still present for the x86_64-linux-kernel2 builds, which contain a `lazy_static!` to determine the path to the `docker` command and `hab-pkg-export-docker` command. If I understand that correctly, that will be evaluated at run-time, which introduces a potential point of failure at run-time as those packages won't be present.  Looking at existing code, it should never be executed, however I'd feel safer if there was a way to not import that module based on the active package target.

Is there a way to do that? Is there a better way to ensure runtime safety in this case? 

closes #913 

### DIFF
--- a/components/builder-worker/src/runner/mod.rs
+++ b/components/builder-worker/src/runner/mod.rs
@@ -40,6 +40,7 @@ use crate::bldr_core::socket::DEFAULT_CONTEXT;
 #[cfg(not(windows))]
 use crate::hab_core::os::users;
 use crate::hab_core::package::archive::PackageArchive;
+use crate::hab_core::package::target;
 #[cfg(not(windows))]
 use crate::hab_core::util::posix_perm;
 
@@ -331,10 +332,15 @@ impl Runner {
         Ok(archive)
     }
 
-    fn do_export(&mut self, tx: &mpsc::Sender<Job>, mut streamer: &mut JobStreamer) -> Result<()> {
+    fn do_export(
+        &mut self,
+        tx: &mpsc::Sender<Job>,
+        mut archive: &mut PackageArchive,
+        mut streamer: &mut JobStreamer,
+    ) -> Result<()> {
         self.check_cancel(tx)?;
 
-        match self.export(&mut streamer) {
+        match self.export(&mut archive, &mut streamer) {
             Ok(_) => (),
             Err(err) => {
                 self.fail(net::err(ErrCode::EXPORT, "wk:run:export"));
@@ -401,8 +407,8 @@ impl Runner {
         self.do_install_key(&tx, &mut streamer)?;
         self.do_clone(&tx, &mut streamer)?;
 
-        let archive = self.do_build(&tx, &mut streamer)?;
-        self.do_export(&tx, &mut streamer)?;
+        let mut archive = self.do_build(&tx, &mut streamer)?;
+        self.do_export(&tx, &mut archive, &mut streamer)?;
         self.do_postprocess(&tx, archive, &mut streamer)?;
 
         self.cleanup();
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/builder-worker/src/runner/mod.rs
```diff
@@ -331,10 +332,15 @@ impl Runner {
         Ok(archive)
     }
 
-    fn do_export(&mut self, tx: &mpsc::Sender<Job>, mut streamer: &mut JobStreamer) -> Result<()> {
+    fn do_export(
+        &mut self,
+        tx: &mpsc::Sender<Job>,
+        mut archive: &mut PackageArchive,
```
eeyun: Ok so, bear with me here because I'm NOT fully versed in the build worker code and thus we'll definitely want some @chefsalim feedback here as well but I also had some questions.

It looks like you're hauling this archive through almost explicitly for the sake of grabbing the package target for branching logic in the export function. If that's the case I don't know that it's actually totally necessary.

I _think_ you can probably get the `PackageTarget` from the archive generated by `do_build` without passing in the extra argument as do_build is just returning `Result<PackageArchive>`

What i think that means is changing the function signature in `do_export` back to the way it was and then at line 410 you've actually already got the archive you need which you should be able to get the target from with a:
`let target = archive.target()?;` and then maybe you could change the `export` function sig to something like:
```
fn export(&mut self, archive: PackageTarget, streamer: &mut JobStreamer) -> Result<()>
```
Then in your export function you can get rid of line 515.

I think that might make this a little more lightweight. If you've tried all this/i'm missing something feel free to tell me i'm an idiot :joy:

chefsalim: Right, yeah I think we can just make this change in the export function, without making any signature changes at all.  You should be able to check `self.workspace.job.target` and skip doing the export if it is Linux2.  It may be good to also emit a debug! message when skipping due to target.

smacfarlane: My first pass was similar to that ( I had the logic wrapped around do_export, rather than pass the archive in),  but decided to go this route because the [following line](https://github.com/habitat-sh/builder/blob/master/components/builder-worker/src/runner/mod.rs#L509-L510 ) could also use `archive.is_a_service()` if we address fnichols comment about the `is_a_service` method on PackageArchive and I _think_ that is probably more correct than "last built".  I'm probably overthinking things though since we don't need to address that as part of this change.   

I think you're right right though, and should make the change need today and not the one we might need at some point in the future.

chefsalim: Yeah the is_a_service thing is a bigger discussion...

components/builder-worker/src/runner/mod.rs
```diff
@@ -506,26 +508,32 @@ impl Runner {
...
+                            &self.config.bldr_url,
+                            &self.bldr_token,
+                        )
+                        .export(streamer)?;
+
+                        if !status.success() {
+                            return Err(Error::ExportFailure(status.code().unwrap_or(-1)));
+                        }
+
+                        section.end()?;
```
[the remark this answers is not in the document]
smacfarlane: I think I'm misunderstanding something.  It looks like section begins only if `is_a_service` is true.  Should that be moved outside that block into the parent (around line 511)?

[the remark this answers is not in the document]
chefsalim: Ah never mind... I take back my comment :)

PR #951, opened 2019-02-28
Changes search results to be sorted in descending order

eeyun:
I'm not really certain how to test this outside the context of a fully populated builder instance like the one in acceptance but it looks to function as intended.

Effectively - a community member noticed recently that because of the combination of pagination of our search results and the fact that we dont order those results from newest packages to oldest, `hab pkg search` was pretty useless for taking a look at what was available in builder. I know we likely have a more serious overhaul coming for our package search mechanisms but I do think returning results ordered by ident from newest to oldest probably returns results in a more expected order. 


thesentinels:

Thank you for contributing!

chefsalim:
I'm not sure this is the right solution to move forward with. Is the ordering actually deterministic if we order by ident ascending and created at descending?  The test result looks like a mix which is not what we want.  Instead, I'd propose 1) we add the pagination to hab client, and 2) we add an optional ordering clause, to let users specify ascending or descending sort on the returned idents.  That will be simpler and more consistent in terms of the results we are returning, and will be a better long term solution.

chefsalim:
Also it would be good to make these changes in context of the overall search epic, to make sure we are considering all the aspects of the improvements we want to make there.

PR #958, opened 2019-03-04
Add build_targets to bldr toml

chefsalim:
This change adds support for build targets in the bldr.toml file.  If no build_targets are specified explicitly, we default to X86_64_LINUX and X86_64_WINDOWS.  Therefore, in order to build Linux Kernel2 packages via github push, that build target will need to be explicitly specified in a bldr.toml. This decision is because we expect most plans will NOT need to automatically build Kernel2 packages.

tenor-90272445

### DIFF
--- a/components/builder-core/src/build_config.rs
+++ b/components/builder-core/src/build_config.rs
@@ -12,8 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::fmt;
+use std::iter::FromIterator;
 use std::ops::Deref;
 use std::path::PathBuf;
 use std::result;
@@ -26,6 +27,7 @@ use toml;
 
 use crate::error::{Error, Result};
 use crate::hab_core::config::ConfigFile;
+use crate::hab_core::package::target::{self, PackageTarget};
 
 /// Postprocessing config file name
 pub const BLDR_CFG: &str = ".bldr.toml";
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

components/builder-core/src/build_config.rs
```diff
@@ -156,6 +158,9 @@ pub struct ProjectCfg {
     /// notification to determine if an automatic rebuild should occur.
     #[serde(default)]
     pub paths: Vec<Pattern>,
+    /// Package targets to build when changes detected
+    #[serde(default = "ProjectCfg::default_build_targets")]
+    pub build_targets: HashSet<PackageTarget>,
```
smacfarlane: Do we need some error handling around the reading of `build_targets`?  I would expect that typos and experimentation would be common with this setting.  Changing the tests below to an invalid PackageTarget  results in the following error: 

```
---- build_config::test::triggered_by stdout ----
thread 'build_config::test::triggered_by' panicked at 'called `Result::unwrap()` on an `Err` value: DecryptError("string cannot be parsed: \"i-am-not-a-package-target\" (Package targets must be in architecture-platform format (example: x86_64-linux)) for key `builder-api.build_targets`")', src/libcore/result.rs:1009:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

Would that cause builder-api to crash if a user connected a repo with a malformed build_target? or would it fail to load the the .bldr.toml file for that project resulting in no builds for the user.  

Is there a way we could provide feedback to the user to let them know?

chefsalim: That is a really great question. Currently, if we fail to correctly parse the bldr.toml, we just fall back to default values (ie, pretend that the bldr.toml file doesnt exist).  There is no panic, but we do drop the error on the floor (other than logging a debug output).  Instead of continuing on this way, we could fail the push request entirely with an error code and some text. Unfortunately that still isn't really easily visible to the user - you have to go to the specific github app settings to find the result of the push, and that's not terribly intuitive either.

Issue #942, opened 2019-02-20
Configure rustfmt


raskchanky:
Do the same configuration for this repo that we did [for habitat](https://github.com/habitat-sh/habitat/pull/6134).

PR #962, opened 2019-03-05
Configure rustfmt

raskchanky:
I copied the scripts that are in use in the habitat repo over here and
adjusted various CI settings to use them. The duplication feels bad.

Closes https://github.com/habitat-sh/builder/issues/942


thesentinels:

Thank you for contributing!

support/ci/shared.sh
```diff
@@ -0,0 +1,34 @@
...
+  # This means that if we're always using the latest version, then we're going to have enormous
+  # churn. Even PRs that don't touch rust code will likely fail CI, since master will have been
+  # formatted with a different version than is running in CI. Because of this, we're going to pin
+  # the version of nightly that's used to run rustfmt and bump it when we do a new release.
+  #
+  # Note that not every nightly version of rust includes rustfmt. Sometimes changes are made that
+  # break the way rustfmt uses rustc. Therefore, before updating the pin below, double check
+  # that the nightly version you're going to update it to includes rustfmt. You can do that
+  # using https://mexus.github.io/rustup-components-history/x86_64-unknown-linux-gnu.html
+  echo "nightly-2019-03-04"
```
chefsalim: A couple of questions - can you confirm:
1. I assume we will need to also make sure dev environments have this specific nightly toolchain installed?
2. VSCode/Atom/RLS/etc plug-ins can pick up the rustfmt options and format-on-save automatically as today?
3. Builds will continue to use stable, not nightly toolchain?

raskchanky: 1. Correct
1. Assuming your editor in question allows you to customize the command that's used to run `rustfmt`, correct. I can confirm this is the case for vim and emacs. I have no experience with the others. For example, for vim, I added this line to my `~/.vimrc`: `let g:rustfmt_command = 'rustup run nightly-2019-03-04 rustfmt'`
1. Correct

chefsalim: Thanks... one more q: there are no actual formatting changes in this PR based on the new settings?

raskchanky: Yeah, I was waiting to see if I hooked up CI right before adding those changes.  I think it's good though.  I'll add them now.

support/ci/shared.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+set -euo pipefail
+
+get_current_toolchain() {
+  # It turns out that every nightly version of rustfmt has slight tweaks from the previous version.
+  # This means that if we're always using the latest version, then we're going to have enormous
+  # churn. Even PRs that don't touch rust code will likely fail CI, since master will have been
+  # formatted with a different version than is running in CI. Because of this, we're going to pin
+  # the version of nightly that's used to run rustfmt and bump it when we do a new release.
```
chefsalim: Maybe rehashing previous conversations - this looks like one more hoop to jump through every hab release, and create extra work downstream every time we release/re-pin nightly. Are we sure we are getting sufficient value from nightly rustfmt to bother?

raskchanky: Yes

raskchanky: The actual work involved is pretty small and the value (imo) is pretty great.

baumanj: As more of the configuration options in `rustfmt` stabilize, it may be worth revisiting this decision, but right now there are so many useful options that I think it's worth the extra effort to use nightly.

.travis.yml
```diff
@@ -152,12 +150,14 @@ matrix:
         - ./support/ci/compile_libarchive.sh
         - ./support/ci/compile_zmq.sh
         - source ./support/ci/rust_env.sh
+        - cargo build --package habitat-builder-protocol
```
christophermaier: A comment on why we need this would be useful.

raskchanky: Added

PR #965, opened 2019-03-07
Fix panic when bad target specified

chefsalim:
This fixes a case where specifying a bad target for package download would cause the builder API to panic. Also adds a test to catch this case.


thesentinels:

Thank you for contributing!

PR #972, opened 2019-03-12
Fix force upload by handling error on DB conflict

chefsalim:
Currently force uploading a package can fail if the metadata already exists but the artifact does not. This fixes the case by accounting for the metadata conflict (on conflict do nothing) case.


thesentinels:

Thank you for contributing!

PR #974, opened 2019-03-13
Set up panic hook for builder worker

chefsalim:
This sets up a custom panic hook for builder worker.  In the (rare) case that the builder worker panics on a thread, it will print out the backtrace and panic info, and then exit the process. The process will be restarted by the supervisor. We prefer this behavior to the default, where a single thread (eg, runner) can die, and leave the heartbeat thread running so it appears that the runner is in a working state when it has actually died.

tenor-129272970


thesentinels:

Thank you for contributing!

PR #976, opened 2019-03-13
Remove airlock usage in builder-worker

chefsalim:
Now that we have consolidated all the build workers to Docker Studio, we can remove the integration code for Airlock. This change removes the configuration and associated Airlock execution code from the worker.  The terraform scripts are updated to no longer create a special user and group for studio builds as well.

tenor-203575642


thesentinels:

Thank you for contributing!

PR #978, opened 2019-03-13
Remove Airlock component from builder

chefsalim:
This is the last part of the Airlock transition - this removes the Airlock component from builder repo, and also removes it as a dep for builder-worker, along with some associated tweaks.

tenor-170761497


thesentinels:

Thank you for contributing!

.travis.yml
```diff
@@ -12,7 +12,7 @@ env:
   global:
     - PATH=$HOME/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin:$HOME/.cargo/bin:$PATH
     # Habitat Rust program components
-    - _RUST_HAB_BIN_COMPONENTS="components/airlock"
+    - _RUST_HAB_BIN_COMPONENTS=""
```
raskchanky: Is this worth keeping at all if it's empty?

Makefile
```diff
@@ -1,5 +1,5 @@
 UNAME_S := $(shell uname -s)
-BIN = airlock
+BIN =
```
raskchanky: Same question here as above.  Why not just delete the whole thing?

.travis.yml
chefsalim: Yeah I started to remove all the bin plumbing  but we may have a new bin tool in the (near to mid term) future so leaving it in for now..

PR #988, opened 2019-03-27
Add package deletion capability

chefsalim:
This PR adds a limited package deletion capability. 

Previously, any uploaded package could never be deleted, since other packages could be depending on that package, and deletion could break the dependency chain for packages installed in the wild. 

However, there are also legit cases where a package can get uploaded and need to be deleted - for example, packages with bad naming or versioning etc that are not yet in a dep chain. 

We can safely allow package deletion with the following two provisions:
1. The package being deleted must not be in the stable channel
2. The package being deleted must not have any reverse dependencies for its project

If those two conditions are met, the deletion will succeed.

tenor-185379840


thesentinels:

Thank you for contributing!

PR #996, opened 2019-03-29
Migrate verification tests to buildkite

scotthain:
Signed-off-by: Scott Hain <shain@chef.io>

This does not include the integration tests, which should be added in a followup PR.

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -11,6 +11,7 @@ steps:
   - label: "[lint] :linux: :bash: Shellcheck"
     command:
       - ./test/shellcheck.sh
+    skip: Need to update this codebase to shellcheck 0.5.0
     timeout_in_minutes: 10
     expeditor:
       executor:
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -97,85 +97,11 @@ unit-$1: linux ## executes the $1 component's unit test suite
...
-               clippy::wrong_self_convention \
-               renamed_and_removed_lints
-
-lint:
-	$(run) cargo clippy --all-targets --tests $(CARGO_FLAGS) -- \
-					$(addprefix -A ,$(UNEXAMINED_LINTS)) \
-					$(addprefix -A ,$(ALLOWED_LINTS)) \
-					$(addprefix -W ,$(LINTS_TO_FIX)) \
-					$(addprefix -D ,$(DENIED_LINTS))
+lint: ## executes the $1 component's linter checks
```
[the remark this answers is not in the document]
scotthain: Yep - good call, this was a copy/paste error.

baumanj:
> The big caveat here is the Shellcheck step is disabled - there are a number of changes required and I don't feel comfortable with this codebase to tackle updating them without knowing the intent of the code itself.

Why not keep it generally enabled, but add targeted suppressions for the changes you're not comfortable with?

scotthain:
@baumanj so the good news is someone already did! I rebased and all the shellcheck issues were resolved :D

scotthain:
I believe the fix for the clippy issues will be in https://github.com/habitat-sh/builder/pull/1027 - waiting on that PR to rebase.

.expeditor/verify.pipeline.yml
```diff
@@ -18,17 +18,177 @@ steps:
...
+    command:
+      - ./test/run_cargo_test.sh builder-api
+    retry:
+      automatic:
+        limit: 1
+    expeditor:
+      executor:
+        docker:
+
+  - label: "[unit] :linux: builder-api-proxy"
```
chefsalim: we don't have any unit tests in builder-api-proxy

.expeditor/verify.pipeline.yml
```diff
@@ -18,17 +18,177 @@ steps:
...
+    command:
+      - ./test/run_cargo_test.sh builder-core
+    retry:
+      automatic:
+        limit: 1
+    expeditor:
+      executor:
+        docker:
+
+  - label: "[unit] :linux: builder-datastore"
```
chefsalim: no unit tests in builder-datastore

.expeditor/verify.pipeline.yml
```diff
@@ -18,17 +18,177 @@ steps:
...
+    command:
+      - ./test/run_cargo_test.sh builder-jobsrv
+    retry:
+      automatic:
+        limit: 1
+    expeditor:
+      executor:
+        docker:
+
+  - label: "[unit] :linux: builder-memcached"
```
chefsalim: no cargo tests in builder-memcached

.expeditor/verify.pipeline.yml
```diff
@@ -18,17 +18,177 @@ steps:
...
+    command:
+      - ./test/run_cargo_test.sh builder-memcached
+    retry:
+      automatic:
+        limit: 1
+    expeditor:
+      executor:
+        docker:
+
+  - label: "[unit] :linux: builder-minio"
```
chefsalim: no cargo tests in builder-minio

.expeditor/verify.pipeline.yml
```diff
@@ -18,17 +18,177 @@ steps:
...
+    command:
+      - ./test/run_cargo_test.sh builder-protocol
+    retry:
+      automatic:
+        limit: 1
+    expeditor:
+      executor:
+        docker:
+
+  - label: "[unit] :linux: builder-web"
```
chefsalim: no cargo tests in builder-web

.expeditor/verify.pipeline.yml
```diff
@@ -18,17 +18,177 @@ steps:
 
   - label: "[lint] :linux: :bash: rustfmt"
     command:
+      - ./test/run_cargo_test.sh builder-jobsrv # this is overkill, we just need the build
```
chefsalim: Do we need to build all of jobsrv, or can we just build builder-protocol?

scotthain: Unfortunately the whole thing. I tried `builder-protocol` first and it didn't build out the things in the correct locations.

scotthain:
@chefsalim I would argue we should exercise the the cargo test functionality for the things that don't have unit tests simply to ensure the code still builds correctly and whatnot. Would you prefer we remove these? Or rename the steps? Or we could add a section that runs `cargo build` instead of test. Also what is the likelihood of us ever adding unit tests to these? I'm fine with whatever we end up deciding :)

chefsalim:
@scotthain I'd prefer to remove the cargo tests where we don't need them... there's no need to queue up long running jobs if we don't need to ,etc.

PR #999, opened 2019-04-01
Ignore generated files

raskchanky:
I'm aware that `components/builder-protocol/src/message/mod.rs` is _not_ a generated file, but typically this file is problematic as well, and it's quite small, so ignoring it should make the silly CI errors stop.


raskchanky:
This isn't actually helping us 😭

raskchanky:
I'm going to close this because it doesn't actually solve the problem that needs solving.

PR #1001, opened 2019-04-01
Initial pass of build dep integration into graph

eeyun:
Woo wee, this is a fun one. I would love all of the feedback on this PR as I haven't done a refinement pass on this code yet. 

This adds new package metadata to the database as well as integrates build deps into the build graph behind a feature flag. There will be another PR/Migration effort that follows on this change.


components/builder-core/src/package_graph.rs
```diff
@@ -88,19 +88,23 @@ impl PackageGraph {
         }
     }
 
-    pub fn build<T>(&mut self, packages: T) -> (usize, usize)
+    pub fn build<T>(&mut self, packages: T, feat: Option<()>) -> (usize, usize)
```
chefsalim: "feat" could just be a bool instead of Option<()>.. eg have something like "use_build_deps : bool"

components/builder-jobsrv/src/server/handlers.rs
```diff
@@ -635,7 +636,14 @@ pub fn job_graph_package_create(req: &RpcMessage, state: &AppState) -> Result<Rp
         }
     };
     let start_time = PreciseTime::now();
-    let (ncount, ecount) = graph.extend(&package);
```
chefsalim: could just pass in the feat::is_enabled(feat::BuildDeps) directly to extend..

terraform/scripts/create_bootstrap_bundle.sh
```diff
@@ -1,186 +0,0 @@
-#!/bin/bash
```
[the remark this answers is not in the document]
eeyun: no!

chefsalim:
Also looks like cargo test and clippy need fixing.

eeyun:
ok this push should resolve the clippy fixes though testing this locally is a tough due to the mistmatch on openssl.

PR #1002, opened 2019-04-03
Add target params to pkg related APIs

chefsalim:
This change adds target query param handling to some package related APIs that were missing it - package promote, demote and channel list.

tenor-172904196

### DIFF
--- a/components/builder-api/src/server/resources/channels.rs
+++ b/components/builder-api/src/server/resources/channels.rs
@@ -193,7 +193,7 @@ fn delete_channel(req: HttpRequest<AppState>) -> HttpResponse {
 }
 
 #[allow(clippy::needless_pass_by_value)]
-fn promote_package(req: HttpRequest<AppState>) -> HttpResponse {
+fn promote_package((qtarget, req): (Query<Target>, HttpRequest<AppState>)) -> HttpResponse {
     let (origin, channel, pkg, version, release) =
         Path::<(String, String, String, String, String)>::extract(&req).unwrap()
                                                                        .into_inner();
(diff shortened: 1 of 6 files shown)


components/builder-api/src/server/resources/jobs.rs
```diff
@@ -432,7 +432,11 @@ fn do_get_job(req: &HttpRequest<AppState>, job_id: u64) -> Result<String> {
 
             if job.get_package_ident().fully_qualified() {
                 let builder_package_ident = BuilderPackageIdent(job.get_package_ident().into());
-                let channels = channels_for_package_ident(req, &builder_package_ident, &*conn)?;
+                let channels =
+                    channels_for_package_ident(req,
+                                               &builder_package_ident,
+                                               PackageTarget::from_str(job.get_target()).unwrap(),
```
raskchanky: Is it possible for this `unwrap` to fail or is `job.get_target()` guaranteed to return a valid target?

chefsalim: Good q - all jobs must have a valid target to be entered in the db, so we should be good with this unwrap

baumanj: The original failure in the release process that led me to this was an `unwrap` that I thought could never fail (an arg that was required by clap, see https://github.com/habitat-sh/habitat/pull/6363). It would've been *much* faster to locate the problem if that code were an `expect` originally, so I'm of the mind to always use expect at this point.

components/builder-api/src/server/resources/channels.rs
```diff
@@ -209,6 +209,21 @@ fn promote_package(req: HttpRequest<AppState>) -> HttpResponse {
...
+            match PackageTarget::from_str(t) {
+                Ok(t) => t,
+                Err(err) => {
+                    debug!("Invalid target requested: {}, err = {:?}", t, err);
+                    return HttpResponse::new(StatusCode::UNPROCESSABLE_ENTITY);
+                }
+            }
+        }
+        None => helpers::target_from_headers(&req),
+    };
```
baumanj: This block of code seems to be duplicated 3 times. Would it be possible to DRY it up?

PR #1018, opened 2019-04-11
Add no tty flags to builder-worker

chefsalim:
Need to add a couple of flags for studio builds to work with hab 0.79


raskchanky:
@chefsalim Now that https://github.com/habitat-sh/habitat/issues/6403 merged, you should be able to remove these flags after we release `0.80.0`.

PR #1027, opened 2019-04-16
adding clippy for windows

mwrock:
Resolves https://github.com/habitat-sh/habitat/issues/6213


components/builder-api/src/server/resources/origins.rs
```diff
@@ -504,7 +505,7 @@ fn list_origin_secrets(req: HttpRequest<AppState>) -> HttpResponse {
         Ok(list) => {
             // Need to map to different struct for hab cli backward compat
             let new_list: Vec<OriginSecretWithOriginId> =
-                list.into_iter().map(|s| s.into()).collect();
+                list.into_iter().map(Into::into).collect();
```
baumanj: I like using the type rather than the trait because I think it makes the code clearer
```suggestion
                list.into_iter().map(OriginSecret::into).collect();
```

baumanj: Another alternative:
```rust
            let new_list: Vec<_> = list.into_iter()
                                       .map(OriginSecretWithOriginId::from)
                                       .collect();
```

components/builder-api/src/server/resources/pkgs.rs
```diff
@@ -1447,6 +1449,7 @@ fn has_circular_deps(req: &HttpRequest<AppState>,
     }
 }
 
+#[allow(clippy::redundant_closure)]
```
baumanj: This can be fixed. The issue with changing the code to `.map(BuilderPackageTarget::to_string)` is that `BuilderPackageTarget` doesn't itself implement `ToString`, [it derefs to PackageTarget](https://github.com/habitat-sh/builder/blob/master/components/builder-db/src/models/package.rs#L662-L666), which does. I'd suggest adding
```rust
impl ToString for BuilderPackageTarget {
    fn to_string(&self) -> String { self.to_string() }
}
```
to https://github.com/habitat-sh/builder/blob/master/components/builder-db/src/models/package.rs

test/allowed_lints.txt
```diff
@@ -0,0 +1,4 @@
+clippy::module_inception
+clippy::new_ret_no_self
+clippy::new_without_default
+clippy::redundant_closure
```
baumanj: I'm still not comfortable with moving this to allowed. I know https://github.com/rust-lang/rust-clippy/issues/3071 exists, but I've yet to see it flag something in our code that *can't* be fixed, even if the automatic suggestion isn't a correct fix. And for anything we decide we really prefer with an explicit closure, we can `#[allow(…)]` it.

I'm ok with moving redundant_closure to `lints_to_fix`, but I still find this lint valuable, and I imagine issue 3071 will be fixed at some point.

chefsalim: Moving to lints_to_fix would be ok as well. Given that the clippy default is warn not deny, and there are known issues, it's also reasonable to just allow it and move on.

PR #1032, opened 2019-04-17
Use habitat-sh/rust-zmq instead of erickt/rust-zmq

christophermaier:
As can be seen by the changes in `Cargo.lock`, this is the exact same
code, but it comes from a fork we control.


christophermaier:
See also https://github.com/habitat-sh/habitat/pull/6444

Issue #1031, opened 2019-04-17
`hab origin delete` returns 500 on used-but-empty origin


pketienne:
### `hab` CLI version
0.79.1/20190410220617

### Platform Details
Ubuntu 18.04.2 LTS

### Scenario:
I was attempting to execute `hab origin delete {origin-name}` in order to delete an origin from my builder account.

### Steps to Reproduce:
1. Navigate to `bldr.habitat.sh/#/origins`.
1. Click `Create origin`.
1. (perform some operations with the origin)
1. Execute `hab origin delete {origin-name}`

### Expected Result:
`✓ Deleted origin {origin-name}.`
Or
```
✗✗✗
✗✗✗ [Origin is not unused, unable to delete.]
✗✗✗
```

### Actual Result:
```
☒ Deleting origin {origin-name}.
✗✗✗
✗✗✗ [500 Internal Server Error]
✗✗✗
```

### Notes:
The documentation at https://www.habitat.sh/docs/habitat-cli/#hab-origin states, under the "hab origin delete" command, that the command "Removes an unused/empty origin". Originally, I understood this to mean *either* an **unused** *or* **empty** origin, however, I believe the behavior of the command is intended only for origins that are *both* **unused** *and* **empty**.

I inquired about this within the habitat slack channel and was informed of the following:

> This is actually not a bug. If you've ever put packages into an origin or had packages built in an origin we can't allow deletion.

However, there was further discussion which posited that `500 Internal Server Error` may yet be a bug as it may be more appropriate to return a 400 with a relevant error message.

See also: https://forums.habitat.sh/t/receiving-error-500-on-origin-delete/1078/2

PR #1038, opened 2019-04-18
Fix origin deletion error and add tests

chefsalim:
This fixes up origin deletion where we were returning error 500 for cases where the origin was not 'empty' - eg, if a project was created for the origin. Now we will fail with a 409 conflict if the origin deletion does not succeed due to associated artifacts in the DB.

Also added a few more tests to validate the origin deletion behavior.

This resolves https://github.com/habitat-sh/builder/issues/1031

tenor-59883593

### DIFF
--- a/components/builder-api/src/server/resources/origins.rs
+++ b/components/builder-api/src/server/resources/origins.rs
@@ -66,7 +66,6 @@ use crate::db::models::{account::*,
                         secrets::*};
 
 use crate::server::{authorize::{authorize_session,
-                                check_origin_empty,
                                 check_origin_owner},
                     error::{Error,
                             Result},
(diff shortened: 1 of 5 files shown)


components/builder-api/src/server/resources/origins.rs
```diff
@@ -278,10 +277,6 @@ fn delete_origin(req: HttpRequest<AppState>) -> HttpResponse {
 
     debug!("Request to delete origin {}", &origin);
 
-    if !check_origin_empty(&req, &origin).unwrap_or(false) {
-        return HttpResponse::new(StatusCode::UNPROCESSABLE_ENTITY);
-    }
-
```
raskchanky: Why remove this and the associated function?

raskchanky:
0a20d724-74f5-11e6-8f4a-e751eeae9ff5

components/builder-api/src/server/resources/origins.rs
chefsalim: I started to beef up that function to try to have it check all possible cases but it seemed more likely that we'd miss a case and then have to come back and keep fixing missed cases when in reality if the DB deletion fails 99% of the time it is going to be because of a conflict that we'd want to flag with a 409 anyways.

raskchanky: Yeah I guess if the db has foreign key constraints between origins and packages, then trying to delete the origin will fail anyway, so that makes sense.

PR #1040, opened 2019-04-19
Allows subset of package metadata into database on force upload

eeyun:

### DIFF
--- a/components/builder-db/src/models/package.rs
+++ b/components/builder-db/src/models/package.rs
@@ -14,6 +14,8 @@ use diesel::{self,
              dsl::{count,
                    sql},
              pg::{expression::dsl::any,
+                  upsert::{excluded,
+                           on_constraint},
                   Pg,
                   PgConnection},
              prelude::*,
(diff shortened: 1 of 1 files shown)


components/builder-db/src/models/package.rs
```diff
@@ -361,19 +363,35 @@ impl Package {
...
+                  force_create: bool)
+                  -> QueryResult<Package> {
         Counter::DBCall.increment();
-        let package = diesel::insert_into(origin_packages::table).values(package)
-                                                                 .returning(ALL_COLUMNS)
-                                                                 .on_conflict_do_nothing()
-                                                                 .get_result::<Package>(conn)?;
+        let pkg;
+        if force_create {
+            pkg = diesel::insert_into(origin_packages::table).values(package)
```
chefsalim: Can we just do this always - not sure if there's any reason to distinguish between forced vs non-forced here...

eeyun: I guess we could, the default behavior is not to do anything if theres a conflict which i would assume is a faster than writing the content to the database in the case of a conflict?

If we want to do this, then we'd need to change the behavior here to write _all_ of the data to the database. Right now it just updates a couple pieces of the package metadata

raskchanky: So you'd essentially be doing the equivalent of a force upload every time you upload?  That feels like it has the potential to break people who might be relying on Builder's current behavior.

I doubt the performance of this particular call is worth worrying about.

eeyun: Probably fair. I think the safest bet is to continue with the path of minimal change and then to iterate should we need to be more extensive in the operation.

chefsalim: What I'm saying is, In the current flow, Package::create is only called in the case where there is no existing entry. Therefore, the only path for conflict handling is force upload. Therefore the added force flag is completely redundant. There's no impact to the normal upload (or perf issues) - just code that is a no-op that can be omitted.

raskchanky: Ahh, thanks.  That makes more sense now 👍

PR #1044, opened 2019-04-25
Adds build dep ingestion tool

eeyun:


support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+}
+
+installDeps() {
+    PDEPS=("core/aws-cli" "core/jq-static" "habitat/s3-bulk-uploader")
+
+    configMsg
+
+    for bin in "${!PDEPS[@]}"; do
+        HAB_AUTH_TOKEN="" hab pkg install -b "${PDEPS[$bin]}"
+    done
```
baumanj: Why not
```bash
    for bin in "${PDEPS[@]}"; do
         HAB_AUTH_TOKEN="" hab pkg install -b "$bin"
     done
```
?

Also, setting `HAB_AUTH_TOKEN` to the empty string is not the same thing as unsetting it. It would be safer to change the `HAB_AUTH_TOKEN=""` prefixed on the command to `env --unset=HAB_AUTH_TOKEN` if running this in an environment which supports the `--unset` option for `env`. If you only have POSIX `env`, you can make `installDeps` run in a subshell by changing the `{ … }` to `( … )`, which will localize the environment changes and do an explicit `unset`, like:
```bash
installDeps() (
    unset HAB_AUTH_TOKEN
    …
)
```

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+}
+
+setBucket() {
+    # Configure the bucket name to use for the upload
+    sgroups=("default" "dev" "prod" "acceptance" "live" "blue" "green")
+    for i in "${sgroups[@]}"; do
```
baumanj: Since you don't use `sgroups` array anywhere else, this could be
```bash
for i in default dev prod acceptance live blue green; do
```
Which I find a bit simpler, but it's really just a personal preference

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+        if curl -s localhost:9631/services/builder-api/"${i}" > /dev/null; then
+            bucket_name=$(curl -s localhost:9631/services/builder-api/"${i}" | jq .cfg.s3.bucket_name)
```
baumanj: This runs the curl command twice. Why not
```bash
if services=$(curl -s localhost:9631/services/builder-api/"${i}"); then
    bucket_name=$(echo "$services" | jq .cfg.s3.bucket_name)
```
?

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+    echo "The following artifacts failed on upload:"
+    for failure in "${failed_uploads[@]}"; do
+        echo "${failure}"
+    done
+    echo "########################################"
+    fi
+}
+
+putArtifacts() {
+    # shellcheck disable=SC2044
```
baumanj: I know it's kind of ugly, but considering that `/hab/tmp/harts/` was populated via a `s3 cp --recursive`, the possibility of unexpected files that could screw this up is real. Why not go with the safer version:
```bash
    while IFS= read -r -d '' file; do
        …
    done < <(find /hab/tmp/harts/ -name '*.hart' -print0)
```
?

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+        if ! hab pkg upload --url "${bldr_url}" "${file}" --force; then
+            failed_uploads+=("${bldr_url}")
+        fi
+    done
+}
+
+genS3Config() {
+    echo ""
+    read -r -p "AWS Access Key ID: " access_key_id
+        export AWS_ACCESS_KEY_ID="${access_key_id}"
```
baumanj: If you're going to read a variable for the sake of exporting it, you can do it a bit more simply:
```bash
    read -r -p "AWS Access Key ID: " AWS_ACCESS_KEY_ID
    export AWS_ACCESS_KEY_ID
```
Additionally, the export can happen before or after the setting of `AWS_ACCESS_KEY_ID`, so if you know you'll always need to export `AWS_ACCESS_KEY_ID`, that can be done elsewhere and this function can merely set it to the desired value.

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+    echo "Please provide a Builder Auth Token."
+    read -r -p "Builder Auth token: " bldr_auth_token
+        export HAB_AUTH_TOKEN="${bldr_auth_token}"
+    echo ""
+    echo "Builder auth token set to: ${HAB_AUTH_TOKEN}"
+    read -r -p "Would you like to use this auth token? [y/N]" response
+    if [[ "$response" =~ ^([yY][eE][sS]|[yY])+$ ]]; then
+        return
+    else
+        genBldrCreds
```
baumanj: Similar to above, this seems like a loop which I don't understand the purpose of

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+    echo ""
+    echo "Package ingestion can be pointed towards public builder or"
+    echo "an on-prem builder instance."
+
+    if [[ -n ${HAB_BLDR_URL} ]]; then
+       echo ""
+       echo "Builder URL configured via ENVVAR detected."
+       echo ""
+       echo "HAB_BLDR_URL=${HAB_BLDR_URL}"
+       export bldr_url="${HAB_BLDR_URL}"
```
baumanj: Why is `bldr_url` exported? Is there some other script that's using it?

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+            genS3Config
+        else
+            echo "Please reconfigure your AWS credentials and re-run s3migrate."
+            exit 1
+        fi
+    fi
+}
+
+credCheck(){
+    CREDS=( "${HOME}/.aws/credentials" "${HOME}/.aws/config" "/root/.aws/credentials" "/root/.aws/config")
```
baumanj: Same comments as before re all caps and creating a constant array for one time iteration

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+            break
+        fi
+
+        if [[ -n ${AWS_ACCESS_KEY_ID:-} ]]; then
+            if [[ -n ${AWS_SECRET_ACCESS_KEY:-} ]]; then
+                echo ""
+                echo "AWS Credentials configured via ENVVAR detected."
+                echo ""
+                echo "aws_access_key_id=${AWS_ACCESS_KEY_ID:-}"
+                echo "aws_secret_access_key=${AWS_SECRET_ACCESS_KEY:-}"
```
baumanj: We know based on the above `-n` test that `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` and non-empty strings, so the `:-` is superfluous and may hide errors later if the code is changed.

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+    echo ""
+    echo "This tool will scrape all the hart files from an s3 or minio"
+    echo "bucket and reingest them to populate build dep and tdep metadata"
+    credsConfigured=false
+
+    setBldrUrl
+
+
+
+    if [[ -n "${HAB_AUTH_TOKEN+x}" ]]; then
```
baumanj: This seems strange. The condition is true if `HAB_AUTH_TOKEN` is set to any value, including the empty string. I'd expect the test to be for a non-empty value:
```suggestion
    if [[ -n "${HAB_AUTH_TOKEN:-}" ]]; then
```

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+                echo "Setting detected credentials"
+            fi
+    else
+        echo ""
+        echo "WARNING: No Builder credentials detected!"
+        read -r -p "Would you like to generate them now? [y/N] " response
+        if [[ "$response" =~ ^([yY][eE][sS]|[yY])+$ ]]; then
+            genBldrCreds
+        else
+            echo "Please configure your builder auth-token and re-run s3migrate."
```
baumanj: Should this be `build_dep_populate.sh`? Or maybe
```suggestion
            echo "Please configure your builder auth-token and re-run $(basename $0)."
```

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+        fi
+    fi
+
+    if [ "$s3type" = "minio" ]; then
+        echo ""
+        echo "It looks like you specified an ingestion from minio!"
+
+        sgroups=("default" "dev" "prod")
+        for i in "${sgroups[@]}"; do
+            if curl -s localhost:9631/services/builder-minio/"${i}" | jq .cfg.key_id > /dev/null; then
```
baumanj: I believe this `if` will evaluate to true so long as the `curl` command doesn't error; regardless of whether `jq .cfg.key_id` matches anything. If you want different behavior consider the `--exit-status` flag to `jq`.

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+                access_key_id=$(curl -s localhost:9631/services/builder-minio/default | jq .cfg.key_id)
```
baumanj: Should this (and the next line) be using `"${i}"` rather than `default` in the URI?

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+}
+
+artifacts=()
+
+if [[ -z ${1:-} ]]; then
+    echo "Invalid Argument. Argument must be either 'minio' or 'aws'"
+    exit 1
+fi
+
+case ${1} in
```
baumanj: Change this to
```suggestion
case "${1:-}" in
```
and the above check becomes redundant

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+            if curl -s localhost:9631/services/builder-minio/"${i}" > /dev/null; then
+                minioIP=$(curl -s localhost:9631/services/builder-minio/"${i}" | jq .sys.ip)
+                echo ""
+                echo "We've detected your minio instance at: ${minioIP}!"
+                echo ""
+                read -r -p "Would you like to use this minio instance? [y/N] " response
+                if [[ "$response" =~ ^([yY][eE][sS]|[yY])+$ ]]; then
+                    echo "Setting endpoint to ${minioIP//\"}:9000"
+                    if curl -s localhost:9631/services/builder-minio/"${i}" | jq .cfg.use_ssl | grep "true"> /dev/null; then
+                        export AWS_REGION="https://${minioIP//\"}:9000"
```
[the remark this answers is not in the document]
eeyun: nope, not in this case. Minio has no concept of URIs, we use the "region" interchangeably with "endpoint-url" simply because its the closest similar abstraction in minio that we can key off of.

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+    done
+    echo "########################################"
+    fi
+}
+
+putArtifacts() {
+    # shellcheck disable=SC2044
+    for file in $(find /hab/tmp/harts/ -name '*.hart'); do
+        if ! hab pkg upload --url "${bldr_url}" "${file}" --force; then
+            failed_uploads+=("${bldr_url}")
```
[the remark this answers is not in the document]
eeyun: Either way should theoretically work since this is just for logging, BUT i actually have this as `${file} locally already. I need to push up my changes.

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+        if checkBucket "${AWS_REGION}" "${bucket_name}" >/dev/null; then
+            echo ""
+            echo "Bucket: ${bucket_name} found!"
+            echo ""
+            read -r -p "Are you sure you would like to use this bucket? [y/N] " response
+            if [[ "$response" =~ ^([yY][eE][sS]|[yY])+$ ]]; then
+                echo "Using specified bucket."
+                export S3_BUCKET=${bucket_name}
+            else
+                setBucket
```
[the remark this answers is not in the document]
eeyun: These loops are to force the user to set a value or to explicitly kill the program themselves. They're also here in the case that the user enters an incorrect value they don't have to rerun the entire program again and can instead just override the values they initially configured.

support/builder/build_dep_populate.sh
eeyun: see above

support/builder/build_dep_populate.sh
eeyun: Some of this code is some very old stuff that i did in the s3sync script. Reading this, I think you're right. But I don't know for certain. I'll change it for my tests.

support/builder/build_dep_populate.sh
eeyun: I dont see a case where anything problematic comes out of this loop as is. Any unexpected files that aren't harts would be ignored, any unexpected files that ARE harts would just be re-uploaded (which isn't really a problem) and any failures due to unexpected harts like for example some windows packages, would get logged.

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+        core_dirs=( $(aws --endpoint-url "${AWS_REGION}" s3 ls "s3://${S3_BUCKET}/${origin}/") )
+    else
+        # shellcheck disable=SC2207
+        core_dirs=( $(aws --region "${AWS_REGION}" s3 ls "s3://${S3_BUCKET}/${origin}/") )
+    fi
+
+    for dir in "${core_dirs[@]}"; do
+        if [[ "${dir}" != "PRE" ]]; then
+            # shellcheck disable=SC2207
+            artifacts+=( $(curl -s "${bldr_url}/v1/depot/channels/core/stable/pkgs/${dir}latest" | jq -r '.ident | [.origin,.name,.version,.release] | join("/")') )
```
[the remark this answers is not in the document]
eeyun: Yep

support/builder/build_dep_populate.sh
baumanj: The problem would occur if there were a match on a file that contained whitespace or glob characters (anywhere in the path). I agree that *probably* won't happen, but since it's straightforward (albeit a tad ugly) to do the right thing, I though it was worth suggesting.

support/builder/build_dep_populate.sh
```diff
@@ -0,0 +1,463 @@
...
+    for i in "${sgroups[@]}"; do
+        if curl -s localhost:9631/services/builder-api/"${i}" > /dev/null; then
+            bucket_name=$(curl -s localhost:9631/services/builder-api/"${i}" | jq .cfg.s3.bucket_name)
+            if [[ -n $bucket_name ]]; then
+                echo ""
+                echo "We've detected your minio bucket configuration set to: ${bucket_name}!"
+                read -r -p "Would you like to use this minio bucket? [y/N] " response
+                if [[ "$response" =~ ^([yY][eE][sS]|[yY])+$ ]]; then
+                    echo "Setting bucket to ${bucket_name}"
+                    bucket_name="${bucket_name//\"}"
```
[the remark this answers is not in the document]
eeyun: I... think this might be stripping a trailing slash?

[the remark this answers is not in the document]
baumanj: I thought that might be the intent, but if so the syntax for it is wrong:
```bash
$ bucket_name="foo/"
$ bucket_name="${bucket_name//\"}"
$ echo $bucket_name
foo/
```
This appears to do the trick:
```bash
$ bucket_name="${bucket_name/%\//}"
$ echo $bucket_name
foo
```
But I'm not sure if it's necessary at all, so I'll leave that to you.

support/builder/build_dep_populate.sh
baumanj: What is the point of the `jq .cfg.key_id` here? If won't affect the truth of the conditional.

Also, if we want to avoid curling the same URL three times, we could
```bash
            if minio_output=$(curl -s localhost:9631/services/builder-minio/"${i}"); then
```
then there are two options for populating the vars with the `jq` output:
```bash
                access_key_id=$(echo "$minio_output" | jq .cfg.key_id)
                secret_access_key=$(jq .cfg.secret_key <<< "$minio_output")
```

eeyun:
Ok @baumanj i think i've implemented all of your suggested changes with the exception of one. Specifically what I thought was stripping a trailing slash is not. It actually removes double quotes from within the content of a variable. When executed through a function the behavior doesnt work. Since that change amounted to very few characters in the script i opted to leave it in and added a comment to the first instance saying what we're doing.

baumanj:
> Specifically what I thought was stripping a trailing slash is not. It actually removes double quotes from within the content of a variable

Oh! I get it now. The part I has missed from the docs was that I was looking at
```bash
${bucket_name//\"}

**Comment by baumanj** [comment]
> Specifically what I thought was stripping a trailing slash is not. It actually removes double quotes from within the content of a variable

Oh! I get it now. The part I missed from the docs was that I was looking at
```bash
${bucket_name//\"}
```
And parsing it in the context of `${parameter/pattern/string}` [described here](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html), so it looked like `pattern` was an empty string, which seemed non-sensical. But upon closer reading:
> If _pattern_ begins with ‘/’, all matches of pattern are replaced with _string_.

and

> If _string_ is null, matches of _pattern_ are deleted and the / following _pattern_ may be omitted.

So, it's more like `${parameter//pattern-to-globally-delete}`, which totally makes sense.

I still think a function is preferable to a comment, so I'm curious what trouble you ran into executing it through a function. This seems to work, but maybe there's something about the use case in the script it's not capturing:
```bash
$ bucket_name='f"oo"bar'
$ echo $bucket_name
f"oo"bar
$ remove_double_quotes() { printf "%s" "${1//\"}"; }
$ echo $(remove_double_quotes "$bucket_name")
foobar
$ bucket_name="$(remove_double_quotes "$bucket_name")"
$ echo $bucket_name
foobar
```

eeyun:
Well it looks like that was pebcak, I swear i had exactly that before and it didnt work. Oh well, I've pushed up that fix as well! Ready for this to get merged I guess!

PR #1050, opened 2019-05-01
Prototype for multiple targets for a package (Prototype / WIP)

mpeck:
This is just a prototype. I wouldn't merge this as it is, but I think it is useful to get an idea of what the interaction would be like.

ss

Issues:

- Currently the UI only supports building a single target at a time. So if you start to build a windows package you have to wait until it is finished before building a linux package.
- The build latest button is a bit flaky. I noticed once during testing that it built a target that was not selected. I don't know if that was a fluke or if there is a deeper problem. The way the sidebar is integrate makes it a bit difficult to work with.
- There should most likely be some work done to clean up the css for the target selection buttons. I'm not super familiar with what custom variable names are being used so I did most of the css as a one-off.


chefsalim:
Nice.. the shape of this looks pretty good... having key functions start to include targets, etc.

chefsalim:
Closing as this will be picked up with a different issue https://github.com/habitat-sh/builder/issues/1098

PR #1051, opened 2019-05-01
Add note on using production access token when building

davidMcneil:


DEVELOPING.md
```diff
@@ -194,6 +194,10 @@ Once the build kicks off, you should be able to see the streaming logs for the b
 
 ## Developing Builder services
 
+Before building builder you must ensure that your Personal Access Token is set to the production instance of builder. If your default Personal Access Token points to production, this can be done by clearing the `HAB_AUTH_TOKEN` environment variable, or you can explicitly set it to your production token.
```
chefsalim: The second sentence is a bit confusing...  maybe remove "If your default Personal Access Token points to production," part?

It would also be great to add a segment on the core/protobuf error you were seeing as a symptom of what can happen if you don't have this set appropriately

PR #1059, opened 2019-05-11
Bump cargo deps as part of 0.81.0 release

raskchanky:
There are a bunch of deprecation warnings that I couldn't fix because they're in an upstream crate, but I did issue a PR against upstream to fix them.  We'll see what happens there.

https://github.com/fnichol/features-rs/pull/2

Update:  My PR was merged and there's a new version of the `features` crate, which I pulled into this PR.  I also added license acceptance to a few different scripts to get tests passing.


raskchanky:
@baumanj Those were not part of the original PR, but the tests for this PR were failing and that looks like the solution.  I can update the description.

PR #1070, opened 2019-05-16
Adds license acceptance to worker instances

eeyun:

### DIFF
--- a/terraform/instances.tf
+++ b/terraform/instances.tf
@@ -397,6 +397,17 @@ resource "aws_instance" "worker" {
     ]
   }
 
+  provisioner "remote-exec" {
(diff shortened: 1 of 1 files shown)


terraform/instances.tf
```diff
@@ -397,6 +397,17 @@ resource "aws_instance" "worker" {
     ]
   }
 
+  provisioner "remote-exec" {
+    inline = [
+      "sudo mkdir -p /home/ubuntu/.hab/accepted-licenses"
```
bixu: @eeyun, do you not need trailing commas here? My VSCode parser for Terraform complains if they are missing, as does Terraform...

PR #1071, opened 2019-05-16
Remove datastore binds from api and jobsrv

chefsalim:
In readiness for move to SaaS datastore to RDS, this removes the datastore binds from the builder-api and builder-jobsrv services.  The host and port will be passed in via config as needed.

### DIFF
--- a/.studiorc
+++ b/.studiorc
@@ -207,9 +207,9 @@ start-cache() {
 }
 
 declare -A svc_params=(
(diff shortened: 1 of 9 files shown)


irvingpop:
Would this break on-prem builder that depends on those binds?

chefsalim:
@irvingpop The on-prem builder repo will be updated to account for this change. There will not be any immediate impact as on-prem builder installs from a specific channel.

PR #1075, opened 2019-05-28
Redirect on-prem banner to chef marketing blog

eeyun:


chefsalim:
Should we just hide the banner for now since we're not really keeping content updated?

PR #1082, opened 2019-05-30
[WIP] Clippy fixes

eeyun:


eeyun:
Because the clippy errors cause the job to fail before we get the full list of clippy errors I'm going to have to make a few updates here before we have a final PR. Just a heads up.

test/run_clippy.sh
```diff
@@ -8,12 +8,11 @@ set -euo pipefail
 source ./support/ci/shared.sh
 
 # Because sadness
-sudo chown buildkite-agent /home/buildkite-agent
+if [[ "${BUILDKITE:-}" = "true" ]]; then
```
baumanj: This will work, but it can be simplified a bit an account of [`true` being a POSIX command](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/true.html) whose purpose is to exit 0.
```suggestion
if ${BUILDKITE:-false}; then
```

If you prefer it the other way, that's fine too.

mwrock: yeah that is better. thanks!

mwrock:
FYI: Added in changes to align this with what we are doing in the habitat re[po.

PR #1090, opened 2019-06-11
0.82.0 Post-release tasks

smacfarlane:
Update to Nightly 2019-06-06
Pin actix-web until we are ready for 1.0.0
Updates to usage of Retry

### DIFF
--- a/components/builder-worker/src/runner/publisher.rs
+++ b/components/builder-worker/src/runner/publisher.rs
@@ -12,12 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use retry::{delay::Fixed,
-            retry};
+use retry::retry;
 
 use crate::{bldr_core::{api_client::ApiClient,
                         logger::Logger},
             hab_core::{package::archive::PackageArchive,
+                       util::wait_for,
                        ChannelIdent}};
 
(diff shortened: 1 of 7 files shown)


components/builder-worker/src/runner/publisher.rs
```diff
@@ -48,20 +49,17 @@ impl Publisher {
...
+        match retry(wait_for(RETRY_WAIT, RETRIES), || {
+                  match client.x_put_package(archive, auth_token) {
+                      Ok(_) => Ok(()),
+                      Err(err) => {
+                          let msg = format!("Upload {}: {:?}", ident, err);
+                          debug!("{}", msg);
+                          logger.log(&msg);
+                          Err(err)
+                      }
+                  }
```
baumanj: Since `client.x_put_package` returns `Result<()>` already, `Ok(_) => Ok(())` is redundant. This could be simplified and made more like the above usage in components/builder-worker/src/runner/mod.rs:
```rust
                  let res = client.x_put_package(archive, auth_token);

                  if let Err(ref err) = res {
                      let msg = format!("Upload {}: {:?}", ident, err);
                      debug!("{}", msg);
                      logger.log(&msg);
                  }

                  res
```
Similarly for `client.create_channel` below.

Issue #1093, opened 2019-06-13
core/drush packages seem to be incorrectly missing


james-stocks:
### `hab` CLI version
hab 0.82.0/20190605214029

### Scenario:
Building `core/drupal` results in this failure:

```
★ Install of core/mysql-client/5.7.21/20190305213913 complete with 6 new packages installed.
   drupal: Resolved dependency 'core/mysql-client' to /hab/pkgs/core/mysql-client/5.7.21/20190305213913
» Installing core/drush
☁ Determining latest version of core/drush in the 'stable' channel
✗✗✗
✗✗✗ Package not found.
✗✗✗
   drupal: WARN: Could not find a suitable installed package for 'core/drush'
   drupal: ERROR: Resolving 'core/drush' failed, should this be built first?
   drupal: Build time: 0m15s
   drupal: Exiting on error
```

In the bldr UI, it appears that there are no packages for core/drush:
https://bldr.habitat.sh/#/pkgs/core/drush/latest

> There are no available Habitat artifacts (.hart files) for this origin/package.

However clicking the build jobs for this package shows completed builds in the stable channel e.g. https://bldr.habitat.sh/#/pkgs/core/drush/jobs/1257965619359449088

### Steps to Reproduce:

Clone [core-plans](https://github.com/habitat-sh/core-plans) and run `hab pkg build drupal`

chefsalim:
Looks like the regex pattern matching we are doing for the latest package is failing for the case where the version is only a single digit number - eg, "8" for drush.

PR #1097, opened 2019-06-17
Fix a missed case in latest version regex

chefsalim:
Fixes https://github.com/habitat-sh/builder/issues/1093

The previous regex was missing an edge case where the version was only a single digit.
This change updates the regex, and adds a new test case (single digit version).

tenor-213258055


eeyun:
Wow EXCELLENT job @chefsalim 🙏

PR #1103, opened 2019-06-19
Move Windows tests into public-cd

tduffield:


chefsalim:
Assuming that the test failures are expected @tduffield @scotthain ?

tduffield:
@chefsalim expect verify-private to fail, but not just verify.

mwrock:
I know that there is a problem with rust cannonicalizing docker mounted paths, but I cannot replicate this locally. I have a newly pulled image and am running:
```
docker run --rm -i --volume c:\dev\builder:c:\workdir --workdir C:\workdir --memory 4gb chefes/buildkite-windows powershell -Command .\test\run_clippy.ps1 .\test\unexamined_lints.txt .\test\allowed_lints.txt .\test\lints_to_fix.txt .\test\denied_lints.txt
```
This runs succesfully for me. Its also odd that this was running fine in the private pipeline.

tduffield:
Yeah, there shouldn't be much any of any difference between the windows instances in public-cd vs private-cd.

mwrock:
I have a suspicion that the `BUILDKITE` var is not getting passed through to the container and thus:
```
if($env:BUILDKITE) {
    # this will avoid a path length limit from the long buildkite working dir path
    $env:CARGO_TARGET_DIR = "c:\target"
}
```
is not running which means that the build is occuring in `c:\workdir` a mounted directory.

When I tested locally, I removed the `if` condition above because I knew that would raise problems and just assumed that var was set in CI. I'm seeing if I can repro locally with the `if` in tact where I do not set `BUILDKITE`.

mwrock:
Well it does fail locally but with a totally different error so this may not be it.

One thing that clued me into this idea was that yesterday when I injected the `RUST_BACKTRACE` var, I did that in that same conditional and was oddly not being set.

One thing that seems different with the public CI output is that it does not spit out the `docker run` command being executed. Is there any way we can see that?

mwrock:
@tduffield based on my last commit, ^^^ is definitely the problem. The `BUILDKITE` var is not set in the container:
```
BUILDKITE is''
```
I added `$env:CARGO_TARGET_DIR = "c:\target"` before running clippy regardless of the value of BUILDKITE and it succeeded.

So it seems we need to figure out why the public pipeline is not adding `--env BUILDKITE` to its `docker run`

scotthain:
@mwrock you'll need to add an environment var to pass into the docker container (by default they are not passed) similar to this: https://github.com/habitat-sh/habitat/compare/shain/cm_base#diff-9f269bb242eb760dee635951a1a1226eR28

mwrock:
ok it looks like that did it @scotthain

PR #1109, opened 2019-06-25
Update crates to move to latest openssl (0.10.x)

chefsalim:
This change bumps up the versions of reqwest and rusoto crates, deprecates use of the habitat http client crate in Builder, deprecates any direct usage of the hyper crate, and consolidates some shared code into a new HttpClient helper. Builder now can be built via cargo on newer versions of Linux that have the openssl 1.1.1 version installed by default.

With this PR, we pin reqwest to a specific version (0.9.17). Further work is going to be required to move to newer versions of reqwest, that will be tackled separately after async/await lands (soon!).

### DIFF
--- a/components/artifactory-client/src/client.rs
+++ b/components/artifactory-client/src/client.rs
@@ -16,12 +16,11 @@ use std::{collections::HashMap,
           fs::File,
           path::PathBuf};
 
-use env_proxy;
-use reqwest::{header::{Headers,
-                       UserAgent},
-              Client,
+use reqwest::{header::{HeaderMap,
+                       HeaderName,
+                       HeaderValue},
+              Body,
               Response};
-use url::Url;
 
 use crate::{config::ArtifactoryCfg,
             error::{ArtifactoryError,
@@ -31,12 +30,14 @@ use crate::hab_core::package::{PackageArchive,
                                PackageIdent,
                                PackageTarget};
 
-const USER_AGENT: &str = "Habitat-Builder";
-header! { (XJFrogArtApi, "X-JFrog-Art-Api") => [String] }
+use builder_core::http_client::{HttpClient,
+                                USER_AGENT_BLDR};
+
+const X_JFROG_ART_API: &str = "x-jfrog-art-api";
 
 #[derive(Clone)]
 pub struct ArtifactoryClient {
-    inner:       Client,
+    inner:       HttpClient,
     pub api_url: String,
     pub api_key: String,
     pub repo:    String,
@@ -44,41 +45,12 @@ pub struct ArtifactoryClient {
 
 impl ArtifactoryClient {
     pub fn new(config: ArtifactoryCfg) -> Self {
-        let mut headers = Headers::new();
-        headers.set(UserAgent::new(USER_AGENT));
-        headers.set(XJFrogArtApi(config.api_key.to_owned()));
+        let mut headers = HeaderMap::new();
+        headers.insert(USER_AGENT_BLDR.0.clone(), USER_AGENT_BLDR.1.clone());
+        headers.insert(HeaderName::from_static(X_JFROG_ART_API),
+                       HeaderValue::from_str(&config.api_key).expect("Invalid API key value"));
 
-        let mut client = Client::builder();
-        client.default_headers(headers);
-
-        let url = Url::parse(&config.api_url).expect("valid Artifactory url must be configured");
-        debug!("ArtifactoryClient checking proxy for url: {:?}", url);
-
-        if let Some(proxy_url) = env_proxy::for_url(&url).to_string() {
-            if url.scheme() == "http" {
-                debug!("Setting http_proxy to {}", proxy_url);
-                match reqwest::Proxy::http(&proxy_url) {
-                    Ok(p) => {
-                        client.proxy(p);
-                    }
-                    Err(e) => warn!("Invalid proxy, err: {:?}", e),
-                }
-            }
-
-            if url.scheme() == "https" {
-                debug!("Setting https proxy to {}", proxy_url);
-                match reqwest::Proxy::https(&proxy_url) {
-                    Ok(p) => {
-                        client.proxy(p);
-                    }
-                    Err(e) => warn!("Invalid proxy, err: {:?}", e),
-                }
-            }
-        } else {
-            debug!("No proxy configured for url: {:?}", url);
-        }
-
-        ArtifactoryClient { inner:   client.build().unwrap(),
+        ArtifactoryClient { inner:   HttpClient::new(&config.api_url, headers),
                             api_url: config.api_url,
                             api_key: config.api_key,
                             repo:    config.repo, }
@@ -97,9 +69,11 @@ impl ArtifactoryClient {
 
         let file = File::open(source_path).map_err(ArtifactoryError::IO)?;
 
+        let body: Body = file.into();
+
         let resp = match self.inner
                              .put(&url)
-                             .body(file)
+                             .body(body)
                              .send()
                              .map_err(ArtifactoryError::HttpClient)
         {
--- a/components/builder-api/src/server/error.rs
+++ b/components/builder-api/src/server/error.rs
@@ -28,6 +28,8 @@ use diesel;
 use github_api_client::HubError;
 use oauth_client::error::Error as OAuthError;
 use protobuf;
+use reqwest;
+use rusoto_core::RusotoError;
 use rusoto_s3;
 use serde_json;
 
@@ -43,22 +45,23 @@ pub enum Error {
     BadRequest,
     BuilderCore(bldr_core::Error),
     Conflict,
-    CreateBucketError(rusoto_s3::CreateBucketError),
+    CreateBucketError(RusotoError<rusoto_s3::CreateBucketError>),
     DbError(db::error::Error),
     DieselError(diesel::result::Error),
     Github(HubError),
     HabitatCore(hab_core::Error),
-    HeadObject(rusoto_s3::HeadObjectError),
+    HeadObject(RusotoError<rusoto_s3::HeadObjectError>),
+    HttpClient(reqwest::Error),
     InnerError(io::IntoInnerError<io::BufWriter<fs::File>>),
     IO(io::Error),
-    ListBuckets(rusoto_s3::ListBucketsError),
-    MultipartCompletion(rusoto_s3::CompleteMultipartUploadError),
-    MultipartUploadReq(rusoto_s3::CreateMultipartUploadError),
+    ListBuckets(RusotoError<rusoto_s3::ListBucketsError>),
+    MultipartCompletion(RusotoError<rusoto_s3::CompleteMultipartUploadError>),
+    MultipartUploadReq(RusotoError<rusoto_s3::CreateMultipartUploadError>),
     NotFound,
     OAuth(OAuthError),
-    PackageDownload(rusoto_s3::GetObjectError),
-    PackageUpload(rusoto_s3::PutObjectError),
-    PartialUpload(rusoto_s3::UploadPartError),
+    PackageDownload(RusotoError<rusoto_s3::GetObjectError>),
+    PackageUpload(RusotoError<rusoto_s3::PutObjectError>),
+    PartialUpload(RusotoError<rusoto_s3::UploadPartError>),
     PayloadError(actix_web::error::PayloadError),
     Protobuf(protobuf::ProtobufError),
     SerdeJson(serde_json::Error),
@@ -84,6 +87,7 @@ impl fmt::Display for Error {
             Error::Github(ref e) => format!("{}", e),
             Error::HabitatCore(ref e) => format!("{}", e),
             Error::HeadObject(ref e) => format!("{}", e),
+            Error::HttpClient(ref e) => format!("{}", e),
             Error::InnerError(ref e) => format!("{}", e.error()),
             Error::IO(ref e) => format!("{}", e),
             Error::ListBuckets(ref e) => format!("{}", e),
@@ -120,6 +124,7 @@ impl error::Error for Error {
             Error::Github(ref err) => err.description(),
             Error::HabitatCore(ref err) => err.description(),
             Error::HeadObject(ref err) => err.description(),
+            Error::HttpClient(ref err) => err.description(),
             Error::InnerError(ref err) => err.error().description(),
             Error::IO(ref err) => err.description(),
             Error::ListBuckets(ref err) => err.description(),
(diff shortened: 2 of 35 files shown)


components/builder-core/src/http_client.rs
```diff
@@ -0,0 +1,98 @@
...
+                }
+            }
+        } else {
+            trace!("No proxy configured for url: {:?}", url);
+        }
+
+        HttpClient(client.default_headers(headers).build().unwrap())
+    }
+
+    pub fn http_call<'a>(&self, spec: HttpCall<'a>) -> Result<Response> {
```
baumanj: I think this method (and the `HttpCall` struct as well, since it seems to be the only user) can be eliminated in favor of exposing the [`reqwest::RequestBuilder`](https://docs.rs/reqwest/0.9.18/reqwest/struct.RequestBuilder.html) interface.

If we add
```rust
impl std::ops::Deref for HttpClient {
    type Target = Client;

    fn deref(&self) -> &Self::Target { &self.0 }
}
```
and some error-converting functions, we can make calls like
```rust
let resp = match self.inner
                     .http_call(HttpCall { method:  Method::PUT,
                                           url:     &url,
                                           headers: None,
                                           params:  None,
                                           body:    Some(file.into()),
                                           auth:    Auth::None, })
                     .map_err(ArtifactoryError::BuilderCore)
{
```
into
```rust
let resp = match self.inner.put(&url).body(file).send()
```
and calls like:
```rust
let auth = if token.is_some() {
    Auth::Bearer(token.unwrap())
} else {
    Auth::None
};

let mut resp = self.inner.http_call(HttpCall { method: Method::GET,
                                                url: &url_path,
                                                headers: None,
                                                params: Some(&query),
                                                body: None,
                                                auth })?;
```
into
```rust
let mut request = self.inner.get(&url_path).query(&query);

if let Some(token) = token {
    request = request.bearer_auth(token);
}

let mut resp = request.send()?;
```

I tried going through a number of the callers to see if it didn't work well, but so far, I haven't found one. For the sake of avoiding duplicate work, [here's a diff](https://github.com/habitat-sh/builder/files/3341174/builder-pr-1109-jbauman.diff.zip) of what I did so far.

components/builder-core/src/http_client.rs
```diff
@@ -0,0 +1,98 @@
...
+              Method,
+              Proxy,
+              Response};
+
+use url::Url;
+
+use crate::error::{Error,
+                   Result};
+
+pub const BLDR_USER_AGENT: &str = "Habitat-Builder";
```
baumanj: Why is this defined here, components/builder-core/src/api_client.rs and components/builder-core/src/rpc.rs? Similarly for other constants.

Along with consolidating the definitions, it may make sense to make them `HeaderValue`s, rather than `&str`, since it appears all the usages are of the form `HeaderValue::from_static(BLDR_USER_AGENT)`. Once [generalized `const fn`s are available](https://github.com/rust-lang/rust/issues/57563),  `HeaderValue::from_static` will probably become one, allowing a declaration like:
```rust
pub const BLDR_USER_AGENT: HeaderValue = HeaderValue::from_static("Habitat-Builder");
```
but until then, we could leverage `lazy_static`:
```rust
lazy_static::lazy_static! {
    pub static ref BLDR_USER_AGENT: HeaderValue= HeaderValue::from_static("Habitat-Builder");
    pub static ref APPLICATION_JSON: HeaderValue= HeaderValue::from_static("application/json");
    // ... any other commonly used `HeaderValue`
}
```
Then the references change from `HeaderValue::from_static(BLDR_USER_AGENT)` to `BLDR_USER_AGENT.clone()`, increasing type safety.

In fact, we could go one step further, and ensure the appropriate pairing of `HeaderName` with `HeaderValue`:
```rust
lazy_static::lazy_static! {
    pub static ref USER_AGENT_BLDR: (HeaderName, HeaderValue) = (USER_AGENT, HeaderValue::from_static("Habitat-Builder"));
    // ...
}
```
Which, together with the suggestion about eliminating the need for `mut` on the `HeaderMap` binding, could make
```rust
    let mut headers = HeaderMap::new();
    headers.insert(USER_AGENT, HeaderValue::from_static(BLDR_USER_AGENT));
    headers.insert(ACCEPT, HeaderValue::from_static(APPLICATION_JSON));
    headers.insert(CONTENT_TYPE, HeaderValue::from_static(APPLICATION_JSON));
```
into:
```rust
    let header_values = vec![USER_AGENT_BLDR.clone(),
                             ACCEPT_APPLICATION_JSON.clone(),
                             CONTENT_TYPE_APPLICATION_JSON.clone()];
    let headers = HeaderMap::from_iter(header_values.into_iter());
```

components/builder-core/src/rpc.rs
```diff
@@ -66,14 +69,12 @@ impl RpcClient {
...
-
-        let mut cli = Client::builder();
-        cli.default_headers(headers);
+        let mut headers = HeaderMap::new();
+        headers.insert(USER_AGENT, HeaderValue::from_static(BLDR_USER_AGENT));
+        headers.insert(ACCEPT, HeaderValue::from_static(APPLICATION_JSON));
+        headers.insert(CONTENT_TYPE, HeaderValue::from_static(APPLICATION_JSON));
 
-        RpcClient { cli:      cli.build().unwrap(),
+        RpcClient { cli:      Client::builder().default_headers(headers).build().unwrap(),
```
baumanj: Similar to elsewhere, [this can fail due to TLS errors,](https://docs.rs/reqwest/0.9.18/reqwest/struct.ClientBuilder.html#errors) and I think it may be better to propagate that error than panic (or at least print out the error details before panicking).

components/builder-jobsrv/src/server/log_archiver/s3.rs
```diff
@@ -71,7 +72,8 @@ impl S3Archiver {
         let region = Region::from_str(config.region.as_str()).unwrap();
 
         let cred_provider = StaticProvider::new_minimal(key, secret);
-        let client = S3Client::new(RequestDispatcher::default(), cred_provider, region);
+        let http_client = HttpClient::new().expect("Rusoto http client must be availalbe");
```
baumanj: Same comment as elsewhere

components/builder-jobsrv/src/server/log_archiver/s3.rs
```diff
@@ -115,11 +117,11 @@ impl LogArchiver for S3Archiver {
             }
         };
 
-        let body = stream.concat2().wait().unwrap();
+        let bytes = stream.concat2().wait().unwrap();
```
baumanj: It's not clear to me why `unwrap` is safe here. If it is, an `expect` with why it is would be helpful.

components/github-api-client/src/error.rs
```diff
@@ -45,6 +47,7 @@ impl fmt::Display for HubError {
                         code, response)
             }
             HubError::AppAuth(ref e) => format!("GitHub App Authentication error, {}", e),
+            HubError::BuilderCore(ref e) => format!("{}", e),
```
baumanj:
```suggestion
            HubError::BuilderCore(ref e) => e.to_string(),
```
Similarly for the other `format!("{}", e),` arms

components/oauth-client/src/a2.rs
```diff
@@ -73,23 +78,24 @@ impl A2 {
...
-        headers.set(ContentType::form_url_encoded());
-
-        let mut resp = client.post(&url)
-                             .headers(headers)
-                             .body(params)
-                             .send()
-                             .map_err(Error::HttpClient)?;
+        let body = format!("client_id={}&client_secret={}&grant_type=authorization_code&code={}&\
+                            redirect_uri={}",
+                           config.client_id, config.client_secret, code, config.redirect_url);
```
baumanj: Leveraging [`format!`'s named parameters](http://doc.rust-lang.org/1.34.1/std/fmt/index.html#named-parameters) makes it easier to ensure the correct mapping and avoid introducing errors later if fields are added or removed:
```rust
        let body = format!("client_id={client_id}&client_secret={client_secret}&\
                            grant_type=authorization_code&code={code}&redirect_uri={redirect_uri}",
                           client_id = config.client_id,
                           client_secret = config.client_secret,
                           code = code,
                           redirect_uri = config.redirect_url);
```

components/oauth-client/src/active_directory.rs
```diff
@@ -72,23 +75,24 @@ impl ActiveDirectory {
...
-        let mut headers = Headers::new();
-        headers.set(Accept(vec![qitem(mime::APPLICATION_JSON)]));
-        headers.set(ContentType::form_url_encoded());
-
-        let mut resp = client.post(&url)
-                             .headers(headers)
-                             .body(params)
-                             .send()
-                             .map_err(Error::HttpClient)?;
+        let body = format!("client_id={}&client_secret={}&grant_type=authorization_code&code={}&\
```
baumanj: Same comment

components/oauth-client/src/azure_ad.rs
```diff
@@ -73,23 +78,24 @@ impl AzureAD {
...
-        let mut headers = Headers::new();
-        headers.set(Accept(vec![qitem(mime::APPLICATION_JSON)]));
-        headers.set(ContentType::form_url_encoded());
-
-        let mut resp = client.post(&url)
-                             .headers(headers)
-                             .body(params)
-                             .send()
-                             .map_err(Error::HttpClient)?;
+        let body = format!("client_id={}&client_secret={}&grant_type=authorization_code&code={}&\
```
baumanj: Same comment

components/artifactory-client/src/client.rs
```diff
@@ -31,54 +31,29 @@ use crate::hab_core::package::{PackageArchive,
...
-                    Err(e) => warn!("Invalid proxy, err: {:?}", e),
-                }
-            }
-        } else {
-            debug!("No proxy configured for url: {:?}", url);
-        }
+        let mut headers = HeaderMap::new();
+        headers.insert(USER_AGENT, HeaderValue::from_static(BLDR_USER_AGENT));
+        headers.insert(HeaderName::from_static(X_JFROG_ART_API),
+                       HeaderValue::from_str(&config.api_key).unwrap());
```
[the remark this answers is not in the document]
chefsalim: In the current builder pattern, services can (and are expected to) panic on startup if the config is bad. Will make it an expect for now.

components/builder-core/src/http_client.rs
chefsalim: This is a really good observation. Ideally we could encapsulate out needing to expose reqwest at all, but that's a lot more work, and for now going this route seems like a much simpler approach. Will make the suggested change.

components/builder-api/src/server/error.rs
```diff
@@ -28,6 +28,7 @@ use diesel;
 use github_api_client::HubError;
 use oauth_client::error::Error as OAuthError;
 use protobuf;
+use rusoto_core;
```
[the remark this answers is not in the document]
chefsalim: Yep - will fix.

components/builder-api/src/server/resources/ext.rs
```diff
@@ -94,32 +100,28 @@ fn do_validate_registry_credentials(body: Json<Body>, registry_type: &str) -> Re
...
-        Ok(c) => c,
-        Err(e) => {
-            debug!("Error: Unable to create HTTP client: {}", e);
-            return Err(Error::BadRequest);
-        }
-    };
+    let mut headers = HeaderMap::new();
+    headers.insert(USER_AGENT, HeaderValue::from_static(BLDR_USER_AGENT));
+    headers.insert(ACCEPT, HeaderValue::from_static(APPLICATION_JSON));
+    headers.insert(CONTENT_TYPE, HeaderValue::from_static(APPLICATION_JSON));
```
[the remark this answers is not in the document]
chefsalim: Yep, will consolidate into this pattern.

components/builder-api/src/server/services/s3.rs
```diff
@@ -87,7 +88,8 @@ impl S3Handler {
         let aws_id = config.key_id;
         let aws_secret = config.secret_key;
         let cred_provider = StaticProvider::new_minimal(aws_id, aws_secret);
-        let client = S3Client::new(RequestDispatcher::default(), cred_provider, region);
+        let http_client = HttpClient::new().expect("Rusoto http client must be availalbe");
```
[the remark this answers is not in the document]
chefsalim: Will change to log the error.

components/builder-api/src/server/services/s3.rs
```diff
@@ -187,12 +189,12 @@ impl S3Handler {
...
         request.key = key;
 
-        let payload = self.client.get_object(&request).sync();
+        let payload = self.client.get_object(request).sync();
         let body = match payload {
             Ok(response) => response.body,
             Err(e) => {
-                warn!("Failed to retrieve object from S3, key={}: {:?}",
-                      request.key, e);
+                warn!("Failed to retrieve object from S3, ident={}: {:?}",
```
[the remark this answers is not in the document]
chefsalim: Ok

components/builder-api/src/server/services/s3.rs
```diff
@@ -267,10 +264,15 @@ impl S3Handler {
                             should_break = true;
                         }
                         part_num += 1;
-                        request.body = Some(buffer.to_vec());
+
+                        let mut request = UploadPartRequest::default();
+                        request.key = key.to_string();
+                        request.bucket = self.bucket.clone();
+                        request.upload_id = output.upload_id.clone().unwrap(); // unwrap safe
```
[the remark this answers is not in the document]
chefsalim: Not sure... @eeyun may know.  This is mostly legacy code so inclined to leave it alone for now unless we notice any actual issues.

components/builder-core/src/http_client.rs
chefsalim: Yep, will fix.

components/builder-core/src/rpc.rs
chefsalim: Changing to an expect

components/builder-jobsrv/src/server/log_archiver/s3.rs
chefsalim: Changing to an expect

components/github-api-client/src/error.rs
chefsalim: There's a lot of this all over the place - nice to have for a different PR maybe.

components/oauth-client/src/a2.rs
chefsalim: Generally I prefer the positional formatting, but this may be useful in some cases like oauth.  Will consider for a separate PR.

chefsalim:
@baumanj Thanks for the review - some good suggestions!

PR #1116, opened 2019-07-09
Allow product name substitution in builder-web

predominant:
Signed-off-by: Graham Weldon <graham@grahamweldon.com>

Automatically places the app.name (from Store) into templates where possible.


predominant:
I have no idea how this works. Any tips?

predominant:
The tests are passing!

tenor-156012235

chefsalim:
Nice work @predominant !  You want a job as UI dev?  
LGTM but @Shadae may want to do a quick review as well.
Also just want to confirm that the app.name is correctly present in the non-spec cases.

PR #1122, opened 2019-07-12
Adds exports for oauth signup URL.

eeyun:
This should help to push the oauth signup url from our deployment config and should help to resolve https://github.com/habitat-sh/on-prem-builder/issues/131 in tandem with https://github.com/habitat-sh/on-prem-builder/pull/161


support/builder/config.sh
```diff
@@ -38,6 +38,7 @@ provider = "$OAUTH_PROVIDER"
 token_url = "$OAUTH_TOKEN_URL"
 userinfo_url = "$OAUTH_USERINFO_URL"
 redirect_url = "$OAUTH_REDIRECT_URL"
+signup_url = "$OAUTH_SIGNUP_URL"
```
chefsalim: I believe this is a no-op?

eeyun: I don't think so, the builder-api-proxy uses this variable to populate the habitat.conf.js file?

chefsalim: Yes, but this line is in the builder-api config, not the builder-api config

chefsalim: See https://github.com/habitat-sh/builder/blob/master/components/oauth-client/src/config.rs#L32 for the builder-api oauth config

eeyun: Ah sorry yes from the perspective of the app this should be a no op, this isn't used inside the actual oauth config and it doesn't need to be, its just for propagating the value to the proxy via the front end webapp. We have to have this here so that the link default gets overriden in the front end, nothing more.

chefsalim: I definitely don't understand why that would be the case - the builder-api-proxy config should be sufficient for propagating the link to the front end...

eeyun: I 100% thought this was the builder-api-proxy config: you are correct it looks like I added this to the api config as well. I'll get that cleaned up!

chefsalim:
Looking at this more, I'm not actually sure we need any of these changes... this is only an issue for the on-premises installation/configuration - for SaaS and dev environments, we will not need to change the sign-in config, as those environments are not meant for arbitrary OAuth configuration.

components/builder-api/habitat/default.toml
```diff
@@ -15,6 +15,7 @@ provider       = "github"
 token_url      = "https://github.com/login/oauth/access_token"
 userinfo_url   = "https://api.github.com/user"
 redirect_url   = ""
+signup_url     = ""
```
chefsalim: Don't think we need this

eeyun:
Indeed i believed initially we needed this as plumbing for the on-prem services to be able to configure themselves appropriately but I no longer believe that to be the case. Closing this one.

PR #1130, opened 2019-07-25
Fixes oauth sign-on for bitbucket

eeyun:

### DIFF
--- a/components/oauth-client/src/bitbucket.rs
+++ b/components/oauth-client/src/bitbucket.rs
@@ -37,12 +37,15 @@ struct AuthOk {
 
 #[derive(Deserialize)]
 pub struct UserOk {
-    pub user: User,
+    #[serde(flatten)]
+    uname: Utyped,
 }
 
 #[derive(Deserialize)]
-pub struct User {
-    pub username: String,
+#[serde(rename_all = "lowercase")]
+enum Utyped {
+    User(String),
+    Username(String),
 }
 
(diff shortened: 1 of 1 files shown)


components/oauth-client/src/bitbucket.rs
```diff
@@ -37,11 +37,6 @@ struct AuthOk {
 
 #[derive(Deserialize)]
 pub struct UserOk {
-    pub user: User,
```
chefsalim: We can't just deprecate this field as it will break other installations.  I think we need to add a username field at the top level, and then change the logic to pick one that exists. If both exist, we can just pick one of them. You'll need to make sure to test this change with at least one existing Bitbucket account to make sure we're not breaking anything.

eeyun: I did test this with a bitbucket account but, I didn't think at all about backwards compatibility. I'll get that cleaned up.

chefsalim:
Related doc link - https://developer.atlassian.com/cloud/bitbucket/bitbucket-api-changes-gdpr/?_ga=2.268450417.1692809982.1567121402-1840790169.1567121402

chefsalim:
As indicated by the documentation, the original username has been completely removed from the bitbucket API.  The only option available seems to be using the User field but that does not seem officially documented either.  We have no good choices here at this point, so this seems like the only path forward at this point.

PR #1133, opened 2019-07-29
Temporarily add bootstrap utilities to PATH, instead of binlinking

christophermaier:
When running this script on a pre-existing Linux system, we would try
to binlink common utilities that the script uses. However, this would
end up trying to overwrite system utilities like `awk` and `tar`,
which we don't necessarily want to do.

Now, instead of binlinking, we'll just prepend the paths to the
relevant binaries to `PATH`, making them available for the remainder
of the script, but leaving system binaries intact.

### DIFF
--- a/terraform/scripts/create_bootstrap_bundle.sh
+++ b/terraform/scripts/create_bootstrap_bundle.sh
@@ -137,26 +137,26 @@ archive_name=${this_bootstrap_bundle}.tar
 log "Generating archive: ${archive_name}"
 
 tar --create \
-       --verbose \
-       --file="${archive_name}" \
-       --directory="${sandbox_dir}"/hab/cache \
-       artifacts >&2
+    --verbose \
+    --file="${archive_name}" \
+    --directory="${sandbox_dir}"/hab/cache \
+    artifacts >&2
 
 # We'll need a hab binary to bootstrap ourselves; let's take the one
 # we just downloaded, shall we?
 hab_pkg_dir=$(echo "${sandbox_dir}"/hab/pkgs/core/hab/"${hab_version}"/*)
 tar --append \
-       --verbose \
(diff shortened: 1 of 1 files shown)


terraform/scripts/create_bootstrap_bundle.sh
```diff
@@ -60,11 +60,11 @@ find_if_exists() {
...
-hab pkg binlink core/aws-cli aws
-hab pkg binlink core/gawk awk
-# hab pkg binlink core/tar tar
-hab pkg binlink core/coreutils sha256sum
-hab pkg binlink core/coreutils sort
+
+# We add the relevant binaries to the beginning of PATH, rather than
+# binlink, in case we are running on a real Linux system; we don't
+# necessarily want to overwrite system binaries with the binlinks.
+export PATH="$(hab pkg binlink core/aws-cli)/bin":"$(hab pkg path core/gawk)/bin":"$(hab pkg path core/tar)/bin":"$(hab pkg path core/coreutils)/bin":${PATH}
```
raskchanky: I think you want `$(hab pkg path core/aws-cli)/bin`, yes?

chefsalim: What is the specific breakage this is fixing?

christophermaier: @raskchanky :man_facepalming:

christophermaier: @chefsalim I ran this on my Linux machine and it didn't work because it couldn't binlink over my system binaries. There's really no reason it needs to permanently overwrite system binaries to do this.

That being said, I think I'll revise this so it binlinks to a separate directory under this script's control and just add _that_ directory to the `PATH`.

PR #1141, opened 2019-08-09
Corrects cmd environment for docker exports

eeyun:


eeyun:
As a couple notes on what we're doing here. By default the exporter takes the path information from whatever is calling it. In our case our envvars are all getting cleaned out before attempting to call the docker exporter and so that information is not available to the exporter at all which lead to a failure in trying to find the docker binary to call the export.

This fix uses the strategy we use in our `pkg exec` subcommand to pull any available path information from the package being executed. Performing this operation after we've wiped the envvars and using it only within the context of the command gets us what we want... docker.

Issue #1160, opened 2019-09-05
Connected plan with no packages displays incorrect information


joderita:
### `hab` CLI version
0.83.0

### Scenario:
When connecting a new plan to bldr, if there are no packages present the UI shows the previously browsed-to package, or if there is no previous package, a spinner. The UI also does not present the Build Package button, preventing a user from building packages. 

<screenshot> 
<img width="1374" alt="Screen Shot 2019-09-05 at 8 51 26 AM" src="https://user-images.githubusercontent.com/15109764/64358003-8395a580-cfba-11e9-866b-2ececb0d55a4.png">

<img width="630" alt="Screen Shot 2019-09-05 at 8 51 34 AM" src="https://user-images.githubusercontent.com/15109764/64357979-7b3d6a80-cfba-11e9-9c3d-fb2a2ffc59c3.png">

In addition, the Plan shows up as connected when viewing the Packages tab for an origin, but the Settings tab for the package incorrectly shows that the package is not connected. 

### Steps to reproduce:
Connect a new plan in builder
Browse to the package
Browse to the packages settings

### Expected Result:
The UI displays a No packages available message, and provides a Build Package button. Additionally, the settings tab should show the plan as connected allow edit or deletion. 

### Workaround
Building a package and uploading it, or using the hab cli to start a build job will correct the UI.

PR #1161, opened 2019-09-10
Handle case where no pkgs have been built yet

scottopherson:
If a project w/ a connected plan has no built packages yet, the build button triggers a build for each possible platform. Fixes https://github.com/habitat-sh/builder/issues/1160

Signed-off-by: Scott Christopherson <scott@chef.io>


scottopherson:
Not sure at the moment why linux2 job requests return 400. The query param used, `x86_64-linux-kernel2`, seems to be correct.

chefsalim:
Looks good..  could we also add some unit and Cypress tests for this scenario

scottopherson:
@chefsalim haha oops 🤦‍♂ Added tests for the other features but not any for the new feature in this PR!  Here's some for 'new project' case:

Screen Shot 2019-09-16 at 11 11 07 AM

chefsalim:
Awesome, thanks @scottopherson

PR #1163, opened 2019-09-11
Tuning for Nginx config

eeyun:
In attempt to minimize the number of connection failures on requests we were seeing we've altered the nginx proxy config to be a little more forgiving. Increased our proxy timeouts and keepalive time. 


components/builder-api-proxy/habitat/default.toml
```diff
@@ -42,9 +42,9 @@ worker_connections        = 8000
 worker_processes          = "auto"
 worker_rlimit_nofile      = 8192
 max_body_size             = "1024m"
-proxy_send_timeout        = 60
-proxy_read_timeout        = 60
-proxy_connect_timeout     = 60
+proxy_send_timeout        = 300
+proxy_read_timeout        = 300
+proxy_connect_timeout     = 300
```
stevendanna: I would maybe keep this one lower? I don't know about the physical layout of the services, but even 60 seconds for a _connect_ timeout seems absurdly high and whoever you are connecting to is almost surely under water performance wise.

eeyun: ohhhh yeah good catch.

PR #1172, opened 2019-09-23
Handle projects that have darwin pkgs

scottopherson:
This commit updates the package-detail view, supporting code, and icon images to handle any projects that include MacOS platform targets. It also ensures errors aren't thrown if the API should ever return platforms that the UI is not aware of.

**Before**

Screen Shot 2019-09-23 at 1 23 05 PM

---

Screen Shot 2019-09-23 at 1 27 03 PM

---

**After**

Screen Shot 2019-09-23 at 1 29 15 PM

---

Screen Shot 2019-09-23 at 4 02 27 PM


components/builder-web/app/package/package-sidebar/package-sidebar.component.ts
```diff
@@ -51,6 +51,12 @@ export class PackageSidebarComponent {
            this.isNewProject ? 'Build latest versions' : `Build latest ${this.platform.name} version`;
   }
 
+  get buildButtonDisabledMessage() {
+    return this.targetIsMac ?
+      `* Builder can't build the package because a macOS plan file is not supported yet. Stay tuned.` :
```
chefsalim: I don't know if I would say "Stay tuned."... might be tuned for a very long time :)

chefsalim:
@apriofrost FYI on new text

components/builder-web/app/package/package-sidebar/package-sidebar.component.ts
scottopherson: Ah lol, well I just went with the text included in the mockup. I can drop the 'stay tuned' or change the message entirely. @apriofrost @susanev can chime in if they have any opinions on alternative messaging.

Screen Shot 2019-09-23 at 3 43 27 PM

apriofrost: @chefsalim yeah. I agree. Let's drop the "Stay tunned".

apriofrost:
Paste my comment from Slack to here:

Looking great! Only feedback is the sequence of icons for each version under the Versions tab: could we put the mac icon at the last instead of the first? So that the sequence match with what's in the platform switcher under Latest.

scottopherson:
@apriofrost updated 👍 

Screen Shot 2019-09-23 at 4 02 27 PM

apriofrost:
@scottopherson Looking good! Thank you! 💯

PR #1182, opened 2019-10-08
[WIP] Changes to bring builder into automate

jaym:
## Add SSL configuration for depot postgres client
To bring this into automate, depot needs to:
- use SSL to connect to postgres
- use username/password auth (for automate-cluster)
- use TLS authentication for standalone automate

This change doesn't address builder-jobsrv and builder-graph. For some reason, they create their database pools in a different way. I'm not sure if I can change it to be more like builder-api

## builder-web needs to be able to run at /bldr
- We need to be able to prepend /bldr to paths

For some reason, base href doesn't work. Also, relative paths don't seem to work inside the custom elements.


components/builder-db/src/config.rs
```diff
@@ -49,7 +53,11 @@ impl Default for DataStoreCfg {
                        connection_retry_ms:    300,
                        connection_timeout_sec: 3600,
                        connection_test:        false,
-                       pool_size:              (num_cpus::get() * 2) as u32, }
+                       pool_size:              (num_cpus::get() * 2) as u32,
+                       ssl_mode:               None,
```
chefsalim: Could you also please add a test for these configs here - https://github.com/habitat-sh/builder/blob/master/components/builder-api/src/config.rs#L260

components/builder-db/src/config.rs
```diff
@@ -65,6 +73,28 @@ impl fmt::Display for DataStoreCfg {
             None => connect,
         };
         connect = format!("{}@{}:{}/{}", connect, self.host, self.port, self.database);
+        let mut opts = Vec::new();
+
+        if let Some(ref m) = self.ssl_mode {
+            opts.push(format!("sslmode={}", m));
```
chefsalim: Do we also need to propagate this to IntoConnectParams below?

jaym: From what I can tell, that isn't possible as you have to create special types that can read/write on ssl connections. I'm not sure why the 2 pools use different types.

jaym: specifically, the problematic bits look like they are here:
https://github.com/habitat-sh/builder/blob/master/components/builder-db/src/pool.rs#L45-L47

the ssl-ness is in a different parameter

chefsalim: Yeah there are 2 pools .. one for Diesel queries, and one for non-Diesel (legacy).
So we should probably fix that line above with the way we have the Diesel pool creation at https://github.com/habitat-sh/builder/blob/master/components/builder-db/src/diesel_pool.rs#L37

chefsalim:
@jaym The suggested changes above should be straightforward, please let us know if you need assistance to get this updated/merged.

jaym:
we're closing this for now while we work on getting a list of all the changes that would be needed

PR #1204, opened 2019-10-23
Use relative paths in builder-web

jaym:
Doing this allows us to run builder at a different root path.
For example, automate will put builder at /bldr


scottopherson:
Tested locally and lgtm 👍

PR #1205, opened 2019-10-23
Add SSL configuration for depot postgres client

jaym:
This change allows builder to connect to postgres using SSL for
authentication


components/builder-db/src/config.rs
```diff
@@ -37,6 +37,10 @@ pub struct DataStoreCfg {
     pub connection_test: bool,
     /// Number of database connections to start in pool.
     pub pool_size: u32,
+    pub ssl_mode: Option<String>,
```
chefsalim: Looks good, should just be a few lines to add to unit test here:  https://github.com/habitat-sh/builder/blob/master/components/builder-api/src/config.rs#L260

PR #1215, opened 2019-11-01
Plumbs new project reqs through jobsrv, preserves backcompat

eeyun:
This PR does a couple of things.

a. It plumbs the new project target requirements through to the jobsrv
b. It provides a mechanism for users to delete ALL projects for a given package by passing a '*' to the query value.
c. Finally, as a goal, it mostly preserves backwards compatibility. By providing the previous behavior of the api through default values.

### DIFF
--- a/components/builder-api/src/server/helpers.rs
+++ b/components/builder-api/src/server/helpers.rs
@@ -67,6 +67,26 @@ pub fn extract_pagination_in_pages(pagination: &Query<Pagination>) -> (isize, is
     (pagination.range / PAGINATION_RANGE_MAX + 1, PAGINATION_RANGE_MAX)
 }
 
+pub fn extract_target(qtarget: &Query<Target>) -> PackageTarget {
+    match qtarget.target {
+        Some(ref t) => {
+            trace!("Query requested target = {}", t);
+            match PackageTarget::from_str(t) {
+                Ok(t) => t,
+                Err(err) => {
+                    debug!("Invalid target requested: {}, err = {:?}", t, err);
+                    debug!("USING DEFAULT = x86_64-linux");
+                    PackageTarget::from_str("x86_64-linux").unwrap()
+                }
+            }
+        }
+        None => {
+            debug!("NO TARGET PASSED. USING DEFAULT = x86_64-linux");
+            PackageTarget::from_str("x86_64-linux").unwrap()
+        }
+    }
+}
+
 // TODO: Deprecate getting target from User Agent header
 pub fn target_from_headers(req: &HttpRequest) -> PackageTarget {
     let user_agent_header = match req.headers().get(header::USER_AGENT) {
--- a/components/builder-api/src/server/resources/projects.rs
+++ b/components/builder-api/src/server/resources/projects.rs
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::env;
+use std::{env,
+          str::FromStr};
 
 use actix_web::{http::{self,
                        StatusCode},
@@ -28,7 +29,8 @@ use serde_json;
 
 use crate::protocol::jobsrv;
 
-use crate::hab_core::package::Plan;
+use crate::hab_core::package::{PackageTarget,
+                               Plan};
 
 use crate::db::models::{jobs::*,
                         origin::*,
@@ -39,7 +41,8 @@ use crate::server::{authorize::authorize_session,
                     error::Error,
                     framework::headers,
                     helpers::{self,
-                              Pagination},
+                              Pagination,
+                              Target},
                     resources::settings::do_toggle_privacy,
                     AppState};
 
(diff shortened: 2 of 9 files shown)


components/builder-api/src/server/resources/projects.rs
```diff
@@ -228,14 +238,36 @@ fn get_project(req: HttpRequest,
...
+            match PackageTarget::from_str(t) {
+                Ok(t) => t,
+                Err(err) => {
+                    debug!("Invalid target requested: {}, err = {:?}", t, err);
+                    return HttpResponse::new(StatusCode::UNPROCESSABLE_ENTITY);
+                }
+            }
+        }
+        None => {
+            warn!("NO TARGET PASSED. USING DEFAULT = x86_64-linux");
```
chefsalim: I'd recommend this be a debug as its totally expected for now

components/builder-api/src/server/resources/projects.rs
```diff
@@ -255,14 +288,36 @@ fn delete_project(req: HttpRequest,
...
```
chefsalim: Also debug..

components/builder-api/src/server/resources/projects.rs
```diff
@@ -284,7 +339,7 @@ fn update_project(req: HttpRequest,
         Err(err) => return err.into(),
     };
 
-    if body.plan_path.is_empty() {
+    if body.plan_path.is_empty() || body.target.is_empty() {
```
chefsalim: Shouldn't we do the same default target here as in the other calls?

components/builder-db/src/models/projects.rs
```diff
@@ -66,15 +66,20 @@ pub struct UpdateProject<'a> {
...
         origin_projects::table.filter(origin_projects::name.eq(name))
+                              .filter(origin_projects::target.eq(target))
                               .get_result(conn)
     }
 
-    pub fn delete(name: &str, conn: &PgConnection) -> QueryResult<usize> {
+    pub fn delete(name: &str, target: &str, conn: &PgConnection) -> QueryResult<usize> {
         Counter::DBCall.increment();
-        diesel::delete(origin_projects::table.filter(origin_projects::name.eq(name))).execute(conn)
+        if target.to_string().eq("*") {
```
chefsalim: I'm not sure about the "*".  I believe we don't have a use case (eg UI) planned for this currently.  If we do really want this, I think we'd be better off with something more structured or even a different API route just so that there's no inadvertent use of the delete all.

chefsalim:
Looks good other than a few comments @eeyun

components/builder-api/src/server/resources/projects.rs
```diff
@@ -131,6 +145,12 @@ fn create_project(req: HttpRequest,
         }
     };
 
+    let target = if body.target.is_empty() {
```
chefsalim: Hmm is the default target not working?

components/builder-api/src/server/resources/projects.rs
```diff
@@ -131,6 +145,12 @@ fn create_project(req: HttpRequest,
...
+        PackageTarget::from_str("x86_64-linux").unwrap()
+    } else {
+        PackageTarget::from_str(&body.target).unwrap()
```
chefsalim: Should check for err instead of unwrap.  Also if we're going to use this code in several places would be good to make it a helper function.

components/builder-api/src/server/resources/projects.rs
```diff
@@ -220,6 +240,7 @@ fn create_project(req: HttpRequest,
 #[allow(clippy::needless_pass_by_value)]
 fn get_project(req: HttpRequest,
                path: Path<(String, String)>,
+               body: Option<Json<ProjectGetReq>>,
```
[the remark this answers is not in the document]
eeyun: So that is actually how I had written this before. I refactored to bring get and delete in line with the previously defined behavior of create and update. Should we convert the target to queryparam for all 4 of these calls or just these two?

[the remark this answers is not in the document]
chefsalim: I think it's fine to leave the target in the body for the other calls where we already have a body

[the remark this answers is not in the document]
eeyun: sure thing!

components/builder-api/src/server/helpers.rs
```diff
@@ -67,6 +67,26 @@ pub fn extract_pagination_in_pages(pagination: &Query<Pagination>) -> (isize, is
...
+                Ok(t) => t,
+                Err(err) => {
+                    debug!("Invalid target requested: {}, err = {:?}", t, err);
+                    debug!("USING DEFAULT = x86_64-linux");
+                    PackageTarget::from_str("x86_64-linux").unwrap()
+                }
+            }
+        }
+        None => {
+            warn!("NO TARGET PASSED. USING DEFAULT = x86_64-linux");
```
chefsalim: I'd make this a debug or even a trace for as it's an expected case for now

PR #1220, opened 2019-11-08
Support TLS for memcached

stevendanna:
This adds optional support for TLS-enabled memcached servers. It
builds on #1218.

Signed-off-by: Steven Danna <steve@chef.io>


eeyun:
This looks pretty good to me, did you happen to run the functional test suite against a cluster configured with all of the new TLS goodness ?

PR #1236, opened 2019-11-19
Update provisioning script

SEAjamieD:
The provisioning script no longer lives in `src/` - updated to reflect


DEVELOPING.md
```diff
@@ -65,7 +65,7 @@ Before you can successfully build, you need to provision the OS with some basic
 
 1. Use `visudo` to grant your account the ability to do passwordless sudo. Add a line similar to the following to the end of your sudoers file: `<username> ALL=(ALL) NOPASSWD: ALL`
 
-2. Run the following provisioning script: `/src/support/linux/provision.sh`
+2. Run the following provisioning script: `/support/linux/provision.sh`
```
chefsalim: This doesn't seem correct as "/support" is not a root level dir?

SEAjamieD: maybe I was running it from in the wrong folder?  I was in `builder` and from there i don't see any sort of `src` : https://github.com/habitat-sh/builder

chefsalim: I think this step should just be:

2. Run the following provisioning script from the root of your Builder repo: 
`
./support/linux/provision.sh
`

SEAjamieD: sounds good to me! i'll update

chefsalim: Reminder to update so we can close this PR out

SEAjamieD: Thanks for the reminder!  Updated and ready to merge.

PR #1241, opened 2019-11-19
Tweaks package search to use origin membership rather than ownership

eeyun:
This is going to need some thorough testing in acceptance. Our functional tests all pass but it's hard to say how search is going to perform without a full dataset.

### DIFF
--- a/test/builder-api/src/packages.js
+++ b/test/builder-api/src/packages.js
@@ -742,6 +742,21 @@ describe('Working with packages', function () {
       request.get(`/depot/pkgs/neurosis/testapp/0.1.3/${release2}`)
         .type('application/json')
         .accept('application/json')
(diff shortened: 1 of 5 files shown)


chefsalim:
I think the perf will probably be comparable since we're replacing one join with another of a similarly sized table.

It would be good to add a new e2e test that checks for the broken scenario.

test/builder-api/src/packages.js
```diff
@@ -1312,4 +1327,54 @@ describe('Working with packages', function () {
         });
     });
   });
+
+  describe('Removing a member from an origin', function () {
```
chefsalim: Is this section left-over duplication from invitations.js ?

bixu:
🎉

PR #1243, opened 2019-11-21
Return 404 if no packages were demoted from the users request

### LINKED ISSUE #1188: Demote Returns 200 Status code on Failure
```
$ hab --version
hab 0.85.0/20190916214854

$ uname
Darwin

$ uname -r
18.7.0
```
### Scenario:
Trying to demote a Windows package (x86_64-linux) through the Builder API

### Steps to Reproduce:
Using the cli, do the following (see "Actual Result" for literal commands)
1. List channels for package
2. Demote package _without specifying a target_
3. List channels for package

### Expected Result:
The package was demoted _OR_ a 5xx status code.

#### Without Target
```
$ curl -X GET -H "Authorization: Bearer $HAB_AUTH_TOKEN" https://bldr.habitat.sh/v1/depot/channels/$HAB_ORIGIN/stable/pkgs/proget/5.1.20/20191008120812?target=x86_64-windows | jq .channels
[
  "unstable",
  "test",
  "stable"
]

$ curl -sw '%{http_code}' -X PUT -H "Authorization: Bearer $HAB_AUTH_TOKEN" https://bldr.habitat.sh/v1/depot/channels/$HAB_ORIGIN/test/pkgs/proget/5.1.20/20191008120812/demote
200

$ curl -X GET -H "Authorization: Bearer $HAB_AUTH_TOKEN" https://bldr.habitat.sh/v1/depot/channels/$HAB_ORIGIN/stable/pkgs/proget/5.1.20/20191008120812?target=x86_64-windows | jq .channels
[
  "unstable",
  "test",
  "stable"
]
```
#### With Target
```
$ curl -X GET -H "Authorization: Bearer $HAB_AUTH_TOKEN" https://bldr.habitat.sh/v1/depot/channels/$HAB_ORIGIN/stable/pkgs/proget/5.1.20/20191008120812?target=x86_64-windows | jq .channels
[
  "unstable",
  "test",
  "stable"
]

$ curl -sw '%{http_code}' -X PUT -H "Authorization: Bearer $HAB_AUTH_TOKEN" https://bldr.habitat.sh/v1/depot/channels/$HAB_ORIGIN/test/pkgs/proget/5.1.20/20191008120812/demote?target=x86_64-windows
200

$ curl -X GET -H "Authorization: Bearer $HAB_AUTH_TOKEN" https://bldr.habitat.sh/v1/depot/channels/$HAB_ORIGIN/stable/pkgs/proget/5.1.20/20191008120812?target=x86_64-windows | jq .channels
[
  "unstable",
  "stable"
]
```

**NOTE: trying to do a `promote` in the same fashion:**
#### Without Target
```
$ curl -s -X GET -H "Authorization: Bearer $HAB_AUTH_TOKEN" https://bldr.habitat.sh/v1/depot/channels/$HAB_ORIGIN/stable/pkgs/proget/5.1.20/20191008120812?target=x86_64-windows | jq .channels
[
  "unstable",
  "stable"
]

$ curl -sw '%{http_code}' -X PUT -H "Authorization: Bearer $HAB_AUTH_TOKEN" https://bldr.habitat.sh/v1/depot/channels/$HAB_ORIGIN/test/pkgs/proget/5.1.20/20191008120812/promote
404

$ curl -s -X GET -H "Authorization: Bearer $HAB_AUTH_TOKEN" https://bldr.habitat.sh/v1/depot/channels/$HAB_ORIGIN/stable/pkgs/proget/5.1.20/20191008120812?target=x86_64-windows | jq .channels
[
  "unstable",
  "stable"
]
```
#### With Target
```
$ curl -s -X GET -H "Authorization: Bearer $HAB_AUTH_TOKEN" https://bldr.habitat.sh/v1/depot/channels/$HAB_ORIGIN/stable/pkgs/proget/5.1.20/20191008120812?target=x86_64-windows | jq .channels
[
  "unstable",
  "stable"
]

$ curl -sw '%{http_code}' -X PUT -H "Authorization: Bearer $HAB_AUTH_TOKEN" https://bldr.habitat.sh/v1/depot/channels/$HAB_ORIGIN/test/pkgs/proget/5.1.20/20191008120812/promote?target=x86_64-windows
200

$ curl -X GET -H "Authorization: Bearer $HAB_AUTH_TOKEN" https://bldr.habitat.sh/v1/depot/channels/$HAB_ORIGIN/stable/pkgs/proget/5.1.20/20191008120812?target=x86_64-windows | jq .channel
[
  "unstable",
  "test",
  "stable"
]
```

smacfarlane:
Closes #1188 

This change is based on the assumption that making `target` a required parameter would be a breaking change, as we have older clients that don't set that parameter and we haven't explicitly stated a minimum required `hab` cli version. 

In addition, this change addresses the case where a user passes an invalid combination of `ident`, `target` and `channel`, so this may be a useful addition even if we want to declare `target` as a mandatory parameter.

Without this change, when a user requests a package to demoted, they specify two pieces of information, `ident` and `channel`, with `target` an optional 3rd parameter which, if not specified, is inferred from the current system (in the case of recent `hab` cli versions) or defaulted to `x86_64-linux`.   When the server runs the query, the database returns the result with the [number of rows affected](https://github.com/habitat-sh/builder/blob/master/components/builder-db/src/models/channel.rs#L350).  In the case where `ident` + `target` is not in `channel`, this results in success, 0 rows affected. While this feels like correct behavior to me, it can be very surprising to the user.

What this change does is, if a user passes a combination of `ident`, `target`, and `channel` that results in no action being taken, we return a 404, i.e. "the requested `ident`+`target` was not found in `channel`. "  We could validate that the combination is valid before running the demote, but this would always result in an extra query against the db.  The demote action itself is safe to run in the case where the input triple is not a valid combination, as it will take no action.  

This covers the case of a demote action being taken on an non-existent `ident`, an `ident` with an incorrect `target`, and a valid `ident+target` that isn't in the requested channel to be demoted from.  

### DIFF
--- a/components/builder-api/src/server/resources/channels.rs
+++ b/components/builder-api/src/server/resources/channels.rs
@@ -505,7 +505,13 @@ fn demote_package(req: HttpRequest,
                                                              channel: channel.clone() },
                                        &*conn).map_err(Error::DieselError)
     {
-        Ok(_) => {
+        Ok(0) => {
+            debug!("Requested package {} for target {} not present in channel {}",
+                   ident, target, channel);
+            return HttpResponse::new(StatusCode::NOT_FOUND);
+        }
+        Ok(result) => {
+            dbg!(result);
             match PackageChannelAudit::audit(
                 &PackageChannelAudit {
                     package_ident: BuilderPackageIdent(ident.clone()),
--- a/components/builder-api/src/server/resources/channels.rs
+++ b/components/builder-api/src/server/resources/channels.rs
@@ -508,10 +508,9 @@ fn demote_package(req: HttpRequest,
         Ok(0) => {
             debug!("Requested package {} for target {} not present in channel {}",
                    ident, target, channel);
-            return HttpResponse::new(StatusCode::NOT_FOUND);
+            HttpResponse::new(StatusCode::BAD_REQUEST);
         }
         Ok(result) => {
-            dbg!(result);
             match PackageChannelAudit::audit(
                 &PackageChannelAudit {
                     package_ident: BuilderPackageIdent(ident.clone()),
--- a/components/builder-api/src/server/resources/channels.rs
+++ b/components/builder-api/src/server/resources/channels.rs
@@ -508,9 +508,9 @@ fn demote_package(req: HttpRequest,
         Ok(0) => {
             debug!("Requested package {} for target {} not present in channel {}",
                    ident, target, channel);
-            HttpResponse::new(StatusCode::BAD_REQUEST);
+            HttpResponse::new(StatusCode::BAD_REQUEST)
         }
-        Ok(result) => {
+        Ok(_) => {
             match PackageChannelAudit::audit(
                 &PackageChannelAudit {
                     package_ident: BuilderPackageIdent(ident.clone()),


components/builder-api/src/server/resources/channels.rs
```diff
@@ -505,7 +505,13 @@ fn demote_package(req: HttpRequest,
                                                              channel: channel.clone() },
                                        &*conn).map_err(Error::DieselError)
     {
-        Ok(_) => {
+        Ok(0) => {
+            debug!("Requested package {} for target {} not present in channel {}",
+                   ident, target, channel);
+            return HttpResponse::new(StatusCode::NOT_FOUND);
```
jeremymv2: minor nit: The `return` isn't really needed here as it's implicit.

I like the solution here. Just curious, from eyeballing the [promote code path](https://github.com/habitat-sh/builder/blob/d093c118d440ec339439533f89b1156bc6347aaf/components/builder-db/src/models/channel.rs#L321) it appears that it's susceptible to the same issue perhaps?

components/builder-api/src/server/resources/channels.rs
```diff
@@ -505,7 +505,13 @@ fn demote_package(req: HttpRequest,
...
+        }
+        Ok(result) => {
+            dbg!(result);
```
chefsalim: Also remove this dbg?

components/builder-api/src/server/resources/channels.rs
smacfarlane: It's possible, I'd have to look at the return type of diesels `get_result`. Since we `?` that call to get the channel and package ids, I'd assume we'll early return an error if channel or ident+target doesn't exist, but it is worth a closer look.

components/builder-api/src/server/resources/channels.rs
```diff
@@ -505,7 +505,12 @@ fn demote_package(req: HttpRequest,
                                                              channel: channel.clone() },
                                        &*conn).map_err(Error::DieselError)
     {
-        Ok(_) => {
+        Ok(0) => {
+            debug!("Requested package {} for target {} not present in channel {}",
+                   ident, target, channel);
+            HttpResponse::new(StatusCode::BAD_REQUEST);
```
chefsalim: Removing the ending semicolon should do the trick

components/builder-api/src/server/resources/channels.rs
```diff
@@ -505,7 +505,12 @@ fn demote_package(req: HttpRequest,
...
+        }
+        Ok(result) => {
```
chefsalim: Also this can be Ok(_) as result isn't used

Issue #1217, opened 2019-11-06
[CI] Research how to incorporate `habitat/builder-worker` builds into the release pipeline


christophermaier:
The new Habitat release pipeline will generate new `core/hab` and `core/hab-plan-build` artifacts as release candidates on an ongoing basis. In the old pipeline, we would generate new builds of `habitat/builder-worker` with the new artifacts only after they were released to the stable channel, but this is too long to wait to find issues that manifest in `habitat/builder-worker`'s usage of these artifacts.

Instead, we should trigger new builds of `habitat/builder-worker` on all platforms as soon as new release candidates are promoted into the `acceptance` channel. This is the point at which the Supervisors in our Acceptance Builder instance consume new Supervisor updates; once new `habitat/builder-worker` candidates are available, they can be promoted into the `acceptance` channel as well, and service instances in the Acceptance environment can update to these artifacts as well.

If these behave appropriately, then we will have more confidence in our Habitat release artifacts.

We will need to do a bit of thinking of how to manage the promotion of these artifacts 
(if at all); when we promote from `acceptance` to `current`, the Production Builder Supervisors update, but I don't know if we would want Production `habitat/builder-worker` instances to also update, unless we could be certain that resulting builds wouldn't pull in pre-release artifacts (I'm thinking of things like getting a pre-release Supervisor bundled into exported Docker containers).

christophermaier:
Thinking a bit more, the promotion issue doesn't necessarily need to be a problem, because the `habitat/builder-worker` services in Production and Acceptance don't need to be updating from the same channel that Supervisors are updating from. For instance, Acceptance workers can update from the `acceptance` channel, but Production workers can continue updating from `stable`. 

We may also want the Acceptance workers updating themselves from the `staging` channel, now that I think about it. Part of the testing we would like to do is to use the new workers to build some packages. This could take a while, and new releases could automatically come into the `acceptance` channel in the meantime, which would complicate things. Promotion into `staging` is a manual process, which gives us more control.

I guess the trick is ensuring that `habitat/builder-worker` packages get built somehow within the context of the Habitat release candidate channel and incorporated into that manifest for proper promotion later. I'm not sure if there's a clean way to achieve that via triggering something in the Builder repository, or if we just want to manage that from the Habitat repository. :thinking:

christophermaier:
We're basically going to set up a new pipeline in the builder repository that just deals with builder-worker, and then trigger it from a Habitat release.

PR #1245, opened 2019-11-22
[CI] Trigger bootstrap bundle creation after Habitat release

christophermaier:
This adds a new Buildkite pipeline for post-Habitat release tasks,
which right now consists of just creating the bootstrap bundle.

The old code was removed, in favor of a more streamlined `hab pkg
download`-based approach.


.expeditor/scripts/generate_bootstrap_bundle.sh
```diff
@@ -0,0 +1,140 @@
...
+# Conveniently, we have just downloaded a hart file for that. To make
+# things easy on us, we can extract the stand-alone `hab` binary from
+# that hart file (this assumes Linux packages, naturally) and store it
+# at `bin/hab` in the tar archive.
+
+# The regex is to add the most general placeholders for "version" and
+# "release" in the hart name, while also avoiding similarly-named
+# packages like
+# `core-hab-launcher-${VERSION}-${RELEASE}-x86_64-linux.hart`, etc.
+hab_hart=$(find "${sandbox_dir}"/artifacts -type f -regex '.*/core-hab-[^-]*-[^-]*-x86_64-linux.hart')
```
smacfarlane: Does this run after builder-worker is built and promoted, and if not, will this always return a single hart that is the latest? 

What I'm thinking here is that if builder-worker hasn't been built and promoted before this runs, we could end up with two versions of `core/hab` in the artifacts path.  `habitat/builder-worker` deps on `core/hab-studio` which deps on `core/hab-backline` which deps on `core/hab-plan-build` which deps on `core/hab`.

.expeditor/builder_seed.txt
```diff
@@ -0,0 +1,14 @@
+# Supervisor and prerequisites
+core/hab-launcher
+core/hab
```
smacfarlane: should `core/hab-plan-build` and `core/hab-backline` also be included here?  They'll get pulled in as deps, but do we want to be explicit?  

Also, do we want to include the docker exporter for Builder?

christophermaier: All those get pulled in as dependencies... I just wanted to keep things simple and high-level.

.expeditor/scripts/generate_bootstrap_bundle.sh
christophermaier: Ugh, that is a very good catch :man_facepalming: 

Once we get the builder-worker incorporated into the pipeline (https://github.com/habitat-sh/builder/issues/1217), it won't be an issue. In the meantime, we can probably just grab the Habitat `manifest.json` file to get a concrete `hab` version. Alternatively, we could sort the results of this `find` and grab the "biggest" one.

I'll play around and see what looks best.

christophermaier: Addressed in https://github.com/habitat-sh/builder/pull/1245/commits/b55f30367d193e3ba1e6c2ef4e17b5c9145b4303

PR #1252, opened 2019-12-03
[terraform] Declaratively handle feature enablement

christophermaier:
This started as simply ensuring that Windows workers got the new `PIDS_FROM_LAUNCHER` feature enabled, but turned into allowing us to declaratively manage which features (plural) were active at any given time.

Read the commit messages for further details.

### DIFF
--- a/.expeditor/scripts/verify/terraform_validate.sh
+++ b/.expeditor/scripts/verify/terraform_validate.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# TODO: sync this with terraform/versions.tf
+readonly terraform_version=0.12.13
+readonly terraform_artifact="terraform_${terraform_version}_linux_amd64.zip"
+
+# Install Terraform
+(
+    # We do this so we don't have to contend with the binary and
+    # directory names (both "terraform") conflicting.
+    mkdir bin
+    cd bin
+    curl -O "https://releases.hashicorp.com/terraform/${terraform_version}/${terraform_artifact}"
+    # leaves a `terraform` binary in the current directory
+    unzip "${terraform_artifact}"
+)
+
+# Validate the terraform directory
+./bin/terraform validate terraform
--- a/.expeditor/scripts/verify/terraform_validate.sh
+++ b/.expeditor/scripts/verify/terraform_validate.sh
@@ -2,8 +2,22 @@
 
 set -euo pipefail
 
-# TODO: sync this with terraform/versions.tf
-readonly terraform_version=0.12.13
+# In the absence of a real Terraform parser, we can use a bit of awk
+# to extract the required version from a file.
+#
+# Given a file with contents like this:
+#
+#   terraform {
+#     required_version = "0.12.13"
+#   }
+#
+# we would extract `0.12.13` (without quotes!).
+terraform_version() {
+    versions_file="${1}"
+    awk 'BEGIN { FS=" *= *"} /required_version/ {gsub("\"","",$2); print $2}' "${versions_file}"
+}
+
+readonly terraform_version="$(terraform_version terraform/versions.tf)"
 readonly terraform_artifact="terraform_${terraform_version}_linux_amd64.zip"
 
 # Install Terraform
--- a/.expeditor/scripts/verify/terraform_validate.sh
+++ b/.expeditor/scripts/verify/terraform_validate.sh
@@ -18,4 +18,5 @@ readonly terraform_artifact="terraform_${terraform_version}_linux_amd64.zip"
 )
 
 # Validate the terraform directory
+./bin/terraform init terraform
 ./bin/terraform validate terraform
--- a/terraform/instances.tf
+++ b/terraform/instances.tf
@@ -9,11 +9,33 @@ provider "aws" {
 locals {
   # We have a few instances that run on Linux, and all should have the
   # same SystemD unit file. We declare it here to keep things DRY.
-  hab_sup_systemd_unit_content = templatefile(
+  hab_sup_service_content = templatefile(
     "${path.module}/templates/hab-sup.service.tpl",
     {
-      flags = "--auto-update --peer ${join(" ", var.peers)} --channel ${var.sup_release_channel} --listen-gossip 0.0.0.0:${var.gossip_listen_port} --listen-http 0.0.0.0:${var.http_listen_port}"
-      log_level = var.log_level
+      flags            = "--auto-update --peer ${join(" ", var.peers)} --channel ${var.sup_release_channel} --listen-gossip 0.0.0.0:${var.gossip_listen_port} --listen-http 0.0.0.0:${var.http_listen_port}"
+      log_level        = var.log_level
+      enabled_features = var.enabled_features
+    })
+
+  # Init file for Linux kernel 2 Supervisors
+  hab_sup_init_content = templatefile(
+    "${path.module}/templates/hab-sup.init.tpl",
+    {
+      flags            = "--auto-update --peer ${join(" ", var.peers)} --channel ${var.sup_release_channel} --listen-gossip 0.0.0.0:${var.gossip_listen_port} --listen-http 0.0.0.0:${var.http_listen_port}"
+      log_level        = var.log_level
+      enabled_features = var.enabled_features
+    })
+
+  # Userdata for Windows workers (the only Windows Supervisors we
+  # currently run)
+  windows_worker_user_data_content = templatefile(
+    "${path.module}/templates/windows_worker_user_data.tpl",
+    {
+      environment      = var.env
+      password         = var.admin_password
+      flags            = "--no-color --auto-update --peer ${join(" ", var.peers)} --channel ${var.sup_release_channel} --listen-gossip 0.0.0.0:${var.gossip_listen_port} --listen-http 0.0.0.0:${var.http_listen_port}"
+      bldr_url         = var.bldr_url
+      channel          = var.release_channel
       enabled_features = var.enabled_features
     })
 }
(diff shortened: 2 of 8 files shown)


terraform/instances.tf
```diff
@@ -100,7 +100,13 @@ resource "aws_instance" "api" {
   }
 
   provisioner "file" {
-    content     = data.template_file.sup_service.rendered
+    content = templatefile(
```
smacfarlane: I'm not certain we can use `templatefile` here. When we were looking at the issue with Windows workers not getting the peer list, we looked at this function and it read like it wouldn't receive any changes made as part of a terraform apply operation. 


> Functions do not participate in the dependency graph, so this function cannot be used with files that are generated dynamically during a Terraform operation.

https://www.terraform.io/docs/configuration/functions/templatefile.html

terraform/instances.tf
```diff
@@ -565,24 +589,4 @@ data "template_file" "sumo_sources_syslog" {
...
-}
-
-data "template_file" "linux2_init" {
-  template = file("${path.module}/templates/hab-sup.init")
-
-  vars = {
-    flags     = "--auto-update --peer ${join(" ", var.peers)} --channel ${var.sup_release_channel} --listen-gossip 0.0.0.0:${var.gossip_listen_port} --listen-http 0.0.0.0:${var.http_listen_port}"
-    log_level = var.log_level
-  }
 }
```
jeremymv2: There appears to be an orphaned `}` at the very end of `instances.tf`

```
11:56 $ terraform validate

Error: Argument or block definition required

  on instances.tf line 592:
 592: }
```

terraform/instances.tf
jeremymv2: I might be mistaken, but I believe if we assign the `templatefile` output to a local variable it will then be injected in the dep graph. It would also DRY up the `file` provisioner content block duplication:

```
provider "aws" {
  region  = var.aws_region
  profile = "habitat"
}

locals {
  hab_sup_service_config = templatefile(
      "${path.module}/templates/hab-sup.service.tpl",
      {
        flags = "--auto-update --peer ${join(" ", var.peers)} --channel ${var.sup_release_channel} --listen-gossip 0.0.0.0:${var.gossip_listen_port} --listen-http 0.0.0.0:${var.http_listen_port}"
        log_level = var.log_level
        enabled_features = var.enabled_features
      })
}
...
  provisioner "file" {
    content = local.hab_sup_service_config
    destination = "/home/ubuntu/hab-sup.service"
  }
```

chefsalim: Yeah the wording is a bit vague - the template file we are pointing to is always on disk, and is not dynamically generated so we should be ok?

terraform/instances.tf
christophermaier: Good catch @jeremymv2 

:thinking: I think I may take this opportunity to add a `terraform validate` stage to the pipeline.

terraform/templates/windows_worker_user_data.tpl
```diff
@@ -42,11 +42,20 @@
...
-  $configXml.Save((Join-Path $svcPath HabService.dll.config))
 
-  # Enable PIDS_FROM_LAUNCHER feature
-  $env:HAB_FEAT_PIDS_FROM_LAUNCHER=1;
+  # Update the arguments for the Supervisor
+  $launcherArgs = $configxml | select-xml -xpath "//appSettings/add[@key='launcherArgs']"
+  $launcherArgs.Node.value = "${flags}"
+
+  %{ for feature in enabled_features ~}
+
```
chefsalim: Remove blank line to be consistent

terraform/templates/windows_worker_user_data.tpl
```diff
@@ -20,7 +20,7 @@
 
   # Set Administrator password
   $admin = [adsi]("WinNT://./administrator, user")
-  $admin.psbase.invoke("SetPassword", "${password}")
+  $admin.psbase.invoke("SetPassword", "$password")
```
mwrock: not sure if removing the brackets here is right. I don't see `$password` being set in this script so I assume it must be coming from terraform.

terraform/instances.tf
christophermaier: @chefsalim That's how I was interpreting it, too.

@jeremymv2 Ah, that's what I was looking for... I'll give that a shot. The duplication was the part of this PR I liked the least.

terraform/templates/windows_worker_user_data.tpl
christophermaier: @mwrock Ahh, good catch; will fix.

christophermaier: Fixed in https://github.com/habitat-sh/builder/pull/1252/commits/6e3ab49d382c9b2cbf5bae52fd5e028fd94a00ce

terraform/instances.tf
christophermaier: Fixed in https://github.com/habitat-sh/builder/pull/1252/commits/892356a8e0e89577f7041d23010a7d1eb4950faf

terraform/templates/windows_worker_user_data.tpl
christophermaier: Fixed in https://github.com/habitat-sh/builder/pull/1252/commits/9df4ace531039d2f7894e1e11b3b4f1f5aa3b517

terraform/instances.tf
christophermaier: local variable usage in https://github.com/habitat-sh/builder/pull/1252/commits/8ce71df0fc3aefc85cc236d48c3cfebae4d73701

.expeditor/scripts/verify/terraform_validate.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# TODO: sync this with terraform/versions.tf
+readonly terraform_version=0.12.13
```
jeremymv2: Not a blocking issue, however eventually it might be nice to consolidate reading the required TF version from the [versions.tf](https://github.com/habitat-sh/builder/blob/master/terraform/versions.tf) file.

terraform/instances.tf
```diff
@@ -6,6 +6,18 @@ provider "aws" {
   profile = "habitat"
 }
 
+locals {
+  # We have a few instances that run on Linux, and all should have the
+  # same SystemD unit file. We declare it here to keep things DRY.
+  hab_sup_systemd_unit_content = templatefile(
```
chefsalim: Should we follow the same pattern for the non-Linux systemds as well by adding them all to local?

terraform/templates/windows_worker_user_data.tpl
```diff
@@ -42,11 +42,19 @@
...
+  # Update the arguments for the Supervisor
+  $launcherArgs = $configxml | select-xml -xpath "//appSettings/add[@key='launcherArgs']"
+  $launcherArgs.Node.value = "${flags}"
+
+%{ for feature in enabled_features ~}
+  $child = $configXml.CreateElement("add")
+  $child.SetAttribute("key", "ENV_HAB_FEAT_${upper(feature)}")
+  $child.SetAttribute("value", "true")
+  $configXml.configuration.appSettings.AppendChild($child)
+
```
chefsalim: Remove blank line to be consistent

.expeditor/scripts/verify/terraform_validate.sh
christophermaier: Addressed in https://github.com/habitat-sh/builder/pull/1252/commits/8cf42e72419021962f04da3c490f28da79634ac7

terraform/templates/windows_worker_user_data.tpl
christophermaier: I left the blank lines like this because it results in the rendered file being nicely delimited, which makes it easier to validate by eye.

If there are no enabled features, the resulting text looks like this:
```powershell
  ...
  # Update the arguments for the Supervisor
  $launcherArgs = $configxml | select-xml -xpath "//appSettings/add[@key='launcherArgs']"
  $launcherArgs.Node.value = "${flags}"

  $configXml.Save((Join-Path $svcPath HabService.dll.config))
  ...
```
If there are enabled features, they each get their own "stanza":
```powershell
  ...
  # Update the arguments for the Supervisor
  $launcherArgs = $configxml | select-xml -xpath "//appSettings/add[@key='launcherArgs']"
  $launcherArgs.Node.value = "${flags}"

  $child = $configXml.CreateElement("add")
  $child.SetAttribute("key", "ENV_HAB_FEAT_FOO")
  $child.SetAttribute("value", "true")
  $configXml.configuration.appSettings.AppendChild($child)

  $child = $configXml.CreateElement("add")
  $child.SetAttribute("key", "ENV_HAB_FEAT_BAR")
  $child.SetAttribute("value", "true")
  $configXml.configuration.appSettings.AppendChild($child)

  $configXml.Save((Join-Path $svcPath HabService.dll.config))
  ...
```
It certainly doesn't make a difference when the file is being consumed, but it came in handy when I was testing this out using `terraform console`. 

That being said, I'm happy to change it to whatever you prefer.

terraform/instances.tf
christophermaier: I was initially thinking "no", since this one is the only one that is being used in multiple contexts.

However, bringing them all together in the same physical space could make it easier to spot commonalities / differences between them.

I think you could make a decent case either way; I'm happy to do it in whichever way you prefer.

chefsalim: Since you're already in the code, I'd say let's do it.. it will reduce the context switching in the future.. these are all great improvements you're bringing in!

terraform/templates/windows_worker_user_data.tpl
chefsalim: :+1:  Thanks for the detailed explanation, makes sense!

terraform/instances.tf
christophermaier: Addressed in https://github.com/habitat-sh/builder/pull/1252/commits/cb7ecef7b542ff6d5cc98359b6ecd0307e89f283

PR #1253, opened 2019-12-03
Adds feature flag for job project selection

eeyun:
This is an attempt to bypass the UI change requirement for the new package flow's backend changes

### DIFF
--- a/components/builder-api/src/server/mod.rs
+++ b/components/builder-api/src/server/mod.rs
@@ -85,6 +85,7 @@ features! {
     pub mod feat {
         const List = 0b0000_0001,
         const Jobsrv = 0b0000_0010,
+        const LegacyProject = 0b0000_0011,
         const Artifactory = 0b0000_0100,
         const BuildDeps = 0b0000_1000
     }
@@ -118,6 +119,7 @@ impl AppState {
 fn enable_features(config: &Config) {
     let features: HashMap<_, _> = HashMap::from_iter(vec![("LIST", feat::List),
                                                           ("JOBSRV", feat::Jobsrv),
+                                                          ("LEGACYPROJECT", feat::LegacyProject),
                                                           ("ARTIFACTORY", feat::Artifactory),
                                                           ("BUILDDEPS", feat::BuildDeps)]);
     let features_enabled = config.api
--- a/components/builder-jobsrv/src/server/handlers.rs
+++ b/components/builder-jobsrv/src/server/handlers.rs
@@ -225,6 +225,12 @@ fn is_project_buildable(state: &AppState, project_name: &str, target: &str) -> b
         Err(_) => return false,
     };
 
+    let target = if feat::is_enabled(feat::LegacyProject) {
+        "x86_64-linux"
+    } else {
+        target
+    };
+
     match Project::get(project_name, &target, &*conn) {
         Ok(project) => project.auto_build,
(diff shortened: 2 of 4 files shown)


components/builder-api/src/server/mod.rs
```diff
@@ -118,6 +119,7 @@ impl AppState {
 fn enable_features(config: &Config) {
     let features: HashMap<_, _> = HashMap::from_iter(vec![("LIST", feat::List),
                                                           ("JOBSRV", feat::Jobsrv),
+                                                          ("OPKGSETTINGS", feat::OpkgSettings),
```
chefsalim: Minor, but might be good to rename this to something like "LEGACYPKG".. once we are stable we can just remove it

components/builder-jobsrv/src/server/handlers.rs
```diff
@@ -225,13 +225,26 @@ fn is_project_buildable(state: &AppState, project_name: &str, target: &str) -> b
...
     };
 
-    match Project::get(project_name, &target, &*conn) {
-        Ok(project) => project.auto_build,
-        Err(diesel::result::Error::NotFound) => false,
-        Err(err) => {
-            warn!("Unable to retrieve project: {:?}, error: {:?}",
-                  project_name, err);
-            false
+    if feat::is_enabled(feat::OPkgSettings) {
```
chefsalim: Recommend reversing the setting - if legacy package setting is enabled, we override the target.

Should be able to have the change be much smaller by just overriding the target instead of duplicating the rest of the code in the if...

components/builder-api/src/server/mod.rs
eeyun: :+1:

components/builder-jobsrv/src/server/handlers.rs
eeyun: oh yeah doh!

components/builder-jobsrv/src/server/scheduler.rs
```diff
@@ -536,20 +537,40 @@ impl ScheduleMgr {
...
-            Err(err) => {
-                self.log_error(&format!("Unable to retrieve project: {:?} (group: {}), error: \
-                                         {:?}",
-                                        project_name, group_id, err));
-                return Ok(None);
-            }
-        };
+        let project;
+
+        if feat::is_enabled(feat::OPkgSettings) {
```
chefsalim: Same feedback here, if legacy pkg setting is on, just override the target param

components/builder-jobsrv/src/server/handlers.rs
```diff
@@ -219,12 +219,16 @@ pub fn job_group_cancel(req: &RpcMessage, state: &AppState) -> Result<RpcMessage
...
 }
 
-fn is_project_buildable(state: &AppState, project_name: &str, target: &str) -> bool {
+fn is_project_buildable(state: &AppState, project_name: &str, mut target: &str) -> bool {
     let conn = match state.db.get_conn().map_err(Error::Db) {
         Ok(conn_ref) => conn_ref,
         Err(_) => return false,
     };
 
+    if feat::is_enabled(feat::LegacyProject) {
```
chefsalim: Minor - you don't need to make the target mut, you should just be able to rebind it by:
```
let target = "x86_64-linux";
```

PR #1278, opened 2020-01-23
Check rdeps package visibility

chefsalim:
This change adds some additional visibility checking to the rdeps APIs. Origins that default to non-package public visibility will have their packages filtered out of the rdeps calls.  Note that one of the API calls (rdeps/.../group) is experimental and may be removed at some later point. 

### DIFF
--- a/components/builder-api/src/server/resources/jobs.rs
+++ b/components/builder-api/src/server/resources/jobs.rs
@@ -15,6 +15,8 @@
 use std::{collections::HashMap,
           str::FromStr};
 
+use protobuf::RepeatedField;
+
 use actix_web::{http::{self,
                        StatusCode},
                 web::{self,
@@ -37,6 +39,7 @@ use crate::hab_core::{package::{Identifiable,
 
 use crate::db::models::{channel::*,
                         jobs::*,
+                        origin::*,
                         package::*,
                         projects::*,
                         settings::*};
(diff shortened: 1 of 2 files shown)


components/builder-api/src/server/resources/jobs.rs
```diff
@@ -122,14 +125,56 @@ fn get_rdeps(req: HttpRequest,
...
+                debug!("Skipping unauthorized non-public origin package: {}", rdep);
+                continue; // Skip any unauthorized origin packages
+            }
+        }
+
+        short_deps.push(rdep.to_string())
+    }
+
+    new_rdeps.set_rdeps(short_deps);
+    return Ok(new_rdeps);
```
jeremymv2: Minor: return is unneeded here since it's the last line of the function. It can simply be: `Ok(new_rdeps)` with no semicolon.

components/builder-api/src/server/resources/jobs.rs
```diff
@@ -157,14 +202,64 @@ fn get_rdeps_group(req: HttpRequest,
...
+        }
+
+        let mut new_group = jobsrv::JobGraphPackageReverseDependencyGroup::new();
+        new_group.set_group_id(group.get_group_id());
+        new_group.set_idents(RepeatedField::from_vec(ident_list));
+        new_groups.push(new_group)
+    }
+
+    new_rdeps.set_rdeps(new_groups);
+    return Ok(new_rdeps);
```
jeremymv2: Same minor nit here with `return` being unnecessary. Simply `Ok(new_rdeps)` suffices.

components/builder-api/src/server/resources/jobs.rs
chefsalim: Good catch .. thanks

components/builder-api/src/server/resources/jobs.rs
```diff
@@ -122,14 +125,54 @@ fn get_rdeps(req: HttpRequest,
...
+            let origin = Origin::get(&origin_name, &*conn)?;
+            origin_map.insert(origin_name.to_owned(),
+                              origin.default_package_visibility.clone());
+            origin.default_package_visibility
+        } else {
+            origin_map.get(origin_name).unwrap().clone()
+        };
+        if pv != PackageVisibility::Public && authorize_session(req, Some(&origin_name)).is_err() {
+            debug!("Skipping unauthorized non-public origin package: {}", rdep);
+            continue; // Skip any unauthorized origin packages
```
jeremymv2: I'm pondering if quietly skipping the data leaves the user in an entirely favorable situation. Perhaps they need to know that there is data missing so that they can make an adjustment?  I'm almost thinking that a "breaking" response such as a HTTP 403 or inserting "<the_origin_name>/unauthorized_for_private_visbility" as the rdep might be more helpful for anyone attempting to actually get value from the data?  Otherwise, they are left assuming that they got a comprehensive list when they might not have if unauthorized for one or more of the rdeps and likely will attempt to use the list in good faith.

Thoughts?

chefsalim: Good q - atm the rdeps API is only used as an advisory when you do a 'hab bldr job start', so the exposure is quite limited.  I'm going to follow up with some changes to that code to start sending in an auth token, and can add a brief warning there that private packages will not be shown.

jeremymv2: I think that sounds like a great plan. 👍

jeremymv2:
@chefsalim I'm curious what the cause of the consistent functional test failure could be.
When I run the tests locally, they seem to succeed:
```
    Retrieving reverse dependencies
      ✓ returns all reverse dependencies for an origin and package name
      ✓ returns group reverse dependencies for an origin and package name
      ✓ sets origin default visibility to private (for test)
      ✓ hides reverse dependencies for private origin and package name
      ✓ hides group reverse dependencies for private origin and package name
```

PR #1283, opened 2020-01-24
Add dbcall specific histogram metrics to DDog

eeyun:
We want to get some more granular metrics on response times for these specific database calls that we know are already expensive calls. The plan is to get these metrics into datadog so we can see exactly which calls are pegging and how long they're taking to respond.

### DIFF
--- a/components/builder-db/src/metrics.rs
+++ b/components/builder-db/src/metrics.rs
@@ -34,6 +34,10 @@ impl metrics::Metric for Counter {
 
 pub enum Histogram {
     DbCallTime,
+    GetAllLatestCallTime,
+    GetLatestChannelPackageCallTime,
+    GetLatestPackageCallTime,
+    ListAllChannelPackagesCallTime,
 }
 
 impl metrics::HistogramMetric for Histogram {}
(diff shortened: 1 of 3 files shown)


components/builder-db/src/metrics.rs
```diff
@@ -42,6 +46,10 @@ impl metrics::Metric for Histogram {
     fn id(&self) -> Cow<'static, str> {
         match *self {
             Histogram::DbCallTime => "db-call.call-time".into(),
+            Histogram::GetAllLatest => "db-call.call-time".into(),
```
chefsalim: I believe you're going to want these to be distinct names

components/builder-db/src/metrics.rs
```diff
@@ -34,6 +34,10 @@ impl metrics::Metric for Counter {
 
 pub enum Histogram {
     DbCallTime,
+    GetAllLatest,
```
chefsalim: It might be good to have these names include a "CallTime" ... eg, "GetAllLatestCallTime"

components/builder-db/src/metrics.rs
eeyun: ah dang you're correct, good catch

eeyun:
Feedback addressed

PR #1289, opened 2020-01-28
Jamie/1178 package tab revamp

SEAjamieD:
This ticket has been modified in that now we will only being showing all packages available regardless of whether or not a binary is involved.

Via https://github.com/habitat-sh/habitat/issues/7423, the api endpoint has been updated so that we can get back the data we need for this plus pagination.

I believe this should show everything we now require, but I'm unable to connect packages in my dev environment as it currently stands with my latest rebuild so will need some help confirming.

Screen Shot 2020-02-18 at 3 32 10 PM


components/builder-web/app/origin/origin-page/origin-packages-tab/origin-packages-tab.component.html
```diff
@@ -8,32 +8,26 @@
...
-          <span>No packages found.</span>
-        </li>
-        <li class="item" *ngFor="let pkg of packages">
-          <a [routerLink]="['/pkgs', pkg.origin, pkg.name, 'latest']">
-            <span class="column name">{{ pkg.name }}</span>
-          </a>
-          <hab-icon symbol="chevron-right"></hab-icon>
-        </li>
-      </ul>
+      <table>
```
SEAjamieD: In it's current state, this would do better as a list element, however its matches the other views and will very likely go back to being a true table soon, so I feel we should leave it as a table.

components/builder-web/stylesheets/_dialog.scss
```diff
@@ -20,15 +20,22 @@
...
     &.controls {
-      margin: $default-margin 0 0 0;
+      margin: $default-margin 0 0;
 
-      button, a {
+      button,
+      a {
         display: inline-block;
         margin-right: $default-margin;
         cursor: pointer;
```
SEAjamieD: These were unrelated to my ticket, but the linter was complaining about them so I fixed while I was looking.

components/builder-web/app/origin/origin-page/origin-packages-tab/origin-packages-tab.component.html
```diff
@@ -8,32 +8,20 @@
...
         </li>
         <li class="item" *ngFor="let pkg of packages">
           <a [routerLink]="['/pkgs', pkg.origin, pkg.name, 'latest']">
             <span class="column name">{{ pkg.name }}</span>
+            <hab-icon symbol="chevron-right" class="chevron-right"></hab-icon>
           </a>
-          <hab-icon symbol="chevron-right"></hab-icon>
         </li>
-      </ul>
+      </ol>
```
SEAjamieD: Turns out this was the only page using Tables, so I converted it over to a list element and then all the styling for it was set globally 🙌

PR #1291, opened 2020-01-30
Add api for departing from an origin

eeyun:

### DIFF
--- a/components/builder-api/doc/api.raml
+++ b/components/builder-api/doc/api.raml
@@ -878,6 +878,19 @@ securitySchemes:
                       description: Cannot transfer to self or non-members
                     500:
                       description: Internal server error
+        /depart:
+          post:
+            description: Leave an origin you no longer wish to be a member of
+            securedBy: [oauth_2_0]
+            response:
+              204:
+                description: Departure successful
+              403:
+                description: Forbidden from departing the origin
+              422:
+                description: Not currently a member of the origin
+              500:
+                description: Internal server error
         /invitations:
             /{invitationId}:
                 put:
--- a/components/builder-api/src/server/resources/origins.rs
+++ b/components/builder-api/src/server/resources/origins.rs
@@ -111,6 +111,8 @@ impl Origins {
                   web::delete().to(origin_member_delete))
            .route("/depot/origins/{origin}/transfer/{user}",
                   web::post().to(transfer_origin_ownership))
+           .route("/depot/origins/{origin}/depart",
+                  web::post().to(depart_from_origin))
            .route("/depot/origins/{origin}/invitations",
                   web::get().to(list_origin_invitations))
            .route("/depot/origins/{origin}/users/{username}/invitations",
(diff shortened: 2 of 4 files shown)


components/builder-api/doc/api.raml
```diff
@@ -878,6 +878,19 @@ securitySchemes:
                       description: Cannot transfer to self or non-members
                     500:
                       description: Internal server error
+        /depart:
+          post:
+            description: Leave an origin you no longer wish to be a member of
+            securedBy: [oauth_2_0]
+            response:
+              204:
+                description: Transfer successful
```
jeremymv2: `s/Transfer/Departure/` ?

components/builder-api/src/server/resources/origins.rs
```diff
@@ -1101,6 +1103,49 @@ fn transfer_origin_ownership(req: HttpRequest,
...
+    if !check_origin_member(&req, &origin, session.get_id()).unwrap_or(false) {
+        return HttpResponse::new(StatusCode::UNPROCESSABLE_ENTITY);
+    }
+
+    let conn = match state.db.get_conn().map_err(Error::DbError) {
+        Ok(conn_ref) => conn_ref,
+        Err(err) => return err.into(),
+    };
+
+    debug!(" Departing from origin {}", &origin);
```
chefsalim: Maybe add the username to the debug output as well, and remove extra leading space?

components/builder-api/src/server/resources/origins.rs
```diff
@@ -1101,6 +1103,49 @@ fn transfer_origin_ownership(req: HttpRequest,
...
+
+    let account_id = match Account::get(&session.get_name(), &*conn).map_err(Error::DieselError) {
```
chefsalim: I believe you don't need to do this, the session id is the account id.. so you should be able to get it via `session.get_id()`

test/builder-api/src/origins.js
```diff
@@ -257,6 +257,44 @@ describe('Related Origin API functions', function () {
...
+            request.post('/depot/origins/umbrella/depart')
+            .set('Authorization', global.weskerBearer)
+            .expect(204)
+                .end(function (err, res) {
+                    expect(res.text).to.be.empty;
+                    done(err)
+                });
+        });
+    });
+
```
chefsalim: Add a check that user is no longer a member of the origin after the depart succeeds?

components/builder-api/doc/api.raml
```diff
@@ -878,6 +878,19 @@ securitySchemes:
                       description: Cannot transfer to self or non-members
                     500:
                       description: Internal server error
+        /depart:
+          post:
+            description: Leave an origin you no longer wish to be a member of
+            securedBy: [oauth_2_0]
```
chefsalim: Side note - these calls are secured by bearer token, not oauth ... that should be a raml-wide update (separate PR)

eeyun: https://github.com/habitat-sh/builder/issues/1294

eeyun:
All feedback addressed

PR #1298, opened 2020-02-06
guardrails for deletion of a previously used origin

jeremymv2:
This PR enables deletion of a previously used origin, but only if it is within the bounds of certain explicit guardrails. 

Namely, the user must use a combination of the hab CLI and/or the Builder UI to clean up origin data from `origin_projects`, `origin_members`, `origin_secrets`, `origin_integrations`, `origin_channels` and `origin_packages`. The last of which, can only be done if the origin package is a leaf package and not in the stable channel.

guardrails


jeremymv2:
@apriofrost I think we should fast follow with some `hab` CLI messaging to help the user understand which commands to use to clean up the data.  Do you have any thoughts on this?

components/builder-api/src/server/resources/origins.rs
```diff
@@ -269,15 +272,108 @@ fn delete_origin(req: HttpRequest, path: Path<String>, state: Data<AppState>) ->
...
+        Ok(count) => {
+            let err = format!("There are {} projects remaining in origin {}. Must be zero.",
+                              count, origin);
+            return Err(Error::BuilderCore(OriginDeleteError(err)));
+        }
+        Err(e) => return Err(Error::DieselError(e)),
+    };
+
+    match OriginMember::count_origin_members(&origin, &*conn) {
+        // allow 1 - the origin owner
```
smacfarlane: Question: Could there ever a case where there is a single member that _isn't_ the owner?  and would that matter here? 

Thinking through this as I type, I suspect the answer is no, as you need to be an owner to delete an origin. It could be a problem if https://github.com/habitat-sh/builder/blob/021a5216584b194f6b6e89d0b4f73487dd2fd74d/components/builder-api/src/server/resources/origins.rs#L264 ever got moved after the pre-flight, or had a logic error in it.

components/builder-api/src/server/resources/origins.rs
```diff
@@ -269,15 +272,108 @@ fn delete_origin(req: HttpRequest, path: Path<String>, state: Data<AppState>) ->
...
+            return Err(Error::BuilderCore(OriginDeleteError(err)));
+        }
+        Err(e) => {
+            return Err(Error::DieselError(e));
+        }
+    };
+
+    match Channel::count_origin_channels(&origin, &*conn) {
+        // allow 2 - [unstable, stable] channels cannot be deleted
+        Ok(2) => {}
```
smacfarlane: Similar to the count above. If at some point we change how the channels work this could no longer be correct. I'll defer to others on if this is a concern for now though.

components/builder-db/src/models/origin.rs
```diff
@@ -121,12 +129,16 @@ impl Origin {
     }
 
     pub fn delete(origin: &str, conn: &PgConnection) -> QueryResult<()> {
+        // By this point, most of the associated origin data has already been manually deleted
+        // by the user. We ensure this by double checking the most critical tables are already empty
+        // via builder_api::server::resources::origins::origin_delete_preflight
+        // The transaction entries below explicitly enumerate deletion of all table data associated
+        // with the origin to ensure no vestigial data remains.
+        //
+        // TODO: Add origin deletion request into an origin audit table.
```
smacfarlane: 👍  We should file an issue so we can pull it into our work queue and link to it here

components/builder-db/src/models/projects.rs
```diff
@@ -95,6 +96,13 @@ impl Project {
         origin_projects::table.filter(origin_projects::origin.eq(origin))
                               .get_results(conn)
     }
+
+    pub fn count_origin_projects(origin: &str, conn: &PgConnection) -> QueryResult<i64> {
+        Counter::DBCall.increment();
+        origin_projects::table.select(count(origin_projects::id))
```
smacfarlane: Naive question: do we call `first` here because `select(count()).filter()` returns a single element Vec (or Vec-like) response?  My initial read made me think it might only ever count a single thing, but doesn't make a lot of sense.

components/builder-api/src/server/resources/origins.rs
eeyun: I think not - we have protection for all of the origin transfer and various behaviors to protect from this ever happening. An orphaned origin would be a bug.

components/builder-api/src/server/resources/origins.rs
eeyun: I think this should be Ok for now, but yeah with the channel refactor it might need to change. I don't imagine any change we make will massively complicate this.

test/builder-api/src/origins_deletion.js
```diff
@@ -0,0 +1,382 @@
+const expect = require('chai').expect;
```
eeyun: Is there any particular reason this didnt get inlined into the pre-existing origin deletion tests in origins.js ?

jeremymv2: Honestly due to formatting differences. I was able to collapse everything into `origin.js` though in https://github.com/habitat-sh/builder/pull/1298/commits/c730a638e289f05fa77022bc3012125b1917a039

components/builder-db/src/models/origin.rs
jeremymv2: I've created the issue and placed the URL into the comments in https://github.com/habitat-sh/builder/pull/1298/commits/25ace14b200f0dcaa374bb405fa01df009ca5fde

components/builder-db/src/models/projects.rs
jeremymv2: Great question @smacfarlane, I had to grok this too initially.
A couple of thoughts:

- According to the [docs](https://docs.diesel.rs/diesel/query_dsl/trait.RunQueryDsl.html#method.first) the behavior of `first`:
> Attempts to load a single record.This method is equivalent to .limit(1).get_result()

🤔 I suppose just `.get_result()` would also work but is not as explicit.

- I modeled the query after the existing [pub fn count_origin_packages](https://github.com/habitat-sh/builder/blob/46e8738ee38fda0e20f2ac33648e28a359c7db99/components/builder-db/src/models/package.rs#L632), my addition is basically a mirror of this one.

components/builder-api/src/server/resources/origins.rs
jeremymv2: I concur. If we ever changed the behavior, the first feedback would be from API tests that begin failing in the developer's studio.

components/builder-api/src/server/resources/origins.rs
```diff
@@ -269,15 +272,108 @@ fn delete_origin(req: HttpRequest, path: Path<String>, state: Data<AppState>) ->
...
+            debug!("Origin preflight determined that {} is not deletable, err = {}!",
+                   origin, err);
+            // Here we want to enrich the http response with a sanitized error
+            // by returning a 409 with a helpful message in the body.
+            HttpResponse::with_body(StatusCode::CONFLICT, Body::from_message(format!("{}", err)))
         }
     }
 }
 
+#[allow(clippy::needless_pass_by_value)]
```
chefsalim: Is this clippy allow needed?

jeremymv2: Turns out it's not! I removed in https://github.com/habitat-sh/builder/pull/1298/commits/6d356c4c2c2f130bb43a14b9c87a2c8a19314a8d

PR #1306, opened 2020-02-12
Change route to pull from origin_package_settings, create settigs on upload

eeyun:
Should be pretty straightforward, this is in service of the ui changes. It should add another migration to refresh the data that exists in the `origin_package_settings` table, alter package upload to create an entry in origin_package_settings in the case that it doesn't yet exist, and alters the origin packages route to look in the new table rather than the previous table. 

Whether we bundle it into this migration or not - I think it probably makes sense to change the 'origin_packages' table to a different name. Perhaps `origin_package_artifacts` or the like.

### DIFF
--- a/components/builder-api/src/server/resources/pkgs.rs
+++ b/components/builder-api/src/server/resources/pkgs.rs
@@ -28,6 +28,7 @@ use crate::{bldr_core::{error::Error::RpcError,
                                    PackageVisibility,
                                    SearchPackages},
                          settings::{GetOriginPackageSettings,
+                                    NewOriginPackageSettings,
                                     OriginPackageSettings}},
             hab_core::{package::{FromArchive,
                                  Identifiable,
@@ -74,6 +75,7 @@ use futures::{future::ok as fut_ok,
               Stream};
 use percent_encoding;
 use protobuf;
+use serde::ser::Serialize;
 use serde_json;
 use std::{fs::{self,
                remove_file,
@@ -855,11 +857,11 @@ fn package_privacy_toggle(req: HttpRequest,
 // Public helpers
 //
 
-pub fn postprocess_package_list(_req: &HttpRequest,
-                                packages: &[BuilderPackageIdent],
-                                count: i64,
-                                pagination: &Query<Pagination>)
-                                -> HttpResponse {
+pub fn postprocess_package_list<T: Serialize>(_req: &HttpRequest,
+                                              packages: &[T],
+                                              count: i64,
+                                              pagination: &Query<Pagination>)
+                                              -> HttpResponse {
     let (start, _) = helpers::extract_pagination(pagination);
     let pkg_count = packages.len() as isize;
     let stop = match pkg_count {
@@ -1154,7 +1156,17 @@ fn do_upload_package_finish(req: &HttpRequest,
             Ok(pkg) => pkg.visibility,
             Err(_) => {
                 match Origin::get(&ident.origin, &*conn) {
-                    Ok(o) => o.default_package_visibility,
+                    Ok(o) => {
+                        match OriginPackageSettings::create(&NewOriginPackageSettings {
+                            origin: &ident.origin,
+                            name: &ident.name,
+                            visibility: &o.default_package_visibility,
+                            owner_id: package.owner_id,
+                        }, &*conn) {
+                            Ok(pkg_settings) => pkg_settings.visibility,
+                            Err(err) => return Error::DieselError(err).into(),
+                        }
+                    },
                     Err(err) => return Error::DieselError(err).into(),
                 }
             }
--- a/components/builder-api/src/server/resources/origins.rs
+++ b/components/builder-api/src/server/resources/origins.rs
@@ -820,7 +820,9 @@ fn list_unique_packages(req: HttpRequest,
                              limit:      per_page as i64, };
 
     match Package::distinct_for_origin(lpr, &*conn) {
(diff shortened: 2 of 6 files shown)


components/builder-api/src/server/resources/pkgs.rs
```diff
@@ -913,6 +915,35 @@ pub fn postprocess_extended_package_list(_req: &HttpRequest,
...
+                                    pagination: &Query<Pagination>)
+                                    -> HttpResponse {
+    let (start, _) = helpers::extract_pagination(pagination);
+    let pkg_count = packages.len() as isize;
+    let stop = match pkg_count {
+        0 => count,
+        _ => (start + pkg_count - 1) as i64,
+    };
+
+    debug!("postprocessing package list, start: {}, stop: {}, total_count: {}",
```
jeremymv2: nit: we might wish to modify this debug just slightly to be something like `postprocessing package settings list` to distinguish it from this other very similar debug https://github.com/habitat-sh/builder/blob/ba06a2ee9b6e8940fbf67e3610ee7aa9a7e3861f/components/builder-api/src/server/resources/pkgs.rs#L870

components/builder-api/src/server/resources/pkgs.rs
```diff
@@ -17,6 +17,7 @@ use crate::{bldr_core::{error::Error::RpcError,
             db::models::{channel::Channel,
                          origin::Origin,
                          package::{BuilderPackageIdent,
+                                   BuilderPackageSetting,
```
chefsalim: Do we need a new struct or could we just leverage OriginPackageSettings?

components/builder-api/src/server/resources/pkgs.rs
```diff
@@ -1154,7 +1185,17 @@ fn do_upload_package_finish(req: &HttpRequest,
             Ok(pkg) => pkg.visibility,
             Err(_) => {
                 match Origin::get(&ident.origin, &*conn) {
-                    Ok(o) => o.default_package_visibility,
+                    Ok(o) => {
+                        match OriginPackageSettings::create(&NewOriginPackageSettings {
```
chefsalim: Is this always a create or do we need to account for update as well?

apriofrost:
I like the idea of calling it `origin_package_artifacts` . I've started to use the terminology of `package artifact` and `package plan connections` in the error messages for deleting an origin (issue link: https://github.com/habitat-sh/habitat/issues/7432). It would be good to unify the languages we use in the front-end and back-end if suitable. 😄

components/builder-api/src/server/resources/pkgs.rs
eeyun: So right now it only calls create on upload and if the entry already exists we "do nothing", basically we only ever use update through a different api which i think is probably the right call.

components/builder-api/src/server/resources/pkgs.rs
eeyun: I actually am not 100% sure on that. Will investigate it to make sure.

components/builder-api/src/server/resources/pkgs.rs
```diff
@@ -913,6 +915,35 @@ pub fn postprocess_extended_package_list(_req: &HttpRequest,
             .body(body)
 }
 
+pub fn postprocess_package_settings(_req: &HttpRequest,
+                                    packages: &[BuilderPackageSetting],
```
jeremymv2: Not a blocking issue for this PR, but I am pondering if we can potentially use a generic here to de-duplicate the `postprocess_package_settings`, `postprocess_package_list` and `postprocess_extended_package_list` functions which appear to be identical with the exception of the type of packages `Vec` passed in as an argument.

ex.
```
use serde::Serialize;

-pub fn postprocess_package_list(_req: &HttpRequest,
-                                packages: &[BuilderPackageIdent],
-                                count: i64,
-                                pagination: &Query<Pagination>)
-                                -> HttpResponse {
+pub fn postprocess_package_list<T>(_req: &HttpRequest,
+                                   packages: &Vec<T>,
+                                   count: i64,
+                                   pagination: &Query<Pagination>)
+                                   -> HttpResponse
+    where T: Serialize
+{
```

components/builder-api/src/server/resources/pkgs.rs
eeyun: So it looks like we probably CAN do this with the existing struct, but doing so will require a bit of a refactor and we may end up returning more data on this call but that shouldn't be a deal breaker of any kind.

components/builder-api/src/server/resources/pkgs.rs
eeyun: It is possible, it's something i was tinkering with and then ultimately scrapped just to get this iteration in. I'll revisit in the a.m. to see if i can get it to work in a sane way.

components/builder-api/src/server/resources/pkgs.rs
chefsalim: Cool... I wouldn't be too worried if there's a bit more data returned on this call if we can simplify the code and concepts

components/builder-api/src/server/resources/pkgs.rs
chefsalim: Might be good to add a test for the cases - first upload (create) and second upload (create that does nothing)

eeyun: I believe those tests already exist but I'll see if there's any expansion we can do. We test that behavior in the settings module already I believe.

eeyun:
All feedback addressed

PR #1307, opened 2020-02-12
[CI] Initial creation of builder-worker pipeline

christophermaier:
This is an initial stub, just to allow Expeditor to create the
pipeline. Once that exists, we can start filling in details.

### DIFF
--- a/.expeditor/builder_worker.pipeline.yml
+++ b/.expeditor/builder_worker.pipeline.yml
@@ -1,0 +1,6 @@
+steps:
+  - label: "Build the worker"
+    command: "echo 'Building, building, building...'"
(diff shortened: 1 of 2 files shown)


christophermaier:
@smacfarlane I wanted a way to be able to verify they worked before promoting them to stable, so I was thinking of another pipeline, similar to what we have with Habitat itself. It builds all three platform packages, then we can promote them to run in acceptance, build something, make sure it works, then promote them to stable. There's a bit of code to copy to make that happen, but I was hoping to keep the pattern as close as possible, so we can kind of standardize on things. I also wanted to automate as much of this as possible.

Issue #1300, opened 2020-02-06
add an origin audit table to Builder postgres


jeremymv2:
### Describe the Enhancement:
Related to the origin deletion work in https://github.com/habitat-sh/builder/pull/1298, we should follow up with introducing an audit table specifically for origins.

### Describe the Need:
Having such a table provides the Builder administrators an audit trail for visibility into origin changes.

PR #1310, opened 2020-02-13
audit trail for key origin operations

jeremymv2:
Closes https://github.com/habitat-sh/builder/issues/1300

Key origin operations are logged.

sherlock

### DIFF
--- a/components/builder-api/src/server/resources/origins.rs
+++ b/components/builder-api/src/server/resources/origins.rs
@@ -213,7 +213,15 @@ fn create_origin(req: HttpRequest,
                                  default_package_visibility: &dpv, };
 
     match Origin::create(&new_origin, &*conn).map_err(Error::DieselError) {
-        Ok(origin) => HttpResponse::Created().json(origin),
+        Ok(origin) => {
+            origin_audit(&body.0.name,
+                         OriginOperation::OriginCreate,
+                         &body.0.name,
+                         session.get_id() as i64,
+                         session.get_name(),
+                         &*conn);
+            HttpResponse::Created().json(origin)
+        }
         Err(err) => {
             debug!("{}", err);
             err.into()
@@ -277,7 +285,15 @@ fn delete_origin(req: HttpRequest, path: Path<String>, state: Data<AppState>) ->
     match origin_delete_preflight(&origin, &*conn) {
         Ok(_) => {
             match Origin::delete(&origin, &*conn).map_err(Error::DieselError) {
-                Ok(_) => HttpResponse::NoContent().into(),
+                Ok(_) => {
+                    origin_audit(&origin,
+                                 OriginOperation::OriginDelete,
+                                 &origin,
+                                 session.get_id() as i64,
+                                 session.get_name(),
+                                 &*conn);
+                    HttpResponse::NoContent().into()
+                }
                 Err(err) => {
                     debug!("Origin {} deletion failed! err = {}", origin, err);
                     // We do not want to expose any database details from diesel
@@ -1192,7 +1208,15 @@ fn transfer_origin_ownership(req: HttpRequest,
     }
 
     match Origin::transfer(&origin, recipient_id, &*conn).map_err(Error::DieselError) {
-        Ok(_) => HttpResponse::NoContent().finish(),
+        Ok(_) => {
+            origin_audit(&origin,
+                         OriginOperation::OwnerTransfer,
+                         &recipient_id.to_string(),
+                         session.get_id() as i64,
+                         session.get_name(),
+                         &*conn);
+            HttpResponse::NoContent().finish()
+        }
         Err(err) => {
             debug!("{}", err);
             err.into()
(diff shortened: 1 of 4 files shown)


components/builder-api/src/server/resources/origins.rs
```diff
@@ -213,7 +213,22 @@ fn create_origin(req: HttpRequest,
...
-        Ok(origin) => HttpResponse::Created().json(origin),
+        Ok(origin) => {
+            match OriginAudit::audit(&OriginAudit { operation:      OriginOperation::OriginCreate,
+                                                    origin:         &body.0.name,
+                                                    target_object:  &body.0.name,
+                                                    requester_id:   session.get_id() as i64,
+                                                    requester_name: session.get_name(), },
+                                     &*conn)
+            {
+                Ok(_) => HttpResponse::Created().json(origin),
```
chefsalim: Just minor note that you can simplify/DRY this by just doing an if instead of a match.
Eg something like:
```
if let Err(err) = OriginAudit::audit(...) { debug!("....") }
HttpResponse::Created().json(origin)
```
Similarly in the other sections below

jeremymv2: Brilliant! I wasn't 100% happy with that but couldn't find a better way. `if let` to the rescue!

components/builder-api/src/server/resources/origins.rs
```diff
@@ -213,7 +213,22 @@ fn create_origin(req: HttpRequest,
                                  default_package_visibility: &dpv, };
 
     match Origin::create(&new_origin, &*conn).map_err(Error::DieselError) {
-        Ok(origin) => HttpResponse::Created().json(origin),
+        Ok(origin) => {
+            match OriginAudit::audit(&OriginAudit { operation:      OriginOperation::OriginCreate,
+                                                    origin:         &body.0.name,
+                                                    target_object:  &body.0.name,
+                                                    requester_id:   session.get_id() as i64,
+                                                    requester_name: session.get_name(), },
```
christophermaier: A point of style on this... When creating new types like this, I have often found it useful to keep the members private and instead expose a functional interface to them, rather than embedding their internal structure at all the call sites. You can pass in the data you need as parameters, which can give you both cleaner code at the call site, as well as the flexibility to rearrange the internal workings of your type later.

This usecase seems somewhat similar to the Supervisor's [event stream interface](https://github.com/habitat-sh/habitat/blob/master/components/sup/src/event.rs)... all the call sites in the broader Supervisor call functions like `service_stopped(service: &Service)`, while the details of creating and sending that event are encapsulated away within the `event` module itself.

Here, you could plausibly end up with something like `audit_origin_creation(&origin_name, &session)`, and hide everything else away.

It's certainly not a blocker, but it's an approach that I've found quite valuable in the past; maybe it could work here, as well.

jeremymv2: I like that! I was pondering a better way. I'll re-work it following your suggestion.

components/builder-db/src/migrations/2020-02-12-195406_audit_origin_table/up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE SEQUENCE IF NOT EXISTS audit_origin_id_seq;
+CREATE TYPE origin_operation AS ENUM ('origin_create', 'origin_delete', 'owner_transfer');
+
+CREATE TABLE IF NOT EXISTS audit_origin (
+    id bigint DEFAULT next_id_v1('audit_origin_id_seq') PRIMARY KEY NOT NULL,
+    operation origin_operation,
+    target_object text,
```
smacfarlane: :thinking:  I had to pause when I looked at this in the db the first time.  It can store either an origin or an account id, which may be ok, I'll defer to @chefsalim and @eeyun  on that,  but I wonder if this should be nullable and only store target_account_id?  Does it provide value to store the origin here?

components/builder-api/src/server/resources/origins.rs
```diff
@@ -213,7 +213,22 @@ fn create_origin(req: HttpRequest,
                                  default_package_visibility: &dpv, };
 
     match Origin::create(&new_origin, &*conn).map_err(Error::DieselError) {
-        Ok(origin) => HttpResponse::Created().json(origin),
+        Ok(origin) => {
+            match OriginAudit::audit(&OriginAudit { operation:      OriginOperation::OriginCreate,
```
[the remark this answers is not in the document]
smacfarlane: I type too slow :joy:

components/builder-db/src/migrations/2020-02-12-195406_audit_origin_table/up.sql
jeremymv2: Great question! I originally had a more extensive list of operation types such as: `IntegrationDelete`, `SecretDelete`, `ChannelDelete`.  I have since removed those additional operations, but I could certainly see them (or others) being required in the future. For that reason, I chose a `target_object` field to hold the value of the "thing" being operated open. Does that answer your question?

components/builder-api/src/server/resources/origins.rs
jeremymv2: Handled in https://github.com/habitat-sh/builder/pull/1310/commits/5ce40956fd473db4c9850791c206c3a1dca8829a

components/builder-db/src/migrations/2020-02-12-195406_audit_origin_table/up.sql
smacfarlane: I think so.  I have vague memories of being bitten by Uber-columns in a previous life, but the alternatives weren't any better.  But that's something we can address when we get there :smiley_cat:

Issue #1263, opened 2019-12-09
Upgrade builder-datastore


chefsalim:
Currently, we are on major version 9 of Postgres in builder-datastore, and on version 11 of Postgres in our acceptance and production environments.  We should upgrade the version of Postgres in the builder-datastore package.

PR #1315, opened 2020-02-20
Upgrades the builder-datastore package to postgres 11

eeyun:


eeyun:
closes https://github.com/habitat-sh/builder/issues/1263

eeyun:
Actually this is correct. builder-datastore isn't a builder component, its just a wrapper package around postgres so the bb subcommands shouldn't work for it.

PR #1318, opened 2020-02-20
Update cargo.lock after Habitat release

### LINKED ISSUE #1232: Automate bumping Cargo dependencies after a Habitat release
This is basically the same thing as https://github.com/habitat-sh/habitat/issues/7199, but for Builder.

smacfarlane:
This pulls in the relevant bits for updating the Cargo.lock after a release from the [habitat repository](https://github.com/habitat-sh/habitat).

Closes #1232

### DIFF
--- a/.expeditor/scripts/post_habitat_release/shared.sh
+++ b/.expeditor/scripts/post_habitat_release/shared.sh
@@ -31,3 +31,52 @@ worker_ident_for_target() {
     target="${1}"
     buildkite-agent meta-data get "${target}-builder-worker"
 }
+
+install_rustup() {
+  if command -v rustup && command -v cargo &>/dev/null; then
+    echo "--- :rust: rustup is currently installed."
+  else
+    echo "--- :rust: Installing rustup."
+    curl https://sh.rustup.rs -sSf | sh -s -- --no-modify-path -y --profile=minimal
+    # shellcheck disable=SC1090
+    source "$HOME"/.cargo/env
+  fi
+}
+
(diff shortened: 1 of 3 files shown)


.expeditor/scripts/post_habitat_release/shared.sh
```diff
@@ -31,3 +31,52 @@ worker_ident_for_target() {
     target="${1}"
     buildkite-agent meta-data get "${target}-builder-worker"
 }
+
+install_rustup() {
+  if command -v rustup && command -v cargo &>/dev/null; then
```
jeremymv2: @smacfarlane not a show stopper by any means, however, I'm curious what your thoughts are if we advertised these new functions up one directory as `./expeditor/scripts/shared.sh`? It just occurs to me that they might not be specific enough to be residing under `./expeditor/scripts/post_habitat_release/`.

smacfarlane: I went back on forth on that when I was copying the functions over from habitat-sh/habitat, but landed on this since they're not currently used anywhere else. I don't have really strong feelings either way though.

eeyun: Feels like an easy enough change to make should we want to. I'm :+1: on merging as is and coming back to juggle should we feel like we need to.

jeremymv2: Sure, let's move it later if needed.

PR #1321, opened 2020-02-25
Jamie/1292 member depart origin

SEAjamieD:
This branch allows a user to depart an origin which they are a member of, as long as they are not the owner of that origin.

To Test (Thanks @apriofrost !):

1. Navigate to the members tab of a public origin
2. If the current user is the owner of the origin, show that there is no Depart from Origin button.
3. If the current user is not the owner of the origin, show that there is a Depart from Origin button.
4. As a member (not the owner), click on the "Depart from Origin" button, a confirmation modal shows up as an overlay.
5. On the confirmation modal, click the "Cancel" button, the overlay disappears.
6. Click on the "Depart from Origin" button again. On the confirmation modal, click the Depart from Origin button.
7. See a toast notification pops up from the bottom of the window with two types of messages: successful and failure, with its corresponding body content described in the Acceptance Criteria.
8. Upon success, the user is presented with the public view of the departed origin with the packages tab selected.
9. repeat the same steps for a private origin. In step 8 shows that the user is navigated to the landing page of My Origins (https://bldr.habitat.sh/#/origins).

### DIFF
--- a/components/builder-web/app/origin/origin-page/origin-members-tab/origin-members-tab.component.ts
+++ b/components/builder-web/app/origin/origin-page/origin-members-tab/origin-members-tab.component.ts
@@ -97,8 +97,8 @@ export class OriginMembersTabComponent implements OnInit, OnDestroy {
     return this.store.getState().session.token;
   }
 
-  private get self() {
-    return this.store.getState().users.current.profile.name;
+  private get isPrivate() {
+    return this.store.getState().origins.current.default_package_visibility === 'private' ? true : false;
   }
 
   canDelete(member) {
@@ -143,7 +143,11 @@ export class OriginMembersTabComponent implements OnInit, OnDestroy {
         if (confirmed) {
           this.store.dispatch(departOrigin(this.origin.name, this.token, (originName) => {
             this.router.navigateByUrl('/', { skipLocationChange: true }).then(() => {
-              this.router.navigate(['/origins', originName]);
+              if (this.isPrivate) {
+                this.router.navigate(['/origins']);
+              } else {
+                this.router.navigate(['/origins', originName]);
+              }
             });
           }));
         }
--- a/components/builder-web/app/origin/origin-page/origin-members-tab/origin-members-tab.component.ts
+++ b/components/builder-web/app/origin/origin-page/origin-members-tab/origin-members-tab.component.ts
@@ -73,6 +73,10 @@ export class OriginMembersTabComponent implements OnInit, OnDestroy {
     return this.store.getState().origins.currentPendingInvitations;
   }
 
+  get isOriginOwner() {
+    return this.store.getState().users.current.profile.id === this.store.getState().origins.current.owner_id;
+  }
+
   get members(): List<Object> {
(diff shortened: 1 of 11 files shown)


SEAjamieD:
@apriofrost  As far as UX is concerned - this ticket is ready for review.  I have some more tests to write but there is no more functional or visual code that will be added unless requested.  So I feel like you can get started on this review at your leisure.

components/builder-web/app/origin/origin-page/origin-members-tab/origin-members-tab.component.ts
```diff
@@ -73,6 +77,10 @@ export class OriginMembersTabComponent implements OnInit, OnDestroy {
     return this.store.getState().origins.currentPendingInvitations;
   }
 
+  get isOriginOwner() {
+    return this.store.getState().users.current.profile.id === this.store.getState().origins.current.owner_id;
+  }
+
```
SEAjamieD: This is the logic i'm using to decide whether or not to show the depart from origin button.

components/builder-web/app/origin/origin-page/origin-members-tab/origin-members-tab.component.ts
```diff
@@ -89,6 +97,10 @@ export class OriginMembersTabComponent implements OnInit, OnDestroy {
     return this.store.getState().session.token;
   }
 
+  private get isPrivate() {
+    return this.store.getState().origins.current.default_package_visibility === 'private' ? true : false;
+  }
```
SEAjamieD: explicitly used in the callback after a user departs to determine where they should be routed.

components/builder-web/app/origin/origin-page/origin-members-tab/origin-members-tab.component.ts
```diff
@@ -119,13 +131,41 @@ export class OriginMembersTabComponent implements OnInit, OnDestroy {
...
+  private departureRouting(isPrivate: boolean, originName: string): void {
+    // Check to see if the origin has default package visibility of private and route accordingly
+    if (isPrivate) {
+      this.router.navigateByUrl('/origins');
+    } else {
+      this.router.navigateByUrl('/', { skipLocationChange: true }).then(() => {
+        this.router.navigate(['/origins', originName]);
+      });
+    }
+  }
```
SEAjamieD: I chose to pull the logic for the routing into its own function so it would be easier to follow.  

`this.router.navigateByUrl` causes a page refresh, so when the origin is private, we'll cause a refresh and navigate explicitly to the `/origins` page.

When an origin is public, we'll navigateByUrl to `/` only to force a refresh, and then navigate to `/origins/<origin name>`, right back where we were before but with a different view since now we are no longer a member.

SEAjamieD:
In the interest of time - I'm going to open a separate ticket to add more testing to this when I have the opportunity.  The delay is a result of me needing to learn better testing methods - not with the code itself.  Flipping this over to Ready

eeyun:
Merging since we don't support running builder on windows DESPITE the fact that the failure here is in our CI and not the submitted code.

PR #1323, opened 2020-02-27
Ma/graph utils

markan:
This is a substantial rework of the builder graph utility. 

Remaining work:
* There's more refactoring to do, which will happen as we develop this to generate build orders
* This needs to be backfitted into the builder graph.

### DIFF
--- a/components/builder-graph/src/main.rs
+++ b/components/builder-graph/src/main.rs
@@ -45,8 +45,8 @@ use std::{collections::HashMap,
           fs::File,
           io::Write,
           iter::FromIterator,
-          time::Instant,
-          str::FromStr};
+          str::FromStr,
+          time::Instant};
 
 use clap::{App,
            Arg};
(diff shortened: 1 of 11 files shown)


components/builder-graph/src/main.rs
```diff
@@ -344,8 +504,9 @@ fn do_check(datastore: &DataStore, graph: &PackageGraph, name: &str, filter: &st
     let mut deps_map = HashMap::new();
     let mut new_deps = Vec::new();
     let ident = resolve_name(graph, name);
+    let target = "x86_64-linux";
```
eeyun: Is this intentionally hardcoded to linux?

components/builder-graph/src/main.rs
```diff
@@ -375,11 +536,14 @@ fn do_check(datastore: &DataStore, graph: &PackageGraph, name: &str, filter: &st
...
-                 ident: &str,
-                 filter: &str) {
-    match datastore.get_job_graph_package(ident) {
+fn check_package(
+    datastore: &DataStore,
+    deps_map: &mut HashMap<String, String>,
+    ident: &str,
+    filter: &str,
+) {
+    let target = "x86_64-linux";
```
eeyun: same question as above

components/builder-graph/src/package_graph.rs
```diff
@@ -0,0 +1,795 @@
...
+    str::FromStr,
+    string::ToString,
+};
+
+use petgraph::{
+    algo::{connected_components, is_cyclic_directed},
+    dot::{Config, Dot},
+    graph::EdgeIndex,
+    graph::NodeIndex,
+    //    visit::Dfs,
```
eeyun: Are these comments necessary?

components/builder-graph/src/package_graph.rs
```diff
@@ -0,0 +1,795 @@
...
+            }
+
+            // early out; manifests can be large...
+            if found_deps && found_bdeps {
+                break;
+            }
+        }
+
+        if !(found_deps && found_bdeps) {
+            // Every package should have deps; if not we need to recheck our assumptions
```
eeyun: I wonder about this. We do have the ability for users to package software using their underlying system deps (though it is not encouraged and to my knowledge we've refrained from heavily documenting the behavior.

components/builder-graph/src/package_graph.rs
```diff
@@ -0,0 +1,795 @@
...
+
+    // Possible refactor would be to put packageinfo in graph structure; complication is in multigraph situations
+    full_graph: Graph<PackageIndex, EdgeType>,
+    full_graph_node_index_map: HashMap<NodeIndex, PackageIndex>,
+
+    // We build this alongside the full graph
+    latest_graph: IdentGraph<PackageIndex>,
+}
+
+//IntoEdgeReferences! {delegate_impl[]}
```
eeyun: More dead comments

components/builder-graph/src/package_graph.rs
stevendanna: Because the studio provides a base set of tooling, in at least one package we've gotten away without declaring build deps:

https://github.com/habitat-sh/core-plans/blob/master/cacerts/plan.sh#L10-L11

I could imagine other such "data" type packages having no deps. Not sure if that applies here.

markan: Yeah, that was an overly optimistic view of the world. Not every package has deps. There are a few classes this falls into:
1) True 'no deps' packages. core/cacerts is a good example
2) User packages statically built using system dependencies. While not 'pure' habitat packages, this is a supported use case and common in windows.
3) Bad packages... There are a lot of packages that don't list deps in their manifest, but have them in their plan file. Or don't list them in their plan file, but use them some how.

There are a lot of bad packages, but unfortunately there are enough legit no-deps packages that we can't just flag them as bad.

chefsalim:
Overall changes look reasonable and good progress.

PR #1329, opened 2020-03-03
Swap package count to use origin_package_settings table

### LINKED ISSUE #1325: Top level user origins Packages column really means Artifacts
As part of the recent Package settings [revamp](https://github.com/habitat-sh/builder/pull/1288) changes, we should probably consider a follow up with a clarification of what we mean by the "Packages" column of the top-level user's Origins page.

It is in truth, a count of the number of **Artifacts** that exist.
This could lead to user confusion.

Screen Shot 2020-03-02 at 11 38 24 AM

Screen Shot 2020-03-02 at 10 21 08 AM

### LINKED ISSUE #1327: apparent discrepancy in privacy settings display
I may be mistaken by the intent of the UI icons and names, however, this appears to be a bug in that several indicators on the same page are apparently at odds.

Screen Shot 2020-03-02 at 12 53 08 PM

Screen Shot 2020-03-02 at 12 54 18 PM

eeyun:
This changes our origins_with_stats endpoint to look at the pkg settings table. It should be a marginally quicker query but also will take into account when a "package" has been created but no artifact built or plan connected.

Fixes #1325

### DIFF
--- a/components/builder-db/src/migrations/2020-03-03-000001_origins_with_stats/up.sql
+++ b/components/builder-db/src/migrations/2020-03-03-000001_origins_with_stats/up.sql
@@ -1,0 +1,10 @@
+DROP VIEW origins_with_stats;
+
+CREATE OR REPLACE VIEW origins_with_stats AS
+    SELECT o.*, count(DISTINCT ops.name) as package_count
+        FROM origins o
+        LEFT OUTER JOIN origin_package_settings ops ON o.name = ops.origin
+        GROUP BY o.name
+        ORDER BY o.name;
+
+


jeremymv2:
@eeyun I think you may have the incorrect issue referenced. This PR fixes https://github.com/habitat-sh/builder/issues/1325 😃

eeyun:
It could be worth it, assuming that the query response is still alphabetically sorted, otherwise the data will be mixed in the UI. I will test this locally before we proceed.

eeyun:
I've validated that removing the DESC from the order_by call still returns appropriate data in the appropriate order so i've gone ahead and update the PR.

PR #1334, opened 2020-03-04
Replace usage of time crate with std::time

### LINKED ISSUE #1333: Replace Time crate with std::time
The `time` crate has substantially changed between 0.1 and 0.2. This [issue](https://github.com/time-rs/time/issues/197) describes some of the changes. The change that most impacts us is the removal of PreciseTime in favor of Instant. The std::time library provides this functionality so we should convert our usage to use std::time::Instant.

See this [issue](https://github.com/time-rs/time/issues/201) for help in converting format strings.

Related PR https://github.com/habitat-sh/habitat/pull/7499

smacfarlane:
Closes #1333 

The `time` crate deprecated `PreciseTime` in 0.2.0, and recommends using `Instant` instead. Since `std::time` also provides an `Instant` that identical (for our usage, see below) I've elected to use that which removes our dependency on the `time` crate.

The exception to this is https://github.com/habitat-sh/builder/compare/sm/remove-time?expand=1#diff-3a84b65f0cf2ec59d85d0c682a55c35a, where I replaced `time::Duration` with `chrono::Duration`. This is used in job timeout, which calls `Duration::minutes`and is provided by `Chrono`, but not `std::time::Duration`. The chrono implementation provides overflow protection, which we would otherwise have to implement ourselves.

https://docs.rs/time/0.2.7/time/struct.Instant.html


components/builder-db/src/models/package.rs
```diff
@@ -461,30 +461,29 @@ impl Package {
...
+        let start_time = Instant::now();
         let result = origin_packages_with_version_array::table
             .distinct_on((origin_packages_with_version_array::origin, origin_packages_with_version_array::name, origin_packages_with_version_array::target))
             .order(sql::<PackageWithVersionArray>(
                 "origin, name, target, string_to_array(version_array[1],'.')::\
                 numeric[] desc, ident_array[4] desc",
             ))
             .get_results(conn);
-        let end_time = PreciseTime::now();
+        let duration_millis = start_time.elapsed().as_millis();
```
eeyun: These variables have been inconsistently named  through the pr, some are `duration_milli` some are `duration_millis` which, is not functionally a concern but could make searching the code one query longer. Not a merge blocker but my personal preference would be to keep them consistently named.

smacfarlane: :man_facepalming: I'll fix that

PR #1335, opened 2020-03-05
handle uploading keys after building api and worker services

jeremymv2:
This PR handles the scenario where the user enters the studio and first runs `bb` expecting all services to be built, then started. However, they find that `builder-api` and `builder-worker` services are infinitely waiting on keys.

With this change, a `bb` invocation upon first entering a studio will ensure that these two services are functionally working. Additionally, if a user builds either of these two services individually, we will also ensure the keys have been uploaded after the build.

This is a minor but important scenario that wasn't handled in https://github.com/habitat-sh/builder/pull/1314

### DIFF
--- a/.studiorc
+++ b/.studiorc
@@ -91,6 +91,12 @@ _build-builder-component() {
   echo "builder-$component build succeeded"
 
   start-builder "$component"
+  case "${component}" in
+    api | worker)
+      generate_bldr_keys && upload_github_keys;;
+    *)
+      ;;
+  esac
 }
 build-builder-component() { stop-on-failure _build-builder-component "$@"; }
 


jeremymv2:
> I wonder if this might be a supervisor issue, where a file rumor is marked received by a service, but never written to disk because the service shut down before it had a chance to? 

That was my guess as well, so I added a `sleep 5` right after the first [upload](https://github.com/habitat-sh/builder/blob/0abf988ad44af8a1c3a3d738d3158c5ba9a118b1/.studiorc#L151-L152)  occurs from the `start-builder`, then immediately checked for the existence of the keys on disk and they were not yet there.. So rather than add further sleeps and risk tickling race conditions, I went with this route, which just gratuitously uploads the keys when context changes so that we can ensure they exist.

In the real world, the service won't be loaded/unloaded on short notice like this.

christophermaier:
Thanks for the heads-up... I'll try and look into this soon.

PR #1336, opened 2020-03-06
Adds head routes to common endgoints with large response payloads

eeyun:
As the title suggests. Primary functional change is that our internal handlers used by these endpoints now return actual json values rather than strings.

### DIFF
--- a/components/builder-api/src/server/resources/channels.rs
+++ b/components/builder-api/src/server/resources/channels.rs
@@ -74,14 +74,20 @@ impl Channels {
                   web::get().to(get_packages_for_origin_channel))
            .route("/depot/channels/{origin}/{channel}/pkgs/{pkg}",
                   web::get().to(get_packages_for_origin_channel_package))
+           .route("/depot/channels/{origin}/{channel}/pkgs/{pkg}",
+                  web::head().to(get_packages_for_origin_channel_package_head))
            .route("/depot/channels/{origin}/{channel}/pkgs/{pkg}/latest",
                   web::get().to(get_latest_package_for_origin_channel_package))
            .route("/depot/channels/{origin}/{channel}/pkgs/{pkg}/{version}",
                   web::get().to(get_packages_for_origin_channel_package_version))
+           .route("/depot/channels/{origin}/{channel}/pkgs/{pkg}/{version}",
+                  web::head().to(get_packages_for_origin_channel_package_version_head))
            .route("/depot/channels/{origin}/{channel}/pkgs/{pkg}/{version}/latest",
                   web::get().to(get_latest_package_for_origin_channel_package_version))
            .route("/depot/channels/{origin}/{channel}/pkgs/{pkg}/{version}/{release}",
                   web::get().to(get_package_fully_qualified))
+           .route("/depot/channels/{origin}/{channel}/pkgs/{pkg}/{version}/{release}",
+                  web::head().to(get_package_fully_qualified_head))
            .route("/depot/channels/{origin}/{channel}/pkgs/promote",
                   web::put().to(promote_channel_packages))
            .route("/depot/channels/{origin}/{channel}/pkgs/demote",
@@ -554,6 +560,36 @@ fn get_packages_for_origin_channel_package_version(req: HttpRequest,
 }
 
 #[allow(clippy::needless_pass_by_value)]
+fn get_packages_for_origin_channel_package_version_head(req: HttpRequest,
+                                                        path: Path<(String,
+                                                              String,
+                                                              String,
+                                                              String)>,
+                                                        pagination: Query<Pagination>)
+                                                        -> HttpResponse {
+    let (origin, channel, pkg, version) = path.into_inner();
+    let channel = ChannelIdent::from(channel);
+
+    let ident = PackageIdent::new(origin, pkg, Some(version), None);
+
+    match do_get_channel_packages(&req, &pagination, &ident, &channel) {
+        // If a count of 0 pkgs are returned, return not found
+        Ok((_packages, 0)) => HttpResponse::new(StatusCode::NOT_FOUND),
+        Ok((_packages, _count)) => {
+            HttpResponse::Ok().header(http::header::CONTENT_TYPE, headers::APPLICATION_JSON)
+                              .header(http::header::CACHE_CONTROL,
+                                      headers::Cache::NoCache.to_string())
+                              .finish()
+        }
+        Err(Error::NotFound) => HttpResponse::new(StatusCode::NOT_FOUND),
+        Err(err) => {
+            debug!("Failed to get packages, err={}", err);
+            err.into()
+        }
+    }
+}
+
+#[allow(clippy::needless_pass_by_value)]
 fn get_packages_for_origin_channel_package(req: HttpRequest,
                                            path: Path<(String, String, String)>,
                                            pagination: Query<Pagination>)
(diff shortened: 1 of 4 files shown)


components/builder-api/src/server/resources/channels.rs
```diff
@@ -764,7 +829,7 @@ fn do_get_channel_package(req: &HttpRequest,
                         return Err(Error::SerdeJson(e));
                     }
                 };
-                return Ok(pkg_json);
+                return Ok(serde_json::from_str(&pkg_json).unwrap());
```
smacfarlane: Could we return Ok(_p) here?  Or possibly move the return into the match?

components/builder-api/src/server/resources/channels.rs
```diff
@@ -553,6 +559,34 @@ fn get_packages_for_origin_channel_package_version(req: HttpRequest,
...
+                                                              String,
+                                                              String)>,
+                                                        pagination: Query<Pagination>)
+                                                        -> HttpResponse {
+    let (origin, channel, pkg, version) = path.into_inner();
+    let channel = ChannelIdent::from(channel);
+
+    let ident = PackageIdent::new(origin, pkg, Some(version), None);
+
+    match do_get_channel_packages(&req, &pagination, &ident, &channel) {
```
smacfarlane: In my local testing, this would always return 200 OK, even for packages that don't exist.  This is the behavior of the non-head request, so it's probably correct, but it does seem odd to me given the `Err(Error::NotFound)` branch.

markan: Huh. I get the same thing. It does seem like here we want to return NOT_FOUND if _count == 0, but that would different from the regular GET. Maybe this should be left alone.

eeyun: I am not seeing that at all:
```
┌─╼[~/development/github/builder] [eeyun/head_routes] 
└────╼ curl --head http://localhost/v1/depot/channels/core/unstable/pkgs/bash/4.4.19/20190115012619
HTTP/1.1 200 OK
Server: nginx/1.17.8
Date: Fri, 13 Mar 2020 15:40:11 GMT
Content-Type: application/json
Content-Length: 0
Connection: keep-alive
cache-control: private, no-cache, no-store
X-Frame-Options: SAMEORIGIN
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Alternate-Protocol: 443:npn-spdy/3
X-UA-Compatible: IE=Edge
Strict-Transport-Security: max-age=31536000; includeSubDomains;

┌─╼[~/development/github/builder] [eeyun/head_routes] 
└────╼ curl --head http://localhost/v1/depot/channels/core/unstable/pkgs/bash/4.4.19/20190115012123
HTTP/1.1 404 Not Found
Server: nginx/1.17.8
Date: Fri, 13 Mar 2020 15:40:17 GMT
Content-Length: 0
Connection: keep-alive

Comment (eeyun) [reply components/builder-api/src/server/resources/channels.rs]: Ah it looks like a bug with a couple of the routes, will take a closer look.

Comment (eeyun) [reply components/builder-api/src/server/resources/channels.rs]: Ok, so the cause of this - is that our `do_get_channel_packages` function returns a tuple of packages and a count of packages. In the case of the standard GET this gets filtered through the `postprocess_channel_package_list` function which constructs a response based on the count. With the GET routes apparently we don't return a 404 if no packages from the request fail which I think is behavior we want to preserve. However the concern for the custom HEAD calls is a bit different. I think in this case we DO want to return a 404 if the search returns nothing. Pushing a commit momentarily to correct this.

File: components/builder-api/src/server/resources/channels.rs [hunk 3]
```diff
@@ -553,6 +559,38 @@ fn get_packages_for_origin_channel_package_version(req: HttpRequest,
...
+        Ok((_packages, count)) => {
```
smacfarlane: This is a take it or leave it;  It reads a bit cleaner (to me), but isn't obvious what the `0` is. We could do this in the match instead. Ex: 
```
Ok((_, 0)) => { HttpResponse::new(StatusCode::NOT_FOUND) },
Ok((_packages, _count)) => { 
  HttpResponse::Ok().header(http::header::CONTENT_TYPE, headers::APPLICATION_JSON)
                                  .header(http::header::CACHE_CONTROL,
                                          headers::Cache::NoCache.to_string())
                                  .finish()
},
....
```

eeyun: Oh I like that! Will get this in today.

eeyun:
I can add some, should be straightforward enough

test/builder-api/src/channels.js
```diff
@@ -190,6 +212,17 @@ describe('Channels API', function () {
         });
     });
 
+    it('returns success on HEAD with the specified name, version, and release', function (done) {
+      request.head('/depot/channels/neurosis/foo/pkgs/testapp/0.1.3/20171205003213')
```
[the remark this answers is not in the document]
eeyun: i've added these checks above for the version check as opposed to the release since the release is likely a bit more specific and the routes have the same behaviors,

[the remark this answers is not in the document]
jeremymv2: @eeyun In my local testing the `target` query was only working as expected on the fully qualified package identifier. On the package and version endpoints, it would ALWAYS return 200, even if that package target did not exist.  That's why I suggested a negation test, checking for a 404. 

The only place in hab cli that [check_package](https://github.com/habitat-sh/habitat/blob/master/components/hab/src/command/pkg/upload.rs#L69) is used, it will be a fully qualified ident, with the release.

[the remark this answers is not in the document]
jeremymv2: For example:

```
[10][default:/src:0]# export HAB_ORIGIN=hank
[11][default:/src:0]# export HAB_AUTH_TOKEN=hank
[12][default:/src:0]# curl -XGET -H "Authorization: Bearer $HAB_AUTH_TOKEN" http://localhost/v1/depot/channels/hank/unstable/pkgs/testapp3/0.1.0/20200320152015?target=x86_64-windows
[13][default:/src:0]# curl -I -XGET -H "Authorization: Bearer $HAB_AUTH_TOKEN" http://localhost/v1/depot/channels/hank/unstable/pkgs/testapp3/0.1.0/20200320152015?target=x86_64-windows
HTTP/1.1 404 Not Found
Server: nginx/1.17.8
Date: Mon, 23 Mar 2020 14:06:37 GMT
Content-Length: 0
Connection: keep-alive

[14][default:/src:0]# hab pkg upload -u http://localhost /src/results/hank-testapp3-0.1.0-20200320152015-x86_64-linux.hart
    75 B / 75 B | [==============================================================================================================================================================================] 100.00 % 89.32 KB/s
→ Using existing public origin key hank-20200320151753.pub
→ Using existing hank/testapp3/0.1.0/20200320152015
[15][default:/src:0]# curl -I -XGET -H "Authorization: Bearer $HAB_AUTH_TOKEN" http://localhost/v1/depot/channels/hank/unstable/pkgs/testapp3/0.1.0/20200320152015?target=x86_64-windows
HTTP/1.1 404 Not Found
Server: nginx/1.17.8
Date: Mon, 23 Mar 2020 14:06:58 GMT
Content-Length: 0
Connection: keep-alive

[16][default:/src:0]# curl -I -XGET -H "Authorization: Bearer $HAB_AUTH_TOKEN" http://localhost/v1/depot/channels/hank/unstable/pkgs/testapp3/0.1.0/20200320152015?target=x86_64-linux
HTTP/1.1 200 OK
Server: nginx/1.17.8
Date: Mon, 23 Mar 2020 14:07:02 GMT
Content-Type: application/json
Content-Length: 1765
Connection: keep-alive
cache-control: private, no-cache, no-store
X-Frame-Options: SAMEORIGIN
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Alternate-Protocol: 443:npn-spdy/3
X-UA-Compatible: IE=Edge
Strict-Transport-Security: max-age=31536000; includeSubDomains;

[17][default:/src:0]# curl -I -XGET -H "Authorization: Bearer $HAB_AUTH_TOKEN" http://localhost/v1/depot/channels/hank/unstable/pkgs/testapp3/0.1.0?target=x86_64-linux
HTTP/1.1 200 OK
Server: nginx/1.17.8
Date: Mon, 23 Mar 2020 14:07:11 GMT
Content-Type: application/json
Content-Length: 137
Connection: keep-alive
cache-control: private, no-cache, no-store
X-Frame-Options: SAMEORIGIN
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Alternate-Protocol: 443:npn-spdy/3
X-UA-Compatible: IE=Edge
Strict-Transport-Security: max-age=31536000; includeSubDomains;

[18][default:/src:0]# curl -I -XGET -H "Authorization: Bearer $HAB_AUTH_TOKEN" http://localhost/v1/depot/channels/hank/unstable/pkgs/testapp3/0.1.0?target=x86_64-windows
HTTP/1.1 200 OK
Server: nginx/1.17.8
Date: Mon, 23 Mar 2020 14:07:23 GMT
Content-Type: application/json
Content-Length: 137
Connection: keep-alive
cache-control: private, no-cache, no-store
X-Frame-Options: SAMEORIGIN
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Alternate-Protocol: 443:npn-spdy/3
X-UA-Compatible: IE=Edge
Strict-Transport-Security: max-age=31536000; includeSubDomains;

[19][default:/src:0]# curl -I -XHEAD -H "Authorization: Bearer $HAB_AUTH_TOKEN" http://localhost/v1/depot/channels/hank/unstable/pkgs/testapp3/0.1.0?target=x86_64-windows
HTTP/1.1 200 OK
Server: nginx/1.17.8
Date: Mon, 23 Mar 2020 14:07:34 GMT
Content-Type: application/json
Content-Length: 0
Connection: keep-alive
cache-control: private, no-cache, no-store
X-Frame-Options: SAMEORIGIN
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Alternate-Protocol: 443:npn-spdy/3
X-UA-Compatible: IE=Edge
Strict-Transport-Security: max-age=31536000; includeSubDomains;

[20][default:/src:0]# curl -I -XHEAD -H "Authorization: Bearer $HAB_AUTH_TOKEN" http://localhost/v1/depot/channels/hank/unstable/pkgs/testapp3/0.1.0/20200320152015?target=x86_64-windows
HTTP/1.1 404 Not Found
Server: nginx/1.17.8
Date: Mon, 23 Mar 2020 14:07:45 GMT
Content-Length: 0
Connection: keep-alive

[21][default:/src:0]#
```

The only response that seems to be behaving as expected, when passed the target, is the fully qualified release - which happens to be the only one currently in use by the hab client when it checks on an `upload`. 🤔

PR #1350, opened 2020-03-12
install_base_packages.sh supports all linux platform targets

jeremymv2:
This PR unifies our Builder bootstrap process, consolidating everything into the `install_base_packages.sh` which now takes optional arguments for platform type and installs from the bootstrap bundle containing .harts for all target types.


jeremymv2:
@christophermaier 
> Out of curiosity, how do the Windows workers get provisioned? 

Windows workers are provisioned via userdata in the form of powershell here: https://github.com/habitat-sh/builder/blob/master/terraform/templates/windows_worker_user_data.tpl#L28-L34

PR #1354, opened 2020-03-13
Ma/graph cli

markan:
The number and variations of the commands has grown substantially. I
got tired of manual parsing so changed it to use clap for the internal
command.  There's a minor fix remaining to get help output working,
because we can't distinguish a help request from a bad arg, but this
already is a much more useful interface IMHO.

### DIFF
--- a/components/builder-graph/src/main.rs
+++ b/components/builder-graph/src/main.rs
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Chef Software Inc. and/or applicable contributors
+// Copyright (c) 2017-2020 Chef Software Inc. and/or applicable contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@
 
 #[macro_use]
 extern crate bitflags;
+use clap::clap_app;
 #[macro_use]
 extern crate features;
 #[macro_use]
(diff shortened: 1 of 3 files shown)


components/builder-graph/src/main.rs
```diff
@@ -18,6 +18,8 @@
 #[macro_use]
 extern crate bitflags;
 #[macro_use]
+extern crate clap;
```
smacfarlane: We should use the Rust 2018 macro import syntax here: 

https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html

Specifically we can drop the `#[macro_use]; extern crate clap;` and add `clap_app` to the `use clap::` below;

The remainder of the `#[macro_use]` instance can be cleaned up similarly, but as it's a bit of a yak shave that might be good to make its own PR.

components/builder-graph/src/main.rs
```diff
@@ -550,16 +524,17 @@ fn do_export(graph: &PackageGraph, filename: &str, filter: &str) {
...
+        if ident.starts_with(&filter) {
             file.write_fmt(format_args!("{}\n", ident)).unwrap();
         }
     }
 }
 
-fn do_target(graph: &mut PackageGraph, target: &str) {
-    match PackageTarget::from_str(target) {
+fn do_target(graph: &mut PackageGraph, matches: &ArgMatches) {
+    println!("ClapDTM {:?}", matches);
```
smacfarlane: I see a few of these `ClapX` `println!` statements.  Could we expand on what the `X` is, in this case `DTM`?  We also may want to consider moveing these to `debug!` lines before merging.

PR #1357, opened 2020-03-18
Add testing steps for core-plans refresh

smacfarlane:
This adds steps to the DEVELOPING document for building against core origin packages that have not been promoted to stable yet. 

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -280,7 +280,7 @@ In some scenarios, it's valuable to test against `core` packages that haven't be
 
 #### Build Habitat components
 
-First, you will need to clone https://github.com/habitat-sh/habitat and build a subset of the components. It is important they are built in the correct order so that dependencies are correct at install time. You can use the blow snippet to build them, replacing the channel as necessary.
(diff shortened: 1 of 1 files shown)


DEVELOPING.md
```diff
@@ -274,6 +274,38 @@ To view the DEBUG level logs from the API tests:
...
+hab pkg install results/core-hab*.hart
+for component in builder-api builder-api-proxy builder-datastore builder-graph builder-jobsrv builder-minio builder-worker; do
+  build components/$component
+done
+hab sup term
+HAB_FUNC_TEST=1 sup-run
+start-builder
+# wait for services to come up
+test/builder-api/test.sh
+```
```
jeremymv2: This looks great. Just a quick question: Is this the same thing that the `test-builder` alias provides?
https://github.com/habitat-sh/builder/blob/master/.studiorc#L291-L334
If so, would it make sense to use the alias?

smacfarlane: I forgot that existed when I was writing :man_facepalming:   I'll make that update.

PR #1378, opened 2020-04-02
Lightweight RBAC

### LINKED ISSUE #1149: Have Read-Only / Key-Only membership to an Origin
### Describe the Enhancement:
Currently, there are only two levels of membership to an origin, an owner and a member. Both have access to the origin signing key as well as the ability to upload packages to the origin. By adding a limited access origin member this would allow a user to be able to use their current auth token to pull private packages from the origin without having access to upload or create new packages for that origin. 
### Describe the Need:

This would allow people to publish packages and grant users the ability to install them without giving them full access to the origin. In my opinion, this is the critical step to premium habitat content.

### Current Alternative
Currently, the only real alternative is to make the packages public. If we are talking about SaaS Builder and a public package is not an option then they need to stand up an on-premise depot and then limit access through OAuth. This causes an additional maintenance requirement for the customer.

### Can We Help You Implement This?:
Yes, I am currently looking at what it will take and have talked with @eeyun and @smacfarlane about the database. Currently, the first iteration could be adding an additional boolean field to the `origin_members` table called `key_access`. This would require a database migration that would update this field for all current members to true. This would probably need to start as an API where the default for a new invitation is `false` then through an API post-call like to `/origin/users/{username}/key_access/` to grant a user key access. There may be some UI changes to remove the keys tab from view for the user or it could be as simple as returning a grayed out warning like `Sorry, you currently do not have access to the {origin} keys please contact the owner of this origin to request access`.

eeyun:
Closes #1149 

I'd like to request a plethora of super dank gifs on this one.

This PR adds the first iteration of the new lightweight RBAC in builder. It preserves current behavior for all current users and provides new routes for altering a user's role. The only fully implemented roles in this PR is the new 'member' which functions as a read only role. User's with the 'maintainer' role will have all of the permissions they've been used to up to this point. We will be adding a follow-on PR to include some kind of doc showing how to use curl to alter a role until the front-end and cli accompanying work lands.

We've preserved the commit history in the hopes that it makes reviewing the PR a little easier. While the changes aren't wildly complex, they are broad.


jeremymv2:
bossboat

smacfarlane:
access denied

components/builder-api/src/server/resources/origins.rs
```diff
@@ -1158,14 +1202,121 @@ fn list_origin_invitations(req: HttpRequest,
...
+
 #[allow(clippy::needless_pass_by_value)]
 fn transfer_origin_ownership(req: HttpRequest,
                              path: Path<(String, String)>,
                              state: Data<AppState>)
                              -> HttpResponse {
     let (origin, user) = path.into_inner();
 
-    let session = match authorize_session(&req, None) {
+    let session = match authorize_session(&req, None, RoleAction::Write, None) {
```
smacfarlane, line 1323: Should this require the owner role?

components/builder-api/src/server/resources/origins.rs
```diff
@@ -1221,7 +1372,7 @@ fn transfer_origin_ownership(req: HttpRequest,
 fn depart_from_origin(req: HttpRequest, path: Path<String>, state: Data<AppState>) -> HttpResponse {
     let origin = path.into_inner();
 
-    let session = match authorize_session(&req, None) {
+    let session = match authorize_session(&req, None, RoleAction::Read, None) {
```
smacfarlane, line 1379: Should this require write + membership?

components/builder-api/src/server/resources/origins.rs
```diff
@@ -1261,7 +1412,7 @@ fn list_origin_members(req: HttpRequest,
                        -> HttpResponse {
     let origin = path.into_inner();
 
-    if let Err(err) = authorize_session(&req, Some(&origin)) {
+    if let Err(err) = authorize_session(&req, Some(&origin), RoleAction::Read, None) {
```
smacfarlane, line 1419: Should listing require membership?

components/builder-api/src/server/resources/origins.rs
```diff
@@ -1331,7 +1486,7 @@ fn fetch_origin_integrations(req: HttpRequest,
                              -> HttpResponse {
     let origin = path.into_inner();
 
-    if let Err(err) = authorize_session(&req, Some(&origin)) {
+    if let Err(err) = authorize_session(&req, Some(&origin), RoleAction::Read, None) {
```
smacfarlane, line 1493: Should this require membership?

components/builder-api/src/server/resources/origins.rs
```diff
@@ -1366,7 +1521,7 @@ fn fetch_origin_integration_names(req: HttpRequest,
                                   -> HttpResponse {
     let (origin, integration) = path.into_inner();
 
-    if let Err(err) = authorize_session(&req, Some(&origin)) {
+    if let Err(err) = authorize_session(&req, Some(&origin), RoleAction::Read, None) {
```
smacfarlane, line 1528: Should this require membership?

components/builder-api/src/server/resources/origins.rs
```diff
@@ -1466,7 +1629,7 @@ fn get_origin_integration(req: HttpRequest,
                           -> HttpResponse {
     let (origin, integration, name) = path.into_inner();
 
-    if let Err(err) = authorize_session(&req, Some(&origin)) {
+    if let Err(err) = authorize_session(&req, Some(&origin), RoleAction::Read, None) {
```
smacfarlane, line 1636: Should this require membership?

components/builder-api/src/server/resources/pkgs.rs
```diff
@@ -1142,7 +1158,7 @@ fn do_upload_package_finish(req: &HttpRequest,
                                        Body::from_message("ds:up:6"));
     }
 
-    let session = authorize_session(&req, None).unwrap(); // Unwrap Ok
+    let session = authorize_session(&req, None, RoleAction::Write, None).unwrap(); // Unwrap Ok
```
smacfarlane, line 1161: Should this require maintainer role?

components/builder-api/src/server/resources/profile.rs
```diff
@@ -210,7 +211,7 @@ fn update_account(req: HttpRequest,
                   body: Json<UserUpdateReq>,
                   state: Data<AppState>)
                   -> HttpResponse {
-    let account_id = match authorize_session(&req, None) {
+    let account_id = match authorize_session(&req, None, RoleAction::Read, None) {
```
smacfarlane, line 214: Should this be `Write`?

components/builder-api/src/server/resources/profile.rs
```diff
@@ -172,7 +173,7 @@ fn revoke_access_token(req: HttpRequest,
         Err(_) => return HttpResponse::new(StatusCode::UNPROCESSABLE_ENTITY),
     };
 
-    let account_id = match authorize_session(&req, None) {
+    let account_id = match authorize_session(&req, None, RoleAction::Read, None) {
```
smacfarlane, line 176: Should this be `Write`?

components/builder-api/src/server/resources/projects.rs
```diff
@@ -236,7 +241,7 @@ fn get_project(req: HttpRequest,
                -> HttpResponse {
     let (origin, name) = path.into_inner();
 
-    if let Err(err) = authorize_session(&req, Some(&origin)) {
+    if let Err(err) = authorize_session(&req, Some(&origin), RoleAction::Read, None) {
```
smacfarlane, line 244: Should this require membership?

components/builder-api/src/server/resources/projects.rs
```diff
@@ -424,7 +437,7 @@ fn update_project(req: HttpRequest,
 fn get_projects(req: HttpRequest, path: Path<String>, state: Data<AppState>) -> HttpResponse {
     let origin = path.into_inner();
 
-    if let Err(err) = authorize_session(&req, Some(&origin)) {
+    if let Err(err) = authorize_session(&req, Some(&origin), RoleAction::Read, None) {
```
smacfarlane, line 440: Should this require membership?

components/builder-api/src/server/resources/projects.rs
```diff
@@ -455,7 +468,7 @@ fn get_jobs(req: HttpRequest,
             -> HttpResponse {
     let (origin, name) = path.into_inner();
 
-    if let Err(err) = authorize_session(&req, Some(&origin)) {
+    if let Err(err) = authorize_session(&req, Some(&origin), RoleAction::Read, None) {
```
smacfarlane, line 471: Should this require membership?

components/builder-api/src/server/resources/projects.rs
```diff
@@ -585,7 +606,7 @@ fn get_integration(req: HttpRequest,
                    -> HttpResponse {
     let (origin, name, integration) = path.into_inner();
 
-    if let Err(err) = authorize_session(&req, Some(&origin)) {
+    if let Err(err) = authorize_session(&req, Some(&origin), RoleAction::Read, None) {
```
smacfarlane, line 609: Should this require membership?

components/builder-api/src/server/resources/settings.rs
```diff
@@ -56,7 +57,7 @@ impl Settings {
 fn get_origin_package_settings(req: HttpRequest, path: Path<(String, String)>) -> HttpResponse {
     let (origin, pkg) = path.into_inner();
 
-    if let Err(err) = authorize_session(&req, Some(&origin)) {
+    if let Err(err) = authorize_session(&req, Some(&origin), RoleAction::Read, None) {
```
smacfarlane, line 60: Should this require membership?

components/builder-api/src/server/resources/origins.rs
```diff
@@ -1095,7 +1139,7 @@ fn rescind_invitation(req: HttpRequest,
                       -> HttpResponse {
     let (origin, invitation) = path.into_inner();
 
-    let _ = match authorize_session(&req, None) {
+    let _ = match authorize_session(&req, None, RoleAction::Write, None) {
```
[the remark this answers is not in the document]
eeyun: It does. Any authorize call that includes a `RoleAction::Write` requires maintainer role

components/builder-api/src/server/resources/origins.rs
```diff
@@ -1132,7 +1176,7 @@ fn list_origin_invitations(req: HttpRequest,
                            -> HttpResponse {
     let origin = path.into_inner();
 
-    if let Err(err) = authorize_session(&req, Some(&origin)) {
+    if let Err(err) = authorize_session(&req, Some(&origin), RoleAction::Read, None) {
```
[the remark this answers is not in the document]
eeyun: No, in this there's no reason a member shouldnt be able to see active invitations and due to users potentially needing to see any invitation that might have been sent to them this should be read.

components/builder-api/src/server/resources/origins.rs
eeyun: This does require owner role - all of the previous deep guards for authorization have remained in place.  This is an early exit which should help with the need to check if the requester is the owner which happens on 1324

components/builder-api/src/server/resources/origins.rs
eeyun: Nope. In this case anyone should be able to leave an origin of their own volition.

components/builder-api/src/server/resources/origins.rs
eeyun: This does require membership. Authorize session still check's if you're a member as it did before. The new arguments are for verifying your level of access.

components/builder-api/src/server/resources/origins.rs
eeyun: Same as above

components/builder-api/src/server/resources/origins.rs
eeyun: Same as above

components/builder-api/src/server/resources/origins.rs
eeyun: Same as above

components/builder-api/src/server/resources/pkgs.rs
eeyun: Same as above maintainer question. Write functions require maintainer or higher.

components/builder-api/src/server/resources/profile.rs
eeyun: No in this case, a user needs to be able to update their own account. If we alter this to `::Write` they would be unable to do so if they have the role of 'member'

components/builder-api/src/server/resources/profile.rs
eeyun: No similar to above. A user needs to be able to revoke their own token.

components/builder-api/src/server/resources/projects.rs
eeyun: Same as above

components/builder-api/src/server/resources/projects.rs
eeyun: same as above

components/builder-api/src/server/resources/projects.rs
eeyun: same as above

components/builder-api/src/server/resources/projects.rs
eeyun: same as above

components/builder-api/src/server/resources/settings.rs
eeyun: same as above

components/builder-api/src/server/authorize.rs
```diff
@@ -23,56 +24,111 @@ use crate::server::{error::{Error,
...
+        write!(f, "{}", value)
+    }
+}
+
+#[allow(clippy::needless_pass_by_value)]
 pub fn authorize_session(req: &HttpRequest,
-                         origin_opt: Option<&str>)
+                         origin_opt: Option<&str>,
+                         action: RoleAction,
+                         min_role: Option<OriginMemberRole>)
```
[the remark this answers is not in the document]
eeyun: This is somewhat confusing for sure.

To break it down. `RoleAction::` is passed to the authz function to determine whether or not you're inherently blocked from using the resource. `min_role` gives us the ability evaluate different roles on different resources. Its added complexity but it gives us a lot more flexibility in the way we can determine permissions which is going to be necessary when the rest of the roles are implemented.

components/builder-api/src/server/resources/pkgs.rs
smacfarlane: :thinking:  Personal preference here, but it feels like it's better to be explicit on requiring maintainer access.  Otherwise wouldn't be be relying on implementation details of `authorize_session` to guarantee that we've set levels correctly?   This may be related to my lack of understanding around `action` vs `min_role` though.

components/builder-api/src/server/resources/origins.rs
```diff
@@ -1158,14 +1202,121 @@ fn list_origin_invitations(req: HttpRequest,
...
+    // The account id of the user being requested
+    let (target_user_id, _) = match Account::get(&username, &*conn).map_err(Error::DieselError) {
+        Ok(account) => (account.id, account.name),
+        Err(err) => {
+            debug!("{}", err);
+            return err.into();
+        }
+    };
+
+    match OriginMember::member_role(&origin, target_user_id, &*conn) {
```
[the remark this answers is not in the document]
eeyun: I'm not sure I follow. In the case of this function specifically you should not be able to get the role of a member if you are also not a member of the origin. Does that answer your question?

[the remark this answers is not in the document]
smacfarlane: My question is more "Can I, as a member of origin O, request the role of Person in O, when Person is not a member of O" or in more concrete terms;  if we removed @smacfarlane  from the `habitat` origin, could you then make a query "What is @smacfarlane role in the `habitat` origin?".  

I'm not sure I see that case handled, unless it's in the `Err` arm of the `member_role` query. This may just be my unfamiliarity with Diesel.

components/builder-api/src/server/authorize.rs
```diff
@@ -23,56 +24,111 @@ use crate::server::{error::{Error,
...
-        }
-
+        };
         match check_origin_member(req, origin, session.get_id()) {
-            Ok(is_member) => {
-                memcache.set_origin_member(origin, session.get_id(), is_member);
-
-                if !is_member {
-                    return Err(Error::Authorization);
+            Ok(false) => {
```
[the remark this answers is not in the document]
eeyun: We should take a look at this, but definitely it should come through in a follow up pr.

components/builder-api/src/server/resources/pkgs.rs
eeyun: In most cases I definitely agree, in this case, this isn't the first authorization call in this function call - its a holdover from the previous behavior where we would call this function later on in the function flow to get access to session data. Same thing goes here. We've preserved the role action because it gives us the quick validation without making the minimum role comparison. I don't know if it's actually any faster but its one fewer operation.

components/builder-api/src/server/resources/origins.rs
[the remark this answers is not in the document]
eeyun: Correct, they cannot. In this case all auth is keyed to be origin specific as it was before. So in this case if you are not a member of an origin you cannot read the roles of users in that origin. I've just added a test case for this so that this is explicit!

components/builder-api/src/server/authorize.rs
```diff
@@ -23,56 +24,111 @@ use crate::server::{error::{Error,
...
+        };
+        write!(f, "{}", value)
+    }
+}
+
+#[allow(clippy::needless_pass_by_value)]
 pub fn authorize_session(req: &HttpRequest,
-                         origin_opt: Option<&str>)
+                         origin_opt: Option<&str>,
+                         action: RoleAction,
```
markan, line 45: So it occurs to me that the action here is actually not very useful. The action is pretty much predetermined by the caller context (e.g. ('repo_file_content') and that's defined to be a read or a write in it's very nature. Really all we need is the min_role; read actions can be represented by a lower min_role (e.g something in a write context requires maintainer but read context requires member.) That might simplify some of the logic.

markan: It occurs to me that the action field might have been something I suggested; I certainly hadn't realized how decomposed our contexts were.

eeyun: So the intention here is around future flexibility around tunables and user roles in tandem with what we assume is going to be a middleware for handling authz. Effectively, there is a future state where we might want a user to be able configure resource access based on some arbitrary schema. In which case having the route actions will remain and the minimum role will go away. This just felt like a low impact way of prepping for that future work without obligating a full implementation on the immediate.

components/builder-api/src/server/authorize.rs
```diff
@@ -89,7 +145,55 @@ pub fn check_origin_member(req: &HttpRequest, origin: &str, account_id: u64) ->
     if account_id == BUILDER_ACCOUNT_ID {
         Ok(true)
     } else {
+        let mut memcache = req_state(req).memcache.borrow_mut();
+        match memcache.get_origin_member(origin, account_id) {
+            Some(val) => {
+                debug!("Origin membership {} {} Cache Hit!", origin, account_id);
+                if val {
```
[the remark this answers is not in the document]
jeremymv2: 👍 Handled in https://github.com/habitat-sh/builder/pull/1378/commits/ae1def6298a0dfe399322a4a5798b1601948c714

components/builder-api/src/server/authorize.rs
```diff
@@ -23,56 +24,111 @@ use crate::server::{error::{Error,
...
+                                           session.get_id(),
+                                           RoleAction::Write,
+                                           origin,
+                                           minimum_req_role,
+                                           member_role);
+                                    return Err(Error::Authorization);
+                                }
+                            }
+                            None => {
+                                debug!("authorize_session: account {} is not a member of origin \
```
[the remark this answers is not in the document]
jeremymv2: 👍 I removed this None variant case in https://github.com/habitat-sh/builder/pull/1378/commits/ae1def6298a0dfe399322a4a5798b1601948c714

components/builder-api/src/server/authorize.rs
```diff
@@ -23,56 +24,111 @@ use crate::server::{error::{Error,
...
+                match action {
+                    RoleAction::Read => {
+                        debug!("authorize_session: account {} has {} permissions in origin {}",
+                               session.get_id(),
+                               RoleAction::Read,
+                               origin);
+                        return Ok(session);
+                    }
+                    RoleAction::Write => {
+                        match check_origin_member_role(req, origin, session.get_id()) {
```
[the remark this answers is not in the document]
jeremymv2: I've added a TODO item for the use of memached optimization on `check_origin_member_role()` in https://github.com/habitat-sh/builder/pull/1378/commits/be4830739566ebad7d28d8d9157b103cf23d6a4e.

I think that's a big enough implementation delta that it merits followup as its own separate work item, distinct from the reasonably large set of changes already present in this PR.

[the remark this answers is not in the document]
jeremymv2: With regard to the optimization around combining `check_origin_member` and `check_origin_member_role`, I definitely agree with you that there is some overlap in what the two functions provide. I think that since `check_origin_member` is already a Public function in use by the server API, it would be a larger refactor lift that would need to change function return handling in multiple places.

With that in mind, and knowing that we plan to optimize `check_origin_member_role` by [leveraging memcached](https://github.com/habitat-sh/builder/issues/1384), I think it will be ok to leave these two functions distinct for now and address your suggested changes once we are informed by our optimization with memcached. 

With that in mind, and knowing that we plan to optimize `check_origin_member_role` by [leveraging memcached](https://github.com/habitat-sh/builder/issues/1384), I think it will be ok to leave these two functions distinct for now.

components/builder-api/src/server/authorize.rs
markan: There's a tradeoff between taking complexity today to support future possible needs; you are paying for something you may never use. In this case, I'd argue to keep things simple:
1) I think it will be easy to add this back in if we need it
2) The read/write formulation isn't the only way to describe things (CRUD comes to mind), and we may need to change it anyways.
and probably most importantly:

3) I found it harder to reason about permissions in this current setup; we have two paths (read/write) which behave differently. Coming at it cold it took me a minute to figure out what was intended. I'm leery of complexity in security related code.

jeremymv2: I agree with @markan and removed the RoleAction in https://github.com/habitat-sh/builder/pull/1378/commits/e3a581121316bba3160d4d3a3a41b380d0e28462

components/builder-api/src/server/authorize.rs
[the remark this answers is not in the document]
jeremymv2: @markan and I arrived at the same conclusion over a pairing session which resulted in a replacing the `check_origin_member` call with `check_origin_member_role`. This change is in https://github.com/habitat-sh/builder/pull/1378/commits/e3a581121316bba3160d4d3a3a41b380d0e28462

Issue #1391, opened 2020-04-14
Extend database query time logging for origin_packages and related views


markan:
As our database size grows, a number of queries are becoming non-performant. In particular the origin_packages table and in particular the views packages_with_channel_platform  origin_package_versions 

We have some logging for query time; we should take that and extend it more generally. 

This is limited to selects; we may want to make sure that no insert/update operators are doing anything strange, but by and large they don't do the complex join/filter operations that seem to be our problem.

markan:
This is a sibling task to #1390

markan:
PR #1559 would be useful if doing this work.

PR #1392, opened 2020-04-15
First pass at adding db metrics

markan:
This adds metrics to everything that queries the origin_packages table
in a way that looks likely perform poorly. This addresses issue #1391.

It follows the existing naming convention as best I could. There's a
lot of boilerplate, which seems a little bit hard to remove.

However I'd like to suggest a different convention: let's construct
the name as structname,functionname using the approprate
inflection/casing scheme. E.g. for the Packages::count_origin_packages
functionthe enum names would be PackagesCountOriginPackagesCallTime,
and the logging keys would be
db-call.packages-count-origin-packages-call-time. I find it easier to
figure out, and I believe it will make it easier to group related
statistics.

This potentially breaks existing logs, and I'd suggest we log twice,
once with the old and once with the new.

One advantage of this scheme is that I believe it makes it easier to
write a macro that would reduce the boilerplate substantially.

### DIFF
--- a/components/builder-db/src/metrics.rs
+++ b/components/builder-db/src/metrics.rs
@@ -34,10 +34,24 @@ impl metrics::Metric for Counter {
 
 pub enum Histogram {
     DbCallTime,
+    CountOriginPackagesCallTime,
     GetAllLatestCallTime,
+    GetAllPackageCallTime,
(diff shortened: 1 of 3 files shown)


eeyun:
Looks like we've got some rustfmt failures but otherwise it looks good!

Issue #1384, opened 2020-04-09
insert check_origin_member_role() return value into memcached


jeremymv2:
Similarly to how we use [authorize_session::memcache.set_origin_member](https://github.com/habitat-sh/builder/blob/7d0aba8a6b0d639b7c6143987f85a1cbae4608cf/components/builder-api/src/server/authorize.rs#L66), we should implement a memcache function for saving the result of `authorize_session::check_orgin_member_role()`.

This will make the calls to `authorize_session` more performant by limiting the number of database calls to check a member's origin role.

PR #1394, opened 2020-04-16
Adds origin member role caching on all requests

eeyun:
Closes https://github.com/habitat-sh/builder/issues/1384

As the title suggests, this adds local caching (both addition and invalidation) of all request role data. PR also slightly tweaks the origin_member_role_update function to avoid some bad output we discovered while writing the caching bits and includes a test for that behaviour change. 

Signed-off-by: Jeremy Miller <jm@chef.io>


components/builder-api/src/server/services/memcache.rs
```diff
@@ -206,6 +210,35 @@ impl MemcacheClient {
...
+        let duration_millis = start_time.elapsed().as_millis();
+        trace!("Memcache get_origin_member_role time: {} ms",
+               duration_millis);
+        Histogram::MemcacheCallTime.set(duration_millis as f64);
+
+        ret
+    }
+
+    pub fn set_origin_member_role(&mut self, origin: &str, account_id: u64, role: &str) {
+        let key = format!("member_role:{}/{}", origin, account_id);
```
markan, line 230: Why doesn't this use member_role_ns_key?

components/builder-api/src/server/services/memcache.rs
```diff
@@ -206,6 +210,35 @@ impl MemcacheClient {
         ret
     }
 
+    pub fn get_origin_member_role(&mut self, origin: &str, account_id: u64) -> Option<String> {
+        trace!("Getting origin role membership for {} {} from memcached",
+               origin,
+               account_id);
+
+        let key = format!("member_role:{}/{}", origin, account_id);
```
markan, line 218: Why doesn't this use member_role_ns_key?

eeyun: see above

components/builder-api/src/server/services/memcache.rs
eeyun: From what we can tell the hashing happens internal to memcache and is only necessary when using the API to pull out a value. Submitting a value this is the pattern that salim had already established so this is what we went with.

components/builder-api/src/server/authorize.rs
```diff
@@ -127,24 +131,48 @@ fn check_origin_member_role(req: &HttpRequest,
...
-                Ok(member_role) => {
-                    debug!("Found account {} has member type {}",
-                           account_id, member_role);
-                    Some(member_role)
-                }
-                Err(err) => {
-                    warn!("Unable to determine member type for account {} in origin {}: {}",
-                          account_id, origin, err);
-                    None
+    if account_id == BUILDER_ACCOUNT_ID {
```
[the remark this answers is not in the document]
eeyun: good catch!

components/builder-api/src/server/services/memcache.rs
eeyun: Good catch! I originally added a comment here that was way off base - i had been confused about which function that was. Corrected!

PR #1396, opened 2020-04-20
Ma/optimize package channel queries

markan:
This PR has two parts:

* some optimizations to the SQL indices and views; these largely serve to improve the performance of the users of origin_packages, origin_package_versions, origin_packages_with_version_array, and packages_with_channel_platform, 

* Tweaks to the queries to constrain on origin and name; we weren't really filtering properly with the contains ident_array construct, as it is 1) unordered, and 2) doesn't interact well with the window in the packages_with_channel_platform_array, and so ends up not using indices and doing linear searches instead.

### DIFF
--- a/components/builder-db/src/models/channel.rs
+++ b/components/builder-db/src/models/channel.rs
@@ -194,6 +194,15 @@ impl Channel {
         trace!("DBCall channel::list_package time: {} ms", duration_millis);
         Histogram::DbCallTime.set(duration_millis as f64);
         Histogram::ChannelListPackagesCallTime.set(duration_millis as f64);
+
+        // Package list for a whole origin is still not very
+        // performant, and we want to track that
+        if lcp.ident.name != "" {
+            Histogram::ChannelListPackagesOriginOnlyCallTime.set(duration_millis as f64);
+        } else {
+            Histogram::ChannelListPackagesOriginNameCallTime.set(duration_millis as f64);
+        }
+
         result
     }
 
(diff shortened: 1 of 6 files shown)


eeyun:
Unfortunately this doesn't currently pass our functional tests. It look like some of the altered queries no longer return data.

components/builder-db/src/models/package.rs
```diff
@@ -585,6 +588,8 @@ impl Package {
 
         let (mut pkgs, _): (std::vec::Vec<PackageWithChannelPlatform>, i64) =
             packages_with_channel_platform::table
+                .filter(packages_with_channel_platform::origin.eq(&pl.ident.origin))
+                .filter(packages_with_channel_platform::name.eq(&pl.ident.name))
```
eeyun, line 592: Same as above

components/builder-db/src/models/package.rs
```diff
@@ -621,7 +626,9 @@ impl Package {
             .select(sql(
                 "concat_ws('/', ident_array[1], ident_array[2]) as ident",
             ))
-            .filter(origin_packages::ident_array.contains(pl.ident.parts()))
+            .filter(origin_packages::origin.eq(&pl.ident.origin))
+            .filter(origin_packages::name.eq(&pl.ident.name))
```
eeyun, line 630: This one is the same issue as above

components/builder-db/src/models/channel.rs
```diff
@@ -170,6 +170,8 @@ impl Channel {
                 origin_channel_packages::table
                     .inner_join(origin_channels::table.inner_join(origins::table)),
             )
+            .filter(origin_packages::origin.eq(&lcp.ident.origin))
+            .filter(origin_packages::name.eq(&lcp.ident.name))
```
eeyun, line 174: I know you're still working on this PR so don't be too concerned with this feedback, wanted to give the the cause of the functional test failures so you don't have to hunt them down - `.filter(origin_packages::name.eq(&lcp.ident.name))` - this function gets called from routes that expect to list all packages in an origin. Filtering on a package name here breaks a couple routes.

PR #1397, opened 2020-04-21
accounting for other oauth providers in api run hook

jeremymv2:
We check to see if Jobsrv is enabled in builder-api, only then do we wait on the `builder-github-app.pem` file. 

Noticed this was an issue raised by a user in our forums: https://forums.habitat.sh/t/on-prem-builder-api-port-not-binding/1316

### DIFF
--- a/components/builder-api/habitat/hooks/run
+++ b/components/builder-api/habitat/hooks/run
@@ -12,11 +12,14 @@ echo "_e{${#title},${#text}}:$title|$text|#api"  >/dev/udp/localhost/8125
 # Wait for the .pem file before starting the service, otherwise the service
 # continually panics in a tight loop, blocking the supervisor from writing
 # out the file.
-while ! [ -f "{{pkg.svc_files_path}}/builder-github-app.pem" ];
-do
-    echo "Waiting for builder-github-app.pem"
+# shellcheck disable=SC2050
+if [ "{{cfg.oauth.provider}}" == "github" ]; then
+  while ! [ -f "{{pkg.svc_files_path}}/builder-github-app.pem" ];
+  do
+    echo "Waiting for builder github app private key at: {{pkg.svc_files_path}}/builder-github-app.pem"
     sleep 10
-done
+  done
+fi
 
 if [ "$(whoami)" = "root" ]; then
   exec chpst \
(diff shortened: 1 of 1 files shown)


rybons:
@jeremymv2 I'm also encountering the `Waiting for builder-github-app.pem` discussed in the forum thread you referenced. Can I consider this a bug in the latest builder-api release (habitat/builder-api/8850/20200330144923)? 

Any workaround or should I pin an older version until this is resolved?

components/builder-api/habitat/hooks/run
```diff
@@ -12,11 +12,14 @@ echo "_e{${#title},${#text}}:$title|$text|#api"  >/dev/udp/localhost/8125
 # Wait for the .pem file before starting the service, otherwise the service
 # continually panics in a tight loop, blocking the supervisor from writing
 # out the file.
-while ! [ -f "{{pkg.svc_files_path}}/builder-github-app.pem" ];
-do
-    echo "Waiting for builder-github-app.pem"
+# shellcheck disable=SC2050
+if [[ "{{cfg.api.features_enabled}}" =~ "jobsrv" ]]; then
```
smacfarlane, line 16: I think this is the right thing to do, but I am concerned that we may be assuming the file lives on disk in other places in the code that may cause a panic if it's missing.  Our api test suite wouldn't necessarily catch this because it's all driven by the studio and that expects the file to be present.  

I also can't recall the context on the above comment as to why it would enter a panic loop if the file wasn't present, but I _think_ the on-prem-depot runs without the file present, which contradicts my concern.  

The tests we run in CI generate one for us before starting the test suite (https://github.com/habitat-sh/builder/blob/master/.expeditor/templates/studiorc#L15-L17).  It doesn't need to be valid, just present, to start. This also makes me think our test suite may not hit any code paths that actually exercise the github-api-client(https://github.com/habitat-sh/builder/tree/master/components/github-api-client/src), since any pem file will do.

All this makes me wonder if there's some code chances we could make in the api or the github-api-client to better handle the optional nature of that file, but that would be a larger slice of work and not something for this PR.

jeremymv2:
@rybons we're sorry for the inconvenience this escaped defect caused. As a workaround, you can simply create an empty file `touch /hab/svc/builder-api/files/builder-github-app.pem`.

The fix will be in a release scheduled for 4/28/2020.

rybons:
@jeremymv2 No worries, I'll give that a try. Thanks!

jeremymv2:
The fix ended up in `habitat/builder-api/8885/20200504153856`

PR #1428, opened 2020-05-28
Update package dependencies for builder-worker

smacfarlane:
This unpins the dependencies for the linux builder-worker to start using the post-refresh versions of packages. It also updates the dependencies to use the renamed container exporter package.

### DIFF
--- a/components/builder-worker/habitat/plan.sh
+++ b/components/builder-worker/habitat/plan.sh
@@ -37,9 +37,9 @@ do_prepare() {
   build_line "Setting PLAN_STUDIO_PKG_IDENT=$PLAN_STUDIO_PKG_IDENT"
 
   # Compile the fully-qualified Docker exporter package identifier into the binary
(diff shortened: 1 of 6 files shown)


components/builder-worker/habitat/plan.sh
```diff
@@ -5,33 +5,11 @@ pkg_origin=habitat
...
-    core/rust/1.41.0/20200217103343
-    core/gcc/8.2.0/20190115004042
-    core/git/2.25.1/20200309023931
-    core/pkg-config/0.29.2/20190115011955
-)
+pkg_deps=(core/glibc core/openssl core/gcc-libs core/zeromq core/libsodium
+  core/libarchive core/zlib core/hab core/hab-studio core/hab-pkg-export-container
+  core/docker core/curl)
+pkg_build_deps=(core/make core/cmake core/protobuf-cpp core/protobuf-rust core/coreutils core/cacerts
+  core/rust core/gcc core/git core/pkg-config)
```
christophermaier, line 12: I find having each dependency on a line by itself makes it a lot easier to read, both in general, as well as in a `diff` situation.

PR #1432, opened 2020-06-01
Upgrade builder-minio service

jsirex:
* New version of minio allows to add servers, so I'm dropping automatic cluster in
favor of local or explicit cluster. This simplifies things dramatically
* Modify run hook to restart service instead of checksumming config
* Drop deprecated reload hook
* Minio now allows `--certs-dir` so no copying is required
* Removed `--config-dir` - deprecated
* `env` section now allows you to specify any additional environment variables
* Allow to change bind address and bind port
* Export port, bucket, access and secret keys

Signed-off-by: Yauhen Artsiukhou <jsirex@gmail.com>


components/builder-minio/habitat/default.toml
```diff
@@ -32,20 +17,29 @@ expected = 2
...
+# Bind to a specific PORT
+bind_port = 9000
 
-# Password for certificate
-ssl_cert_pw = ""
+# Setup additional environment variables to pass to minio
+# Doc: https://docs.min.io/docs/minio-server-configuration-guide.html
+[env]
+# Set minio admin user and password
+MINIO_ACCESS_KEY = "habitat-depot"
```
eeyun, line 35: This needs to be changed back to "depot" to avoid breaking our default envvars for on-prem.

components/builder-minio/habitat/default.toml
```diff
@@ -32,20 +17,29 @@ expected = 2
...
+MINIO_SECRET_KEY = "habitat-password"
```
eeyun, line 36: This needs to be changed back to "password" to avoid breaking our default envvars for on-prem.

eeyun:
Sorry for the delay on this. Thanks for fixing that up. Works great!

Issue #1419, opened 2020-05-14
Need one call that returns the full list of latest packages on a channel


gavindidrichsen:
## The Enhancement:

I want to use one call and return only the latest packages associated with a channel.

This feature request is related to #1418

### The Need:

This would be useful especially when attempting to verify the change or delta to a channel that has occurred.  For example, 
* I've uploaded 50 packages to my channel, how do I verify that the exact identifiers have been successfully uploaded?
* Are the latest packages on the channel equal to the ones I've uploaded?  This was an actual issue during the latest refresh, where an incorrect package core/mage version, which was greater than the uploaded package, was shadowing the new refreshed version.
* What's changed on the stable channel for the x86-linux target between yesterday and today?  

Being able to download yesterday's and today's full list allows a delta to be computed 

### Current Alternative

There doesn't seem to be an alternative that I know of.  However, this is not blocking us.  I do have a workaround at the minute which is to (1) download the full list (2) normalize the list by stripping off the version/timestamp and uniquify; (3) use the unique list and do individual curl's to Builder to obtain the 'latest' version for each.

If there are 700+ unique core/plans on a channel, then this means 700 individual curl requests are required to get the full 'latest' list for all channel packages

markan:
It seems like an endpoint of the form:
/depot/channels/ORIGIN/CHANNEL/latest?target=TARGET
which would for ORIGIN and CHANNEL list the latest packages for TARGET would fit the bill.

@gavindidrichsen does this sound suitable.

This looks relatively straightforward to add to the existing API, and would probably make a decent first project or pairing opportunity.

gavindidrichsen:
> /depot/channels/ORIGIN/CHANNEL/latest?target=TARGET
> a decent first project or pairing opportunity.

@markan: That looks perfect.  Yes, I'd love to pair in on this update.  Cheers!

jeremymv2:
This looks like a great idea and I would recommend a slight modification to the endpoint:
`/depot/channels/ORIGIN/CHANNEL/pkgs/latest?target=TARGET`. 

I think the above URL more closely adheres REST semantics declaratively indicating that the action requested is on `pkgs` and not on `CHANNEL`. It would fit nicely into the existing API [endpoint](https://github.com/habitat-sh/builder/blob/75682134f5a547666f6341c53d1fe208ad219242/components/builder-api/doc/api.raml#L1262-L1295)

PR #1444, opened 2020-06-05
Add API to list latest packages for channel

markan:
First pass at #1419

This provides a way to list all latest packages for a channel.

We chose the endpoint
/depot/channels/{origin}/{channel}/pkgs/latest?target=TARGET which is
mildly problematic, as it aliases with any package called latest. This
is an existing problem, in that we also alias for promote, but in that
case we can differentiate because one is GET and the other PUT.

One possibility would be to revert to the suggested
/depot/channels/ORIGIN/CHANNEL/latest?target=TARGET endpoint, but that
has it's own problems, as discussed in #1419
(https://github.com/habitat-sh/builder/issues/1419#issuecomment-632667908)

Further testing would be useful; the current test only has one package
in the origin.

### DIFF
--- a/components/builder-api/src/server/helpers.rs
+++ b/components/builder-api/src/server/helpers.rs
@@ -83,8 +83,7 @@ pub fn channel_listing_results_json<T: Serialize>(channel: &str,
     let results = ChannelListingResults { channel: channel.to_string(),
                                           target:  target.to_string(),
                                           data:    packages, };
-    let r = serde_json::to_string(&results).unwrap();
-    r
+    serde_json::to_string(&results).unwrap()
 }
 
 pub fn extract_pagination(pagination: &Query<Pagination>) -> (isize, isize) {
--- a/components/builder-api/src/server/helpers.rs
+++ b/components/builder-api/src/server/helpers.rs
@@ -39,6 +39,13 @@ pub struct PaginatedResults<'a, T: 'a> {
     data:        &'a [T],
 }
 
+#[derive(Serialize)]
+pub struct ChannelListingResults<'a, T: 'a> {
+    channel: String,
+    target:  String,
+    data:    &'a [T],
+}
+
 #[derive(Deserialize)]
 pub struct ToChannel {
     #[serde(default)]
@@ -67,6 +74,19 @@ pub fn package_results_json<T: Serialize>(packages: &[T],
                                      data:        packages, };
 
     serde_json::to_string(&results).unwrap()
+}
+
+pub fn channel_listing_results_json<T: Serialize>(channel: &str,
+                                                  target: &str,
+                                                  packages: &[T])
+                                                  -> String {
+    let results = ChannelListingResults { channel: channel.to_string(),
+                                          target:  target.to_string(),
+                                          data:    packages, };
+    let r = serde_json::to_string(&results).unwrap();
+    trace!("LOGLOG-t {} {} {}", channel, target, r);
+    println!("LOGLOG-p{} {} {}", channel, target, r);
+    r
 }
 
 pub fn extract_pagination(pagination: &Query<Pagination>) -> (isize, isize) {
--- a/components/builder-api/src/server/helpers.rs
+++ b/components/builder-api/src/server/helpers.rs
@@ -84,8 +84,6 @@ pub fn channel_listing_results_json<T: Serialize>(channel: &str,
                                           target:  target.to_string(),
                                           data:    packages, };
     let r = serde_json::to_string(&results).unwrap();
-    trace!("LOGLOG-t {} {} {}", channel, target, r);
-    println!("LOGLOG-p{} {} {}", channel, target, r);
     r
 }
 
--- a/components/builder-api/src/server/resources/channels.rs
+++ b/components/builder-api/src/server/resources/channels.rs
@@ -73,6 +73,8 @@ impl Channels {
                   web::delete().to(delete_channel))
            .route("/depot/channels/{origin}/{channel}/pkgs",
                   web::get().to(get_packages_for_origin_channel))
+           .route("/depot/channels/{origin}/{channel}/pkgs/latest",
+                  web::get().to(get_latest_packages_for_origin_channel))
            .route("/depot/channels/{origin}/{channel}/pkgs/{pkg}",
                   web::get().to(get_packages_for_origin_channel_package))
            .route("/depot/channels/{origin}/{channel}/pkgs/{pkg}/latest",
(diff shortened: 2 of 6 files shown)


components/builder-api/src/server/helpers.rs
```diff
@@ -69,6 +76,19 @@ pub fn package_results_json<T: Serialize>(packages: &[T],
...
+pub fn channel_listing_results_json<T: Serialize>(channel: &str,
+                                                  target: &str,
+                                                  packages: &[T])
+                                                  -> String {
+    let results = ChannelListingResults { channel: channel.to_string(),
+                                          target:  target.to_string(),
+                                          data:    packages, };
+    let r = serde_json::to_string(&results).unwrap();
+    trace!("LOGLOG-t {} {} {}", channel, target, r);
+    println!("LOGLOG-p{} {} {}", channel, target, r);
```
smacfarlane, line 88: I think we can remove the println! here.

components/builder-api/src/server/helpers.rs
```diff
@@ -69,6 +76,19 @@ pub fn package_results_json<T: Serialize>(packages: &[T],
...
 
+pub fn channel_listing_results_json<T: Serialize>(channel: &str,
+                                                  target: &str,
+                                                  packages: &[T])
+                                                  -> String {
+    let results = ChannelListingResults { channel: channel.to_string(),
+                                          target:  target.to_string(),
+                                          data:    packages, };
+    let r = serde_json::to_string(&results).unwrap();
+    trace!("LOGLOG-t {} {} {}", channel, target, r);
```
smacfarlane, line 87: Could we update the trace prefix message to be more descriptive about where this log is occuring?

components/builder-api/src/server/resources/channels.rs
```diff
@@ -678,9 +680,69 @@ fn get_package_fully_qualified(req: HttpRequest,
...
 
+fn get_latest_packages_for_origin_channel(req: HttpRequest,
+                                          path: Path<(String, String)>,
+                                          qtarget: Query<Target>)
+                                          -> HttpResponse {
+    let (origin, channel) = path.into_inner();
+    let channel = ChannelIdent::from(channel);
+
+    trace!("LOGLOG-t1 {} {}", origin, channel);
+    println!("LOGLOG-p1 {} {}", origin, channel);
```
smacfarlane, line 691: Same comments as above.

components/builder-db/src/models/channel.rs
```diff
@@ -159,6 +166,37 @@ impl Channel {
...
+        let channel = String::from(req.channel.as_str());
+        let target = String::from(req.target);
+
+        let result = origin_packages_with_version_array::table
+            .inner_join(origin_channel_packages::table.inner_join(origin_channels::table))
+            .filter(origin_packages_with_version_array::origin.eq(&req.origin))
+            .filter(origin_channels::name.eq(&channel))
+            .filter(origin_packages_with_version_array::target.eq(&target))
+            .filter(origin_packages_with_version_array::visibility.eq(any(req.visibility)))
+            .select(origin_packages_with_version_array::ident)
```
smacfarlane, line 183: I  think we need a `.distinct_on(origin_packages_with_version_array::name)`.  In my test instance with three packages I got the following result: 

```
> curl http://localhost/v1/depot/channels/smacfarlane/unstable/pkgs/latest?target=x86_64-linux |jq
{
  "channel": "unstable",
  "target": "x86_64-linux",
  "data": [
    {
      "origin": "smacfarlane",
      "name": "foo",
      "version": "0",
      "release": "20200605152530"
    },
    {
      "origin": "smacfarlane",
      "name": "foo",
      "version": "0",
      "release": "20200427172822"
    },
    {
      "origin": "smacfarlane",
      "name": "foo",
      "version": "0",
      "release": "20200427172727"
    }
  ]
}
```

test/builder-api/src/channels.js
```diff
@@ -119,6 +119,34 @@ describe('Channels API', function () {
...
+
+    it('returns latest packages in a channel', function (done) {
+      request.get('/depot/channels/neurosis/foo/pkgs/latest?target=x86_64-linux')
+        .type('application/json')
+        .accept('application/json')
+        .expect(200)
+        .end(function (err, res) {
+          expect(res.body.channel).to.equal('foo');
+          expect(res.body.target).to.equal('x86_64-linux');
+//          expect(res.body.data).length.to.equal(1);
```
smacfarlane, line 142: Given the above failure case, it may be worth adding a second version of the 'testapp' package, as well as possibly a differently named package to make assertions that our `distinct_on` is working correctly.

components/builder-api/src/server/resources/channels.rs
```diff
@@ -678,9 +680,67 @@ fn get_package_fully_qualified(req: HttpRequest,
...
+        Err(_) => None,
+    };
+
+    // This is a new API, so we only look at the query string not the headers.
+    let target = match qtarget.target {
+        Some(ref t) => {
+            trace!("Query requested target = {}", t);
+            t
+        }
+        None => return Err(Error::BadRequest),
```
jeremymv2, line 727: This makes me wonder if defaulting to `x86_64-linux` would be more delightful first time experience when a user discovers this endpoint? I believe often users will intuitively GET an endpoint first without passing arguments in order to interrogate its typical output. If they don't like what they see or want refining, they begin passing in arguments. Just a hunch. Thoughts? 

We should also probably document this route in our `components/builder-api/doc/api.raml` since `target` is a meaningful, if note required, argument.

markan: Definitely should add this to RAML. 

So my understanding is that target was added when we added our 2nd platform. Initially everything was x86_64-linux, but when we added windows we had to add the the query parameter to existing APIs. We made it optional to preserve backwards compatibility. I suspect in a future where we add more targets, that default makes less sense. In any event, making it mandatory on a new API seemed reasonable since there was no back-compat issues 
The reasoning behind requiring target is that

components/builder-api/src/server/resources/channels.rs
```diff
@@ -73,6 +73,8 @@ impl Channels {
                   web::delete().to(delete_channel))
            .route("/depot/channels/{origin}/{channel}/pkgs",
                   web::get().to(get_packages_for_origin_channel))
+           .route("/depot/channels/{origin}/{channel}/pkgs/latest",
```
markan, line 76: More and more I'm wondering if .route("/depot/channels/{origin}/{channel}/latest" makes more sense to avoid conflicts with a package named 'latest'.

eeyun: It's a philosophical question to be sure. It might make sense for us to namespace out `latest` as an unusable package name. The only concern with refactoring the route would be that it breaks slightly away from the established pattern but, that doesn't matter very much I don't think.  It might be the right thing to change it.

jeremymv2: A possible approach to retain clarity that `pkgs` are being queried would be to borrow from the REST patterns used by Elasticsearch endpoints, many of which use `_`, for instance:
`-XGET localhost:9200/_snapshot/_status?pretty`.

It could translate into: `/depot/channels/{origin}/{channel}/pkgs/_latest`. We could make a reasonable rule to disallow uploads of packages with beginning with `_` and also retro fit all the current relevant endpoints to accept `_latest` as a synonym for `/latest`.

Issue #1319, opened 2020-02-21
Prototype a stand alone tool to test our build order


markan:
While we want to eventually have our process in builder, a lighter weight approach would be to build an standalone tool. That would serve two purposes:
1) It would be a cheap way to prove out the ideas we are developing, without having to touch the existing builder code and retrofit it.
2) It would allow us to offline build core plans, potentially relieving schedule pressure on the final builder implementation.

The pieces of this would be
1) Extend the existing graph tool to generate a build order graph. This at first should be a relatively naive implementation to prove out the basic idea. We may want to emit this graph in some near-topological form to make the ordering explicit.
2) Take the existing build scripts for core plans refresh and adapt them to use that graph and actually build plans. This again would be intended to be as simple as possible; most likely we'd run the actual builds sequentially on a big AWS box or the like, and start by restricting ourselves to the core 'gang of 42' plans, which form the most difficult part of the problem, but avoid the full 900 odd plans in core. The output of this would be a built set of packages that we could if we wanted to upload to builder.

Our expected outcomes are
1) we'd understand if the basic concept is workable
2) we'd have a path to update core plans, even if its not ideal/our final outcome.
3) we'd have a clearer understanding of the work required to retrofit builder to do this.

The essential component of this is proving we have a viable end-to-end approach, not producing anything shippable. Work we'd explicitly not take on include dealing with updates to plans while we're building, triggering of builds, management of promotion, security, parallelism etc. It would also avoid thinking about optimizing the build order or finding any way to improve the build speed, or more robustly handle errors.

markan:
This is intended to happen prior to https://github.com/habitat-sh/builder/issues/1281.

markan:
Closed by #1445

PR #1445, opened 2020-06-11
CLI tool generates build ordering

markan:
This adds functionality to generate build ordering information from the
CLI. Of course, it's never that simple.

On the way, we:

* Rewrote the graph structure multiple times
* Rewrote the incremental cycle detection on package insertion to be about 100x faster
* Added code to solve the cyclic build problem
* Handled some interesting corner cases around
  * Tightly coupled plan dependencies (gcc & gcc-libs)
  * Plans depending on partially or fully qualified idents
  * Unbuildable plans
* Added an interned package ident structure to allow us to use more appropriate graph libraries.
* Reworked the main CLI loop to use clap option parsing,
* Added testing both at the integration level and unit level

107 commits were squashed to spare the reviewers

Co-authored-by: Mark Anderson <mark@chef.io>
Co-authored-by: Scott Macfarlane <smacfarlane@chef.io>


.expeditor/scripts/verify/builder-graph-functional.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
eeyun, line 1: Are these setup tasks required for users that want to execute these tests in a studio without expeditor or build-kite?

.gitignore
```diff
@@ -26,3 +26,6 @@ upstream.toml
 #### Generated Files
 components/builder-protocol/src/message
 !components/builder-protocol/src/message/mod.rs
+
+#### Vendored test helpers
+test/test_helper
```
eeyun, line 31: It seems the test.sh script for the component reaches out to this directory for some test related goodies:      `local library_install_path="$project_root/test/test_helper/$library" `

Is this dir in a fixed state that won't change?

components/builder-graph/src/data_store.rs
```diff
@@ -25,16 +32,181 @@ use crate::{config::Config,
...
+        let file = File::open(&path)?;
+        let reader = BufReader::new(file);
+        let u: SerializedDatabase = serde_json::from_reader(reader)?;
+        Ok(u)
+    }
+
+    pub fn package_count(&self) -> usize { self.packages.len() }
+}
+
+#[cfg(test)]
```
eeyun, line 126: Just a nit - In most of our other components where we have unit tests we've typically created an aggregate test section at the bottom of the functional code. Is there a particular style reason we have this here as opposed to the end of the file?

.expeditor/scripts/verify/builder-graph-functional.sh
```diff
@@ -0,0 +1,23 @@
...
+
+set -euo pipefail
+
+echo "--- Generating signing key"
+hab origin key generate "$HAB_ORIGIN"
+
+# TODO: core/graphviz doesn't appear to support png
```
eeyun, line 8: Did this commented block need to be left in?

components/builder-graph/src/data_store.rs
```diff
@@ -55,8 +227,41 @@ impl DataStore {
...
 
-    pub fn get_job_graph_packages(&self) -> Result<Vec<PackageWithVersionArray>> {
+        let conn = self.pool.get_conn()?;
+
+        let query = origin_projects::table.select(origin_projects::name)
+                                          .filter(origin_projects::auto_build.eq(true))
+                                          .filter(origin_projects::target.eq(target.to_string()))
+                                          .filter(origin_projects::name.eq_any(package_idents));
+
+        // let debug = diesel::query_builder::debug_query::<diesel::pg::Pg, _>(&query);
```
eeyun, line 249: Does this need to remain?

components/builder-graph/src/data_store.rs
```diff
@@ -55,8 +227,41 @@ impl DataStore {
...
+
+        let result: Vec<BuilderPackageIdent> = query.get_results(&conn).unwrap();
```
eeyun, line 251: Is this unwrap safe, is there any case in which the query selection might return empty?

components/builder-graph/src/data_store.rs
```diff
@@ -69,25 +274,156 @@ impl DataStore {
...
+                "origin_packages_with_version_array.name, \
+        origin_packages_with_version_array.target, \
+        string_to_array(origin_packages_with_version_array.version_array[1],'.')::\
+        numeric[] desc, origin_packages_with_version_array.ident_array[4] desc",
+            ))
+            .filter(origin_channels::name.eq(&channel))
+            .filter(origin_packages_with_version_array::origin.eq(&origin))
+            .filter(origin_packages_with_version_array::target.eq(target.to_string()));
+
+        let result: Vec<BuilderPackageIdent> = query.get_results(&conn).unwrap();
```
eeyun, line 323: As above is this unwrap safe? As a follow up, I checked on some of our other database query functions to verify the behavior, it seems like we typically hand off the result of the diesel query to the calling function inside of a match. Is the unwrap necessary? https://github.com/habitat-sh/builder/blob/049a18bd2d6694fc4ea9344805e9010d1f9edb79/components/builder-db/src/models/package.rs#L447

components/builder-graph/src/data_store.rs
```diff
@@ -69,25 +274,156 @@ impl DataStore {
...
+        sd.write_to_file(filename).unwrap();
+        Ok(())
+    }
+}
+
+// When we are working with the serialized dataset (mostly for tests)
+// this is a list of things that are known unbuildable. This could
+// be stored in the serialized data, but for now we're defining it statically
+// here.
+lazy_static! {
```
eeyun, line 359: Ok, this is awesome. :joy:

.expeditor/scripts/verify/builder-graph-functional.sh
smacfarlane: These should all be part of a normal developer setup.  The exception being the build and BLDR_GRAPH_PATH. The reason behind that is that `builder-graph` doesn't fit into the same workflow as the rest of our components; It's not a service so `build-builder` doesn't work and doesn't have `habitat-dev`plan definition. We've been using a `cargo build` workflow external to the studio to develop this package, so the `test/builder-graph/test.sh` script defaults to using the `target/debug/bldr-graph` binary produced by that. `BLDR_GRAPH_PATH` is an escape hatch, primarily for CI,  to tell the driver "no, I actually want to use this other bldr-graph binary", so shouldn't be needed in a normal workflow.

.expeditor/scripts/verify/builder-graph-functional.sh
smacfarlane: Probably not?  I left it as a reminder we had a test we weren't exercising in CI, but I don't have strong feelings either way.  I'll take a look and see if there's an easy fix, and remove otherwise.

.gitignore
smacfarlane: Yup, this dir should always be empty in the repository. We had a preference for not committing the vendored  bats helpers, and the `test/builder-graph/test.sh` script ensures that they're populated.  It does add an additional failure point in CI, but I figured that if you can't `git clone` from GitHub we're probably not going to be able to run CI anyways.

markan:
Closes #1319

PR #1446, opened 2020-06-16
Remove devstudio dependency in CI

smacfarlane:
In the process of trying to determine why [these](https://buildkite.com/chef-oss/habitat-sh-builder-master-verify/builds/889#eec1d823-46b1-4aac-9adf-7cb83ed1dad7) CI tests were failing, I hit a wall with being able to debug the reproduction due to how the tests were called. 

This work is an attempt to make it easier to debug. It uses `hab pkg build` to generate packages, rather than using specialized build calls from a mutated .studiorc to perform the builds and tests as part of a `studio enter`. 

One drawback to this is that it's no longer able to use sccache to speed up the jobsrv and worker builds.  It's possible that this doesn't matter in practice though. 


.expeditor/scripts/verify/builder-api-functional.sh
```diff
@@ -5,13 +5,142 @@ set -euo pipefail
 echo "--- Generating signing key"
 hab origin key generate "$HAB_ORIGIN"
 
-echo "--- Updating .studiorc" 
-cat .expeditor/templates/studiorc >> .studiorc
+# Using the dev studio is very fragile so we get to replicate some of the
```
eeyun, line 8: I'd like some more information on this problem before approving this PR. What about the studio is very fragile? I'm not going to gatekeep progress on this, but I'd (probably obviously) prefer we have a single way these tests are getting called from local-dev on through to CI if at all possible. This feels like a pretty heavy hammer from the outside.

smacfarlane: I should have been more clear here. This is in the context of CI only, and fragile is perhaps not the right word.  

The way our tests drivers are set up now are intended be driven from within a studio environment which is inherently an interactive process.  Trying to get that to fit in the shape of CI, while it works, when it fails it becomes incredibly difficult to debug as you have an additional layer of shell wrapping around a long chain of actions that don't always fail in the way you want them to in CI. 

In addition, the way our packages are built in the studio is _slightly_ different than how our final packages are built, in that the resulting binaries are injected into the existing packages. This doesn't give us a great way to verify that the new version is actually loaded (unless I've missed something, which is entirely possible).  

The majority of this change is bringing the build of the packages more in line with how our final package would be built, and then the necessary running of the supervisor and our services since we no longer have a studio providing these for us (There's probably a story here in making testing Habitat services in CI easier).  The bulk of the test helpers are not necessary in CI, as we can be guaranteed a clean database every time. I'm not opposed to delegating to that script rather than calling npm directly, but I have a preference for keeping our call stack as flat as possible in CI, as well as avoiding unnecessary things that can introduce failure.  We would have do some additional refactoring as the test driver relies on some of the studio helpers, such as the psql command being wrapped by a function of the same name.

smacfarlane:
I'm going to go ahead and close this out for now.  30m turnaround for CI on PRs is too high and I don't have the necessary cycles to work on improving that at this time.

PR #1447, opened 2020-06-18
Smma/graph trait

### LINKED ISSUE #1456: Factor out acyclic/cyclic build graph APIs into a trait.
We want to abstract out the difference between the two implementations as much as possible. While the two implementations will differ on how they compute and represent builds, they still have a lot in common in the way edges are added and queried.

markan:
This is some refactoring work to make it easier to run the new and old graph side by side.

Closes #1456


components/builder-graph/src/package_graph_trait.rs
```diff
@@ -0,0 +1,38 @@
...
+#[derive(Debug)]
+pub struct Stats {
+    pub node_count:     usize,
+    pub edge_count:     usize,
+    pub connected_comp: usize,
+    pub is_cyclic:      bool,
+}
+
+pub trait PackageGraphTrait: Send + Sync {
+    fn build(&mut self,
```
davidMcneil, line 26: Maybe this could be `new`? If having empty graphs is useful, you could add an `empty` trait method which calls `build` on an empty slice.

components/builder-graph/src/package_graph_trait.rs
```diff
@@ -0,0 +1,38 @@
...
+             packages: &[originsrv::OriginPackage],
+             use_build_deps: bool)
+             -> (usize, usize);
+    fn extend(&mut self,
+              package: &originsrv::OriginPackage,
+              use_build_deps: bool)
+              -> (usize, usize);
+    fn check_extend(&mut self, package: &originsrv::OriginPackage, use_build_deps: bool) -> bool;
```
davidMcneil, line 34: This is a pretty high level operation. I wonder if it could be expressed genericly as a set of other graph operations (eg `contains`)? This would allow `check_extend` to have a default implementation. This could potentially reduce effort and code duplication as more things implement this trait.

components/builder-graph/src/target_graph.rs
```diff
@@ -55,7 +55,11 @@ impl TargetGraph {
         }
     }
 
-    pub fn graph_mut(&mut self, target_str: &str) -> Option<&mut PackageGraph> {
+    #[allow(clippy::borrowed_box)]
+    // Clippy wants
+    // pub fn graph_mut(&mut self, target_str: &str) -> Option<&mut dyn PackageGraphTrait>
+    // But then lifetime issues intrude if we follow the as_ref pattern above.
+    pub fn graph_mut(&mut self, target_str: &str) -> Option<&mut Box<dyn PackageGraphTrait>> {
```
davidMcneil, line 62: Using the signiture `pub fn graph_mut(&mut self, target_str: &str) -> Option<&mut (dyn PackageGraphTrait + 'static)>` fixes the clippy and lifetime errors. The `&` case has the `'static` lifetime as well, but for some reason the compiler is better at infering it in that case :shrug:.

davidMcneil: `Box` with a trait object implicitly has the `'static` lifetime. This [thread](https://users.rust-lang.org/t/box-with-a-trait-object-requires-static-lifetime/35261) has some more details if you are interested.

components/builder-graph/src/package_graph_trait.rs
markan: We made an implementation decision to try and slide the trait in first, then address later some of the type and shape issues with the trait. 
In this case there will only be the old and new graph code here, and hopefully only the new code long term. The trait becomes a little bit redundant at that point, like scaffolding left up after a refurbishing.

components/builder-graph/src/acyclic_package_graph.rs
```diff
@@ -59,39 +35,19 @@ fn short_name(name: &str) -> String {
...
-            self.package_max += 1;
-
-            (self.package_max - 1, node_index)
-        }
-    }
-
-    pub fn build<T>(&mut self, packages: T, use_build_deps: bool) -> (usize, usize)
-        where T: Iterator<Item = originsrv::OriginPackage>
-    {
+impl PackageGraphTrait for AcyclicPackageGraph {
```
jeremymv2, line 46: 👍

components/builder-graph/src/target_graph.rs
```diff
@@ -28,25 +29,25 @@ pub struct TargetGraphStats {
 
 #[derive(Default)]
 pub struct TargetGraph {
-    graphs: HashMap<PackageTarget, PackageGraph>,
+    graphs: HashMap<PackageTarget, Box<dyn PackageGraphTrait>>,
```
jeremymv2, line 32: :wrench: interesting choice! 👍

components/builder-graph/src/package_graph_trait.rs
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) 2017 Chef Software Inc. and/or applicable contributors
```
jeremymv2, line 1: very minor: You have all other ones at 2020. this one might have slipped past 😃

PR #1449, opened 2020-06-24
Unpin habitat_core crate

smacfarlane:
This PR removes the `habitat_core` pins added in 720ce93ecda78a439411e52b05e171a4fd163b67.

The changes are largely mechanical, working the `Result` returned by `PackageArchive::new()` and the `String->&str` return through.  

Lines 1023-1030 and 1044-1050 in  components/builder-api/src/server/resources/pkgs.rs bear closer examination to ensure I've use the correct pattern for returning a new error so that the failure is reported correctly back to the user.


smacfarlane:
@jeremymv2 conflicts fixed up!

PR #1477, opened 2020-07-29
Smma/graph trait in scheduler 1457

### LINKED ISSUE #1457: Rework the handler/scheduler code to use the graph trait
This is a followon to #1456

This takes the trait created and reworks the jobsrv handler and scheduler code 

Converting the code to use these as a trait enables us to swap them out and do various kinds of A/B testing. Specifically out of scope is the actual build, which will at first not be supported for the cyclic version.

markan:
This is the next phase of working the graph trait through the code. This is a preview for early review and should not be merged. It's based on smma/graph_merge and will need a new PR once that work merges.

Closes #1457


jeremymv2 (approved):
With this it seems like we are fast approaching the ability to test the new graph trait code. Nice work!

Issue #1389, opened 2020-04-13
Cannot open any public package without login at bldr.habitat.sh


jsirex:
1. Go to builder
2. Find any public package
3. Open it

Expected Result: Package view is opened
Actual Result: Sign In page is forced

jsirex:
UPD: 
4. Sign in into builder
5. Open public package you don't own

Expected Result: Package view is opened
Actual Result: Sign In page is forced

jeremymv2:
Just a note on the why: as of 2 releases ago, the frontend UI now queries `/settings/{origin}/{pkg_name}` and gets a 401 because there is no existing session. The backend API `/settings` endpoint expects an existing session token, and for it to be a member of the target origin:

https://github.com/habitat-sh/builder/blob/75682134f5a547666f6341c53d1fe208ad219242/components/builder-api/src/server/resources/settings.rs#L59-L61

jeremymv2:
**Context**
The frontend UI is making a request to `https://bldr.habitat.sh/v1/settings/ORIGIN/PKG_NAME` in order to determine the package's `visibility` settings. The Package Visibility information is then displayed as either public or private.

The trouble arises here because the `/settings` API endpoint requires authentication information in the request as the visibility information can be considered sensitive metadata. If a user is unauthenticated, the UI receives a 401 unauthorized response from the backend `/settings` endpoint and forces the login window; even if the package was public.

**Possible Fix**
The UI understands if the user is current authenticated or not. In the event that the user is not, the UI could silently eat the 401 error and possibly display some helpful information that the package visibility setting for the package is missing until the user first authenticates.

PR #1481, opened 2020-08-07
Don't request package settings if not an origin member

scottopherson:
Fixes issue https://github.com/habitat-sh/builder/issues/1389 where users are redirected and can't view a public package if they are unauthenticated. The request to `/settings` is now only made for users who are authenticated origin members.

Before:
b

After:
c

### DIFF
--- a/components/builder-web/app/package/package/package.component.ts
+++ b/components/builder-web/app/package/package/package.component.ts
@@ -14,7 +14,7 @@
 
 import { Component, OnInit, OnDestroy } from '@angular/core';
 import { combineLatest, Subject, Subscription } from 'rxjs';
-import { filter, takeUntil } from 'rxjs/operators';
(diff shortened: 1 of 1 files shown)


components/builder-web/app/package/package/package.component.ts
```diff
@@ -66,15 +69,23 @@ export class PackageComponent implements OnInit, OnDestroy {
...
-    combineLatest(origin$, name$)
+    combineLatest(origin$, name$, isOriginMember$)
       .pipe(takeUntil(this.isDestroyed$))
-      .subscribe(() => {
-        this.fetchPackageSettings();
+      .subscribe(([origin, name, isOriginMember]) => {
         this.fetchPackageVersions();
+        if (isOriginMember) {
+          this.fetchPackageSettings();
+        }
```
scottopherson, line 88: The important `if` statement.

PR #1483, opened 2020-08-11
enable ssh access to windows worker

### LINKED ISSUE #1473: Enable SSH on windows workers
This should streamline remote access across platforms and allow us to not be limited to RDP

mwrock:
resolves #1473

### DIFF
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -165,6 +165,6 @@ variable "enabled_features" {
   default = []
 }
 
-variable "aws_public_key" {
-  description = "File path to the public key to use for authentication to windows builder"
+variable "connection_public_key" {
+  description = "File path to the public key to use for SSH authentication to windows builder"
 }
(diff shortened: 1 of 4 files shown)


terraform/variables.tf
```diff
@@ -164,3 +164,7 @@ variable "enabled_features" {
   type    = list(string)
   default = []
 }
+
+variable "aws_public_key" {
+  description = "File path to the public key to use for authentication to windows builder"
```
jeremymv2, line 169: In order to reduce confusion as to which key this actually is or should be, we could mirror the variable name with the name of the one used for the ssh connection to do the actual provisioning of the linux nodes: `connection_public_key`. This is the same variable used in `habitat-sh/cloud-environments` and would operationally make sense to harmonize the ssh key for all connections.

We could also clarify the description or the variable name such that it includes `ssh`? Perhaps:
```
File path to the public key to use for SSH authentication to windows builder
```

mwrock: great suggestion! Applied in 633fa54c7db48dd17cf875b199d6bab7da3bf7f0

PR #1486, opened 2020-08-12
Ma/fixup lints

markan:
At some point it looks like our lints got stricter; fixing the warnings so we can see new issues. 

Fixes are broken up into separate commits for
* unused 'use'
* nth(0) instead of next() in iterators
* removing deprecated lint for unused labels
* fixing a possibly missing test


components/builder-db/src/models/origin.rs
```diff
@@ -434,6 +434,7 @@ mod tests {
         let maintainer = OriginMemberRole::Maintainer;
         let administrator = OriginMemberRole::Administrator;
         let owner = OriginMemberRole::Owner;
+        assert_eq!(owner > administrator, true);
```
jeremymv2, line 437: 💯

PR #1489, opened 2020-08-14
updates builder-api-proxy to allow for nginx rate limiting

collinmcneese:
updates builder-api-proxy to allow for nginx rate limiting to depot locations

Allows for uses of `on-prem-builder` to have a method to implement nginx rate limiting for depot locations if needed to help as a component prevent network over-saturation in the event of an intended or accidental mass package deployment.  Applies configuration to `depot` locations from `builder-api-proxy` and seeds with some sample initial values, also adding a toggle to default to false.

Signed-off-by: Collin McNeese <cmcneese@chef.io>


mwrock (approved):
There was concern over the limits being on the low side but a "toggle" was added which turns this off by default.

PR #1493, opened 2020-08-21
Adds settings deletion route and tests

eeyun:

### DIFF
--- a/components/builder-api/src/server/authorize.rs
+++ b/components/builder-api/src/server/authorize.rs
@@ -55,8 +55,8 @@ pub fn authorize_session(req: &HttpRequest,
             Some(r) => r,
             None => {
                 let r = OriginMemberRole::ReadonlyMember;
-                warn!("authorize_session: minimum role parameter not set! Assuming {}",
-                      r);
+                debug!("authorize_session: minimum role parameter not set! Assuming {}",
+                       r);
                 r
             }
         };


components/builder-api/src/server/resources/settings.rs
```diff
@@ -47,7 +50,9 @@ impl Settings {
            .route("/settings/{origin}/{name}",
                   web::get().to(get_origin_package_settings))
            .route("/settings/{origin}/{name}",
-                  web::put().to(update_origin_package_settings));
+                  web::put().to(update_origin_package_settings))
+           .route("/settings/{origin}/{name}",
+                  web::delete().to(delete_origin_package_settings));
```
jeremymv2, line 55: Splendid! Can we get an entry into our raml doc for this route as well?

components/builder-api/src/server/resources/settings.rs
```diff
@@ -168,6 +173,53 @@ fn update_origin_package_settings(req: HttpRequest,
...
+                                                                                  Some(account_id),
+                                                                                  &ident.origin),
+                             page:       1 as i64,
+                             limit:      50 as i64, };
+
+    // Validate that no packages for the setting still exists
+    match Package::list(lpr, &*conn).map_err(Error::DieselError) {
+        Ok((_packages, count)) => {
+            if count > 0 {
+                return HttpResponse::new(StatusCode::UNPROCESSABLE_ENTITY);
```
jeremymv2, line 204: While technically probably an appropriate error, I personally believe that a 409 is a bit more helpful in this context. After reading the [RFC](https://tools.ietf.org/html/rfc4918#section-11.2) on 422, I understand it as: the request and the included entity was syntactically correct but semantically didn't make sense. Whereas a 409: the request could not be completed due to a conflict with the current state of the target resource.

A pattern that we used on the origin deletion was to respond with a 409 and also [embelish](https://github.com/habitat-sh/builder/blob/master/components/builder-api/src/server/resources/origins.rs#L315-L317) the response with a [body](https://github.com/habitat-sh/builder/blob/76ec0d7a3dc05a3766380a673d523c62acf1043e/test/builder-api/src/origins.js#L381) containing information for the user-agent to resolve the conflict. It would be a bit more work but may be a more helpful UX when we implement the hab CLI portion of this. Thoughts?

eeyun: Yeah I like that Idea. I'll get that into the work I'm doing now with wiring this into the deletion process. Will push that up shortly.

components/builder-api/src/server/resources/settings.rs
eeyun: Absolutely, good catch!

PR #1499, opened 2020-08-26
Return Ok when public key already exists

eeyun:

### DIFF
--- a/components/builder-api/src/server/resources/origins.rs
+++ b/components/builder-api/src/server/resources/origins.rs
@@ -516,46 +516,62 @@ fn upload_origin_key(req: HttpRequest,
                      -> HttpResponse {
     let (origin, revision) = path.into_inner();
 
-    let account_id =
-        match authorize_session(&req, Some(&origin), Some(OriginMemberRole::Administrator)) {
-            Ok(session) => session.get_id(),
-            Err(err) => return err.into(),
-        };
+    // Since this route allows users to upload keys, we verify their membership
+    // before we determine if the key they're using actually exists. This is a
+    // backward compatibility workaround needed when RBAC was introduced. hab
+    // pkg upload optimistically uploads keys as well as packages, but the RBAC
+    // changes only give 'members' and 'maintainers' upload permissions for
+    // packages, not keys.
+    if let Err(err) = authorize_session(&req, Some(&origin), Some(OriginMemberRole::Member)) {
+        return err.into();
+    }
 
     let conn = match state.db.get_conn().map_err(Error::DbError) {
         Ok(conn_ref) => conn_ref,
         Err(err) => return err.into(),
     };
 
-    match parse_key_str(&body) {
-        Ok((PairType::Public, ..)) => {
-            debug!("Received a valid public key");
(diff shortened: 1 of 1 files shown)


components/builder-api/src/server/resources/origins.rs
```diff
@@ -516,46 +516,61 @@ fn upload_origin_key(req: HttpRequest,
...
-            debug!("Received a secret key instead of a public key");
-            return HttpResponse::new(StatusCode::UNPROCESSABLE_ENTITY);
-        }
-        Err(e) => {
-            debug!("Invalid public key content: {}", e);
-            return HttpResponse::new(StatusCode::UNPROCESSABLE_ENTITY);
-        }
-    }
+    match OriginPublicSigningKey::get(&origin, &revision, &conn) {
+        Ok(_) => HttpResponse::new(StatusCode::CONFLICT),
```
markan, line 531: Do we have a way to test that the hab pkg upload command behaves as we expect (and will continue to do so?)

components/builder-api/src/server/resources/origins.rs
```diff
@@ -516,46 +516,61 @@ fn upload_origin_key(req: HttpRequest,
                      -> HttpResponse {
     let (origin, revision) = path.into_inner();
```
markan, line 518: I might add a comment to the effect:

We return conflict when uploading a pre-existing key even if the user doesn't have permissions to actually upload the key. This is a backward compatibility workaround needed when RBAC was introduced. `hab pkg upload` optimistically uploads keys as well as packages, but the RBAC changes only give maintainers upload permissions for packages, not keys. Returning permission denied broke this flow. Somewhat silently accepting the upload was chosen as a way to not break that path while still maintaining the chosen RBAC semantics.

components/builder-api/src/server/resources/origins.rs
```diff
@@ -516,46 +516,61 @@ fn upload_origin_key(req: HttpRequest,
...
-            return HttpResponse::new(StatusCode::UNPROCESSABLE_ENTITY);
-        }
-        Err(e) => {
-            debug!("Invalid public key content: {}", e);
-            return HttpResponse::new(StatusCode::UNPROCESSABLE_ENTITY);
-        }
-    }
+    match OriginPublicSigningKey::get(&origin, &revision, &conn) {
+        Ok(_) => HttpResponse::new(StatusCode::CONFLICT),
+        Err(_) => {
```
davidMcneil, line 532: Minor - We could get rid of this `match` by having:

```
if OriginPublicSigningKey::get(&origin, &revision, &conn).is_ok() {
    return HttpResponse::new(StatusCode::CONFLICT);
}
```

and then the `Err` case of the `match` would just be the rest of the function.

components/builder-api/src/server/resources/origins.rs
eeyun: We don't have a way to automatically test this in our current testing paradigm. We could add something to the pipeline to do so. But I don't think it would be part of this PR. The basic upload behaviors are already thoroughly tested however.

Issue #1480, opened 2020-08-07
Redirect unauthenticated users who landed on a private artifact


SEAjamieD:
With the latest fix allowing viewing access for unauthenticated users to public packages, it was noticed that, as an example:

A user has a link bookmarked to a public artifact
That public artifact is then made private
The user attempts to visit the now private artifact (while unauthenticated)
The page displays the template with all empty data

It would be ideal to have a case like that redirect to another desired location, likely sign in.

PR #1501, opened 2020-08-27
Redirect non-origin members from private artifact view

scottopherson:
If an unauthenticated or non-origin member happens to navigate directly to a private artifact, they are redirected to the top-level packages route `/pkgs` instead of seeing a blank artifact view.

Fixes https://github.com/habitat-sh/builder/issues/1480

**Before**

**After**

### DIFF
--- a/components/builder-web/app/package/package-release/package-release.component.ts
+++ b/components/builder-web/app/package/package-release/package-release.component.ts
@@ -17,6 +17,7 @@ import { Title } from '@angular/platform-browser';
 import { Subject } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
 import { AppStore } from '../../app.store';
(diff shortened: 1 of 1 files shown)


components/builder-web/app/package/package-release/package-release.component.ts
```diff
@@ -34,6 +35,14 @@ export class PackageReleaseComponent implements OnDestroy {
       .subscribe(({ origin, name, version, release }) => {
         this.title.setTitle(`Packages › ${origin}/${name}/${version}/${release} | ${store.getState().app.name}`);
       });
+
+    this.store.observe('packages.ui.current.errorMessage')
+      .pipe(takeUntil(this.isDestroyed$))
+      .subscribe(errorMessage => {
```
SEAjamieD, line 41: smart! 💡

PR #1505, opened 2020-09-01
Update Builder to use new crypto types

christophermaier:
Uses code from habitat-sh/habitat#7887... should be merged after that PR is merged.


components/builder-api/src/config.rs
```diff
@@ -406,7 +391,7 @@ mod tests {
         assert_eq!(config.api.log_path,
                    PathBuf::from("/hab/svc/hab-depot/var/log"));
         assert_eq!(config.api.key_path,
-                   PathBuf::from("/hab/svc/hab-depot/files"));
+                   KeyCache::new("/hab/svc/hab-depot/files"));
```
jeremymv2, line 394: Perhaps not for this PR, but I think it would be nice to make a note here or decide on some future work to give `KeyCache` its own dedicated directory. This current directory contains non `KeyCache` type keys, for instance an RSA private key for the builder github app also exists here.

christophermaier: I had that same thought... in practice, it won't really matter, as all Builder needs the `KeyCache` for is grabbing Builder encryption keys, but it would be nice to have that conceptual distinction.

Also, isn't that the directory where `hab file upload` files go?

jeremymv2: > Also, isn't that the directory where hab file upload files go?

That's exactly how the `.pem` file gets there. Also, I meant to add this comment to line 127 and not the test. Anyways, I think in the future we could change this to `/hab/svc/builder-api/var/keycahe` or something.

components/builder-api/src/server/resources/origins.rs
```diff
@@ -430,60 +411,24 @@ fn create_keys(req: HttpRequest, path: Path<String>, state: Data<AppState>) -> H
...
-            error!("create_keys: Failed to create private key, err={}", err);
-            return err.into();
-        }
+    if let Err(e) = save_secret_origin_signing_key(account_id,
+                                                   &origin,
+                                                   &state.config.api.key_path,
+                                                   &secret,
+                                                   &*conn)
+    {
+        error!("create_keys: Failed to create private key, err={}", e);
```
jeremymv2, line 430: It might be valuable to capture the origin here. Same for line 420 above. This could prove useful when sifting through the logs. Thoughts on the value add?  Just thinking that if we do end up adding origin to the error, it might also be nice to distinguish that it's the origin private|public _signing_ key.

components/builder-api/src/server/resources/origins.rs
```diff
@@ -867,32 +755,19 @@ fn upload_origin_secret_key(req: HttpRequest,
...
-            debug!("{}", err);
-            err.into()
-        }
+    if let Err(e) = save_secret_origin_signing_key(account_id,
+                                                   &origin,
+                                                   &state.config.api.key_path,
+                                                   &key,
+                                                   &*conn)
+    {
+        error!("create_keys: Failed to create private key, err={}", e);
```
eeyun, line 766: Trivial but is it worth clarifying in the error message here what type of private key we failed to create? Might also be good to include the origin the key was associated with?

components/builder-core/src/integrations.rs
```diff
@@ -1,132 +1,66 @@
-// Copyright (c) 2017 Chef Software Inc. and/or applicable contributors
+//! This module holds code that's common to dealing with the
```
eeyun, line 1: This isn't an issue with your PR necessarily, more a comment on the previous state of things: Naming things is hard, and in builder our integrations primitive is specific to cloud integrations like docker, aws, etc. I find myself wondering why our crypto code was in this module to begin with.

eeyun (commented):
Any PR that removes more code than it adds is truly delightful. I had a few comments, nothing serious. All my functional testing in the studio seems to work. So i'm :+1: on this. My only remaining question is around https://github.com/habitat-sh/builder/pull/1505/files#diff-0cb2e72f725bc8ddc402669486038752R16 . Is this because we're waiting on these crypto changes to merge to hab_core? If so, what is timeline if we wait for that code to merge so as not to have these annotations in the Cargo.toml?

jeremymv2 (approved):
This looks fantastic and clarifies so much. With the extra clarity, and with fixtures used in the `habitat` sister PR, I bet we could probably implement a functional API test for uploading encrypted origin secrets! Something we've wanted to do for a long time.

components/builder-core/src/integrations.rs
christophermaier: Heh... I was wondering the same thing, and meant to just rename this, but forgot. Thanks for reminding me :joy:

components/builder-api/src/server/resources/origins.rs
christophermaier: Can do!

components/builder-api/src/server/resources/origins.rs
christophermaier: Sounds good!

christophermaier:
> My only remaining question is around https://github.com/habitat-sh/builder/pull/1505/files#diff-0cb2e72f725bc8ddc402669486038752R16 . Is this because we're waiting on these crypto changes to merge to hab_core? If so, what is the timeline if we wait for that code to merge so as not to have these annotations in the Cargo.toml?

@eeyun The cargo changes were needed so I could get this out for review and have it actually be able to compile. The plan is to merge the Habitat PR first, then modify things here to pick up the changes from master (rather than my branch), and then merge this PR. I think we should be able to merge both PRs either today or tomorrow at the latest.

eeyun (approved):

PR #1508, opened 2020-09-02
bump postgres max_connections for local studio dev

jeremymv2:
Sometimes a developer's CPU thread number can cause the default workers to exceed the prior limits, inhibiting their ability to test in the studio.

This raises the `max_connections` from the default of 100->200.

### DIFF
--- a/support/builder/datastore.toml
+++ b/support/builder/datastore.toml
@@ -1,3 +1,4 @@
+max_connections = 200
 max_locks_per_transaction = 128
 dynamic_shared_memory_type = 'none'
 port = 5433 # To avoid conflict with travis' own postgres instance


eeyun:
Am I mistaken that this also alters the default max connections for on-prem? If so, maybe we ought to mutate this in the studiorc or the like?

jeremymv2:
@eeyun This is only going to take effect in the studio:  https://github.com/habitat-sh/builder/blob/88596448ae0a33f23bcc929457bbe8abe919450e/.studiorc#L42-L45

The on-prem Builder does not use this file: https://github.com/habitat-sh/on-prem-builder/blob/4c9d2ed95ec9529b6eb0dd2e83eadc37aa0ba79f/scripts/provision.sh#L39-L50

eeyun (approved):

markan (approved):

PR #1512, opened 2020-09-09
[MIGRATION] Store key bodies as TEXT, not BYTEA

christophermaier:
There doesn't seem to be any reason to store key bodies as byte
arrays; let's simplify things!


jeremymv2:
This is looking good and I'm all for the simplifications @christophermaier! I don't see any reason either why it was necessary to store the keys as "raw bytes".  We should certainly test things out in Acceptance though to ensure there isn't any "fidelity" loss on any of the existing conversions from bytea to text.

jeremymv2 (approved):

PR #1516, opened 2020-09-14
[MIGRATION] Stop storing encrypted content as base64-encoded strings

christophermaier:
Previously, we would store all encrypted data in the database as a
base64-encoded string. However, this doesn't actually do anything
meaningful, since our encrypted messages are themselves already
strings (partly plaintext metadata indicating which key(s) are
involved in the encryption/decryption process, and partly
base64-encrypted ciphertext) and not simply binary data. Adding an
additional layer of base64 encoding amounts to "busywork", in addition
to increasing the size of the stored data by 1/3.

We also modify the `crypto` module to stop encoding and decoding
encrypted data, which means that the rest of the code should continue
to work transparently. We still use `String` as the type we pass into
and out of this module because the rest of Builder doesn't really care
what kind of encryption we're doing in there; the `String` format of
our Habitat encrypted messages is fine. It's just that now this
`String` isn't base64-encoded.

The one time where it *is* important to base64-encode an encrypted
message is when we're generating tokens. There's only one place we do
that, though, so it's easy enough to simply encode it where it's
generated.

If there are any old Builder installations that have not already run
the encryption migration that is implemented in Rust code, this should
still be OK, because our pure SQL migration here will run before it
does. Since the data in the database at that point will no longer be
base64-encoded, and since the Rust migration code uses our
newly-modified encryption and decryption functions, everything should
continue to work.

### DIFF
--- a/components/builder-api/src/server/resources/origins.rs
+++ b/components/builder-api/src/server/resources/origins.rs
@@ -1619,7 +1619,6 @@ fn save_public_origin_signing_key(account_id: u64,
                                   key: &PublicOriginSigningKey,
                                   conn: &PgConnection)
                                   -> Result<()> {
-    // Note that this is *not* base64 encoded
     let key_body = key.to_key_string();
 
     let new_pk = NewOriginPublicSigningKey { owner_id: account_id as i64,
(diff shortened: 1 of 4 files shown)


jeremymv2 (approved):
This looks like a very pragmatic simplification of what had been layers of obfuscation. Thank you!

PR #1518, opened 2020-09-15
[MIGRATION + REFACTOR] Ensure keys are encrypted at rest in database

christophermaier:
There's a mixture of refactorings here to further consolidate key retrieval operations. Ideally, we just "ask for keys" and we get the core business logic types, with things like decryption handled transparently for us. Similarly, when saving keys to the database, we should only have to pass in one of the core business logic key types, and the encryption and saving is handled for us.

Additionally, recent refactorings revealed that origin secret encryption keys were not themselves encrypted in the database. This PR introduces a migration to encrypt everything currently in the database, and handle encryption/decryption on save/retrieval as necessary.

Of particular note for reviewers is the structure set up around the "live" migration code. Since this migration involves complex business logic, it was not possible to express it as pure SQL, so we need some actual Rust code.

### DIFF
--- a/components/builder-db/src/models/migration_support.rs
+++ b/components/builder-db/src/models/migration_support.rs
@@ -1,0 +1,48 @@
+//! Accessory code to add to various models to support various
+//! non-SQL, active data migrations.
+//!
+//! The idea is that these are not intended for use in "normal" code,
+//! wil be removed at some point in the future, and thus should be
+//! sequestered apart as "special" code
+
+use crate::models::keys as db_keys;
+use diesel::{self,
+             pg::PgConnection,
+             result::QueryResult,
+             ExpressionMethods,
+             QueryDsl,
+             RunQueryDsl};
+use habitat_core::crypto::keys as core_keys;
+
+impl db_keys::OriginPrivateEncryptionKey {
+    /// Encrypts any unencrypted secret encryption keys in the
+    /// database with the given Builder secret encryption key.
+    ///
+    /// Returns the number of updated rows for user feedback purposes.
+    ///
+    /// Should be run in a transaction!
+    pub fn encrypt_unencrypted_keys(conn: &PgConnection,
+                                    encryption_key: &core_keys::BuilderSecretEncryptionKey)
+                                    -> QueryResult<u32> {
+        use crate::schema::key::origin_private_encryption_keys::dsl::*;
+
+        let mut updated_rows = 0;
+        for row in origin_private_encryption_keys.for_update()
+                                                 .get_results::<Self>(conn)?
+        {
+            // If contents are not encrypted, then encrypt and update. The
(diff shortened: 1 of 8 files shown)


jeremymv2 (approved):
Working locally for me. This is really nice!

smacfarlane (approved):
Behaving as expected in my testing, thanks for this work @christophermaier!

components/builder-db/src/models/migration_support.rs
```diff
@@ -0,0 +1,48 @@
...
+    /// Returns the number of updated rows for user feedback purposes.
+    ///
+    /// Should be run in a transaction!
+    pub fn encrypt_unencrypted_keys(conn: &PgConnection,
+                                    encryption_key: &core_keys::BuilderSecretEncryptionKey)
+                                    -> QueryResult<u32> {
+        use crate::schema::key::origin_private_encryption_keys::dsl::*;
+
+        let mut updated_rows = 0;
+        for row in origin_private_encryption_keys.for_update()
```
markan, line 30: @smacfarlane and I were working through this, and wondering if a table lock might be better here. If I understand things correctly, for_update locks the rows included in the select, but doesn't lock new rows. In a world with multiple front ends, one could update and start this migration while the other remains on the previous version. If a new key is added and it goes to the old  FE, it could create a key that was unencrypted. In the short term it might break the new instance. In the long term I think it will resolve itself when the old instance takes the update.

markan (approved):
This looks really good to me. One minor question around the migration, but it's probably better than the previous migrations we've done for encryption.

components/builder-db/src/models/migration_support.rs
christophermaier: Yeah, the way the code is written, things would ultimately heal themselves, but it's probably safer to ensure that everything is done in a controlled manner.

christophermaier: After a bit more pondering and discussion, we're going to leave the code as is. While using a table lock would reduce the window of opportunity for the above scenario, it wouldn't actually close it entirely (an old frontend could theoretically insert a new key after the new frontend's migration completed, but before the old frontend updated itself). As such, the self-healing nature of this migration would still come into play.

In any case, the scenario above is a low probability one in our currently deployed systems. If this were to be more of an issue, we could investigate other ways to approach this. For example, we could allow the new frontend code to be able to cope with encrypted and non-encrypted data, or put encrypted data into a new table and have the new frontend code juggle data across both the old and new tables until the old frontends are decommissioned. It basically comes down to a cost-benefit analysis.

markan (changes requested):
Please hold on merging this, would like to chat as a team about sequencing and deployments beforehand.

markan (approved):
Hold removed, let's get this in.

PR #1521, opened 2020-09-18
[REFACTOR] Better error handling in API resources

christophermaier:


christophermaier:
Closing this out for now. It was intended as an example showing how to rewrite the endpoint logic more concisely, as well as how to ensure the database calls are asynchronous. I had to turn my attention to other things and wasn't able to get back to this. It can hang around in case someone is able to pick this up in the future.

PR #1522, opened 2020-09-18
Smma/graph manifest 1458

markan:
This adds capability for the old graph to generate graph manifests; this will let us use the old graph code on the new scheduler.

### DIFF
--- a/components/builder-graph/src/acyclic_package_graph.rs
+++ b/components/builder-graph/src/acyclic_package_graph.rs
@@ -15,18 +15,31 @@
 use petgraph::{algo::{connected_components,
                       is_cyclic_directed},
                graph::NodeIndex,
+               graphmap::DiGraphMap,
                Direction,
                Graph};
 use std::{cmp::Ordering,
           collections::{BinaryHeap,
-                        HashMap},
+                        HashMap,
+                        HashSet,
+                        VecDeque},
+          iter::FromIterator,
           str::FromStr};
 
+use habitat_core::package::PackageTarget;
+
 use crate::{acyclic_rdeps::rdeps,
+            data_store::Unbuildable,
+            error::Result,
             hab_core::package::PackageIdent,
+            package_build_manifest_graph::{PackageBuildManifest,
+                                           UnbuildableReason,
+                                           UnresolvedPackageIdent},
             package_graph_trait::{PackageGraphTrait,
                                   Stats},
-            protocol::originsrv};
+            package_ident_intern::PackageIdentIntern,
+            protocol::originsrv,
+            util::EdgeType};
 
 fn short_name(name: &str) -> String {
     let parts: Vec<&str> = name.split('/').collect();
@@ -34,13 +47,13 @@ fn short_name(name: &str) -> String {
     format!("{}/{}", parts[0], parts[1])
 }
 
-#[derive(Default)]
 pub struct AcyclicPackageGraph {
     package_max:   usize,
     package_map:   HashMap<String, (usize, NodeIndex)>,
     latest_map:    HashMap<String, PackageIdent>,
     package_names: Vec<String>,
     graph:         Graph<usize, usize>,
+    target:        PackageTarget,
 }
 
 impl PackageGraphTrait for AcyclicPackageGraph {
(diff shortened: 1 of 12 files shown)


components/builder-graph/src/acyclic_package_graph.rs
```diff
@@ -234,6 +246,153 @@ impl PackageGraphTrait for AcyclicPackageGraph {
...
+    //    Remove that package from the sub-graph
+    // 5) Walk the remaining sub-graph, collecting dependencies that are external to
+    //    the sub-graph.
+    // 6) Create a PackageBuildManifestGraph using the information generated in 1-5.
+    fn compute_build(&self,
+                     touched: &[PackageIdentIntern],
+                     unbuildable: &dyn Unbuildable,
+                     target: PackageTarget /* WRONG WRONG WRONG WRONG, WRONG WRONG WRONG
+                                            * WRONG */)
+                     -> PackageBuildManifest {
```
jeremymv2, line 269: I'm curious if you weighed the effort vs benefits of having `compute_build` instead return a `Result<PackageBuildManifest`? There's an `unwrap()`, an `.expect()` and two `panic!` calls in the code path of this function which leads me to think that it has the potential to be fallible and our future selves may be better served handling the failure as opposed to taking an outage if a service crashes.

components/builder-graph/src/acyclic_package_graph.rs
```diff
@@ -409,4 +568,185 @@ mod test {
...
         let (..) = graph.extend(&package2, true);
     }
+
+    fn make_package(ident: &str, deps: &[&str]) -> originsrv::OriginPackage {
+        let mut package = originsrv::OriginPackage::new();
+        package.set_ident(originsrv::OriginPackageIdent::from_str(ident).unwrap());
+
+        let mut package_deps = RepeatedField::new();
+        for dep in deps {
+            package_deps.push(originsrv::OriginPackageIdent::from_str(dep).unwrap());
```
jeremymv2, line 578: For your consideration: taking into account our future direction for validation of `__Ident`'s, we may also want to consider removing these `.unwrap()` calls here, utilize the `?` operator and return a `Result<originsrv::OriginPackage>`.

components/builder-graph/src/acyclic_package_graph.rs
jeremymv2: Also, it may prove useful to add a `TODO` around `PackageTarget` so that the uneducated reader can understand why this is so wrong :smiley:

components/builder-graph/src/cyclic_package_graph.rs
```diff
@@ -77,4 +79,14 @@ impl PackageGraphTrait for CyclicPackageGraph {
...
 
     fn stats(&self) -> Stats { self.graph.stats() }
+
+    fn compute_build(&self,
+                     touched: &[PackageIdentIntern],
+                     unbuildable: &dyn Unbuildable,
+                     _target: PackageTarget /* WRONG WRONG WRONG WRONG, WRONG WRONG WRONG
+                                             * WRONG */)
+                     -> PackageBuildManifest {
+        // TODO FIX assert!(_target == self.graph.current_target());
```
jeremymv2, line 89: Yeah, this would be nice to ensure always holds true.

components/builder-graph/src/graph_helpers.rs
```diff
@@ -443,23 +445,36 @@ pub fn transitive_deps(graph: &DiGraphMap<PackageIdentIntern, EdgeType>,
...
-                           unbuildable: &dyn Unbuildable,
-                           touched: &[PackageIdentIntern],
-                           origin: &str,
-                           target: PackageTarget)
-                           -> Vec<PackageIdentIntern> {
+pub fn compute_rebuild_set(
+    graph: &DiGraphMap<PackageIdentIntern, EdgeType>,
+    unbuildable: &dyn Unbuildable,
+    touched: &[PackageIdentIntern],
+    origin: Option<&str>,
```
jeremymv2, line 452: It might be nice to eventually leverage [Origin](https://github.com/habitat-sh/habitat/blob/4fa20d0742bf0cb09d43f9695aef2b60893162cb/components/core/src/origin.rs#L9-L23) here.

components/builder-graph/src/package_graph_target.rs
```diff
@@ -464,83 +471,131 @@ impl PackageGraphForTarget {
...
 
-        let mut file = File::create("latest_from_base.txt").expect("Failed to initialize file");
-        for (k, v) in &latest {
-            file.write_all(format!("{}: {}\n", &k, &v).as_bytes())
-                .unwrap();
-        }
-
-        let mut built: Vec<PackageBuild> = Vec::new();
+        let mut build_graph: DiGraphMap<UnresolvedPackageIdent, EdgeType> = DiGraphMap::new();
+        //        let mut built: Vec<PackageBuild> = Vec::new();
```
jeremymv2, line 533: We can probably remove this commented out variable as it appears to be a vestige of a previous development iteration.

jeremymv2 (commented):
@markan this is coming along really nicely! My only minor concerns are around error handling. I left some comments and am curious what the feasibility is of returning `Result` when computing the build.

components/builder-graph/src/acyclic_package_graph.rs
smacfarlane: > Also, it may prove useful to add a `TODO` around `PackageTarget` so that the uneducated reader can understand why this is problematic smiley

For context, this is a bit of left over from the previous implementation when target was added late.  The wrongness of it came from the fact that the graph has a target implicitly as part of it, but we still had to thread that value through in order to constrain queries to the database about the package. On our original pass through, we were hesitant to remove it, given it was in legacy code that we intend to remove in the near future.  However, as we revisited it for this review, we realized it would be a very small change to make the PackageTarget part of the struct. 

Fixed in 48ba35769a9de934379d165377e8cc9d0c708333

components/builder-graph/src/cyclic_package_graph.rs
smacfarlane: Fixed in 48ba35769a9de934379d165377e8cc9d0c708333

components/builder-graph/src/acyclic_package_graph.rs
```diff
@@ -234,10 +247,163 @@ impl PackageGraphTrait for AcyclicPackageGraph {
...
+    // 6) Create a PackageBuildManifestGraph using the information generated in 1-5.
+    fn compute_build(&self,
+                     touched: &[PackageIdentIntern],
+                     unbuildable: &dyn Unbuildable)
+                     -> Result<PackageBuildManifest> {
+        let mut rebuild_graph: DiGraphMap<PackageIdentIntern, EdgeType> = DiGraphMap::new();
+        let mut unbuildable_map: HashMap<PackageIdentIntern, UnbuildableReason> = HashMap::new();
+        let mut all_external_dependencies: HashSet<PackageIdentIntern> = HashSet::new();
+
+        let mut worklist: VecDeque<PackageIdentIntern> =
```
christophermaier, line 273: Is there a particular reason for using `VecDeque` over a plain old `Vec`? It looks like you only ever interact with one end of the data structure.

components/builder-graph/src/acyclic_package_graph.rs
```diff
@@ -234,10 +247,163 @@ impl PackageGraphTrait for AcyclicPackageGraph {
...
+        // Starting with our 'touched' set, walk the acyclic graph adding all
+        // reverse dependencies to a new sub-graph.
+        //
+        // Error case: A package is missing from the graph
+        //   * This package is flagged as missing in the unbuildables list.
+        while !worklist.is_empty() {
+            // Assumption: package is a short ident
+            let package: PackageIdentIntern = worklist.pop_back().unwrap(); // List is not empty
+
+            if let Some((_, node_index)) = self.package_map.get(&package.to_string()) {
```
christophermaier, line 286: Can `package_map` be re-keyed to use `PackageIdentIntern`? Seems like you'll be doing a lot of incidental `to_string` calls otherwise.

components/builder-graph/src/acyclic_package_graph.rs
```diff
@@ -234,10 +247,163 @@ impl PackageGraphTrait for AcyclicPackageGraph {
...
+                // We reverse the sense of dependency edges when building the graph, so 'outgoing'
+                // edges are really pointing to packages that depend on us
+                for neighbor in self.graph
+                                    .neighbors_directed(*node_index, Direction::Outgoing)
```
christophermaier, line 290: Not necessarily for this PR, but this kind of information seems like a good candidate for encapsulating in some kind of API around the graph itself.

components/builder-graph/src/acyclic_package_graph.rs
```diff
@@ -234,10 +247,163 @@ impl PackageGraphTrait for AcyclicPackageGraph {
...
+                {
+                    let short_ident =
+                        PackageIdentIntern::from_str(&self.package_names[neighbor.index()])?;
+
+                    if seen.insert(short_ident) {
+                        worklist.push_back(short_ident);
```
christophermaier, line 296: It looks like we're implicitly assuming that all these identifiers used throughout this code are short idents (i.e., `origin/package` resolution only). Is that true? If so, is that enforced in any way? Are there consequences if this assumption does not hold?

components/builder-graph/src/acyclic_package_graph.rs
```diff
@@ -234,10 +247,163 @@ impl PackageGraphTrait for AcyclicPackageGraph {
...
+        // Error case: A package is missing from the graph
+        //   * This package is flagged as missing in the unbuildables list.
+        while !worklist.is_empty() {
+            // Assumption: package is a short ident
+            let package: PackageIdentIntern = worklist.pop_back().unwrap(); // List is not empty
+
+            if let Some((_, node_index)) = self.package_map.get(&package.to_string()) {
+                // We reverse the sense of dependency edges when building the graph, so 'outgoing'
+                // edges are really pointing to packages that depend on us
+                for neighbor in self.graph
```
christophermaier, line 289: Might be good to call this "dependency" rather than the more general "neighbor" to help code readability.

components/builder-graph/src/acyclic_package_graph.rs
```diff
@@ -234,10 +247,163 @@ impl PackageGraphTrait for AcyclicPackageGraph {
...
+                                    .neighbors_directed(*node_index, Direction::Outgoing)
+                {
+                    let short_ident =
```
christophermaier, line 292: Similarly, naming this "dependency_ident" (or similar) would be better for readability... naming after its role / what it does, rather than what it concretely is.

components/builder-graph/src/acyclic_package_graph.rs
christophermaier: And reading further, perhaps "runtime_dependency" would be even better.

components/builder-graph/src/acyclic_package_graph.rs
```diff
@@ -234,10 +247,163 @@ impl PackageGraphTrait for AcyclicPackageGraph {
...
+                    }
+
+                    rebuild_graph.add_edge(short_ident, package, EdgeType::RuntimeDep);
+                }
+            } else {
+                // It's possible we've never seen this package because it's the first time it was
+                // built, and so never uploaded. That should only happen when we're explicitly
+                // rebuilding the package, e.g. it's in the touched set.
+                if touched.contains(&package) {
+                    rebuild_graph.add_node(package);
```
christophermaier, line 306: Are there any edges we need to add, or does that get handled somewhere else?

components/builder-graph/src/acyclic_package_graph.rs
```diff
@@ -234,10 +247,163 @@ impl PackageGraphTrait for AcyclicPackageGraph {
...
+        let flooded = if let Ok(unbuildables) =
+            unbuildable.filter_unbuildable(&rebuild_idents, self.target)
+        {
+            for package in &unbuildables {
+                unbuildable_map.entry(*package)
+                               .or_insert(UnbuildableReason::Direct);
+            }
+            crate::graph_helpers::flood_deps_in_origin(&rebuild_graph, &unbuildables, None)
+        } else {
+            warn!("Filter unbuildable returned an error?");
```
christophermaier, line 327: Should the error itself be logged?

christophermaier (commented):
Not done yet, but wanted to release the comments I had sooner rather than later.

jeremymv2 (approved):

christophermaier (approved):

PR #1531, opened 2020-09-28
Set HOME in job dispatch for local development

smacfarlane:
The local development environment uses the chroot studio for building packages in the worker.   This requires the HOME variable to be set. For production builder we clear the environment before creating the studio which includes clearing the HOME variable. This sets it to a hard-coded  path (the same as what is set by default in the config) only when we are in the development studio. 

The change here is a little awkward in shape,  but the job dispatch would likely need to be refactored to make supporting both the docker and chroot studio, along with windows builds, a little more ergonomic to set up. While this would be worthwhile taking on, it's a little out of scope for the scheduler work currently in flight, but we do need to be able to run the Worker in the studio as part of validating the scheduler changes.  

### DIFF
--- a/components/builder-worker/src/runner/studio.rs
+++ b/components/builder-worker/src/runner/studio.rs
@@ -148,7 +148,6 @@ impl<'a> Studio<'a> {
         cmd.stdout(Stdio::piped());
         cmd.stderr(Stdio::piped());
 
-
         cmd.arg("studio");
         cmd.arg("build");
 
--- a/components/builder-worker/src/runner/studio.rs
+++ b/components/builder-worker/src/runner/studio.rs
@@ -72,13 +72,11 @@ impl<'a> Studio<'a> {
     /// * If the calling thread can't wait on the child process
     /// * If the `LogPipe` fails to pipe output
     pub fn build(&self, streamer: &mut JobStreamer) -> Result<Child> {
-        let dev_mode = if let Some(_val) = env::var_os(DEV_MODE) {
+        let dev_mode = env::var_os(DEV_MODE).is_some();
+        if dev_mode {
             debug!("RUNNER_DEBUG_ENVVAR ({}) is set - using non-Docker studio",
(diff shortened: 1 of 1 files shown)


components/builder-worker/src/runner/studio.rs
```diff
@@ -80,6 +88,10 @@ impl<'a> Studio<'a> {
 
         let mut cmd = self.studio_command()?;
         cmd.current_dir(self.workspace.src());
+        if dev_mode && cfg!(not(windows)) {
+            cmd.env("HOME", "/hab/svc/builder-worker/data");
+        }
```
christophermaier, line 93: It would be useful for future maintainers if some of the discussion you wrote up in the PR message was included at this point as a comment. While someone could figure out why this was written like it is, it would require some Git archaeology; it's also not necessarily obvious from simply reading the code.

christophermaier (changes requested):

components/builder-worker/src/runner/studio.rs
```diff
@@ -72,6 +72,14 @@ impl<'a> Studio<'a> {
     /// * If the calling thread can't wait on the child process
     /// * If the `LogPipe` fails to pipe output
     pub fn build(&self, streamer: &mut JobStreamer) -> Result<Child> {
+        let dev_mode = if let Some(_val) = env::var_os(DEV_MODE) {
+            debug!("RUNNER_DEBUG_ENVVAR ({}) is set - using non-Docker studio",
+                   DEV_MODE);
+            true
+        } else {
+            false
+        };
```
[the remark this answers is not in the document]
smacfarlane: Addressed in f1935f1

components/builder-worker/src/runner/studio.rs
smacfarlane: Addressed in f1935f1

christophermaier (approved):

markan (approved):

PR #1532, opened 2020-09-30
Expose worker zmq poll interval as config object

smacfarlane:
The builder-worker currently blocks its main run loop for up to 60 seconds, or until a message from jobsrv arrives.  This has a side effect of blocking job completion until the polling times out, as the jobserv in normal operation won't send a message to a busy worker.   The blocking occurs because we use an mspc channel to signal back to the main loop that work has completed, which is only read after the zmq poll finishes. https://github.com/habitat-sh/builder/blob/master/components/builder-worker/src/runner/mod.rs#L715-L755

This PR exposes that value as a config object, with the default of its current value 60 seconds (converted to ms for zmq).  It also sets the default for the studio dev environment to 5 seconds, so that we can tighten up our feedback loop while iterating on the scheduler.   

A side effect of this change, is that our `user.toml` files we lay down as part of `start-builder` are now written to the `/hab/user/<svc>` path, getting rid of the deprecation warnings.

### DIFF
--- a/components/builder-worker/src/config.rs
+++ b/components/builder-worker/src/config.rs
@@ -18,25 +18,27 @@ pub type JobSrvCfg = Vec<JobSrvAddr>;
 #[serde(default)]
 pub struct Config {
     /// Enable automatic publishing for all builds by default
-    pub auto_publish:     bool,
+    pub auto_publish:       bool,
     /// Filepath where persistent application data is stored
-    pub data_path:        PathBuf,
+    pub data_path:          PathBuf,
     /// Location of Builder encryption keys
-    pub key_dir:          KeyCache,
+    pub key_dir:            KeyCache,
     /// Path to worker event logs
-    pub log_path:         PathBuf,
+    pub log_path:           PathBuf,
     /// Default channel name for Publish post-processor to use to determine which channel to
     /// publish artifacts to
-    pub bldr_channel:     ChannelIdent,
+    pub bldr_channel:       ChannelIdent,
     /// Default URL for Publish post-processor to use to determine which Builder to use
     /// for retrieving signing keys and publishing artifacts
-    pub bldr_url:         String,
+    pub bldr_url:           String,
     /// List of Job Servers to connect to
-    pub jobsrv:           JobSrvCfg,
-    pub features_enabled: String,
+    pub jobsrv:             JobSrvCfg,
+    pub features_enabled:   String,
     /// Github application id to use for private repo access
-    pub github:           GitHubCfg,
-    pub target:           PackageTarget,
+    pub github:             GitHubCfg,
+    pub target:             PackageTarget,
+    /// The frequency to poll the zmq socket for messages from jobsrv, in seconds
+    pub work_poll_interval: usize,
 }
 
 impl Config {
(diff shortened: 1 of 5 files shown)


christophermaier (changes requested):
For user-facing configuration value names, I find it useful to encode the type into the name in order to reduce confusion. Here, this would be `work_poll_interval_sec`. That way you never have to scratch your head and ask "was this one seconds or milliseconds?". This is particularly important when the configuration language doesn't let you represent a time period in a rich type like `Duration`. When you're basically limited to numbers, providing as much help to the user as you can is important.

Once I get into the actual developer-facing code, however, I usually like to get into rich types as soon as possible and practical. Here, when parsing the config, I'd immediately convert this to a `Duration` using seconds as the units.

Then, when you convert this to the milliseconds needed for ZMQ, you can just use the `as_millis` function, and leave out the manual multiplication. The function call also works as a nice bit of documentation here.

One potential bug you might want to head off is defining a lower limit on the polling interval. If the user were to enter `0`, then the poll would return immediately (http://api.zeromq.org/2-1:zmq-poll), which may not be what you actually want. Your lower limit could just be, say, 1. You may also want to define an upper limit, or perhaps issue some kind of warning if a user specifies a potentially problematic value (perhaps waiting for 3600 seconds is something they want to think twice about?)

Incidentally, all these concerns are why I often like to create custom types for parameters like this. You can centralize all this logic into a single place, rather than having it scattered across several places. It also makes it really easy to find exactly where the information is getting used (tracing a `usize` through the code is a little more tricky!)

smacfarlane:
> EDIT: The bit about a custom type is just my standard pitch for this kind of thing; you don't have to do it smile I think the other typing concerns, as well as bounds checking, are the more important bits of feedback.

I explored this for bit but decided to proceed with a simple Duration. What I was finding was that getting the de-serialization to work correctly would probably not be time well spent for something that is for development purposes only and we hope to rip out in the near future as we isolate and remove the zmq usage.  I've left a note to that effect in the code :smile: 

I decided the bounds checking could be `warn!`s, rather than any type of enforcement.  The documentation for zmq leads me to believe this will be safe (though perhaps unfriendly to your cpu fans)  for the `timeout=0` case:

>If none of the requested events have occurred on any zmq_pollitem_t item, zmq_poll() shall wait timeout microseconds for an event to occur on any of the requested items. If the value of timeout is 0, zmq_poll() shall return immediately.

christophermaier:
@smacfarlane I agree that it should be "safe" with a timeout of 0, but you're really gonna peg your CPU in that loop that way. I think a better way might be to just set the value to 1 (second) if the user gives you "0", and give a warning message to that effect (i.e. "You said '0', but trust me, '1' is gonna work out much better for you here").

@markan: here's a ping for your ZMQ know-how 😄

christophermaier (approved):

christophermaier:
tenor-254948610

components/builder-worker/src/config.rs
```diff
@@ -89,6 +95,31 @@ impl Default for JobSrvAddr {
...
+{
+    let duration: u64 = match serde::Deserialize::deserialize(duration)? {
+        0 => {
+            warn!("WorkerPollInterval is 0 seconds; zmq::poll will return immediately");
+            warn!("Setting to 1 second. Trust us, your cpu fans will thank you.");
+            1
+        }
+        d @ 1..=60 => d,
+        d => {
+            warn!("WorkerPollInterval is {} seconds; This may adversely impact job throughput",
```
markan, line 114: This is a short term fix, but if it weren't I explicitly tell them the interval is too long.

markan (approved):

PR #1533, opened 2020-09-30
Adds feature-flagged POC for kafka producer to builder api

eeyun:
This PR adds all the required bits for the initial piece of confluent cloud integration for the builder API hidden behind a feature flag. It only generates an event on individual package promotions. Getting this merged means we can start to investigate more interesting levels of kafka interaction in the SaaS services and serves as a model for event production and submission to confluent as an event bus. It should have no impact on live/acceptance whatsoever.

### DIFF
--- a/components/builder-api/src/config.rs
+++ b/components/builder-api/src/config.rs
@@ -17,7 +17,8 @@ use std::{env,
                 SocketAddr,
                 ToSocketAddrs},
           option::IntoIter,
-          path::PathBuf};
+          path::PathBuf,
+          time::Duration};
 
 pub trait GatewayCfg {
     /// Default number of worker threads to simultaneously handle HTTP requests.
@@ -31,7 +32,7 @@ pub trait GatewayCfg {
     fn listen_port(&self) -> u16;
 }
 
-#[derive(Clone, Debug, Deserialize)]
+#[derive(Clone, Debug, Default, Deserialize)]
 #[serde(default)]
 pub struct Config {
     pub api:         ApiCfg,
@@ -44,21 +45,7 @@ pub struct Config {
     pub memcache:    MemcacheCfg,
     pub jobsrv:      JobsrvCfg,
     pub datastore:   DataStoreCfg,
-}
-
-impl Default for Config {
-    fn default() -> Self {
-        Config { api:         ApiCfg::default(),
-                 artifactory: ArtifactoryCfg::default(),
-                 github:      GitHubCfg::default(),
-                 http:        HttpCfg::default(),
-                 oauth:       OAuth2Cfg::default(),
-                 s3:          S3Cfg::default(),
-                 ui:          UiCfg::default(),
-                 memcache:    MemcacheCfg::default(),
-                 jobsrv:      JobsrvCfg::default(),
-                 datastore:   DataStoreCfg::default(), }
-    }
+    pub kafka:       KafkaCfg,
 }
 
 #[derive(Debug)]
@@ -207,6 +194,28 @@ impl ToSocketAddrs for HttpCfg {
             IpAddr::V4(ref a) => (*a, self.port).to_socket_addrs(),
             IpAddr::V6(ref a) => (*a, self.port).to_socket_addrs(),
         }
+    }
+}
+
+#[derive(Clone, Debug, Deserialize)]
+#[serde(default)]
+pub struct KafkaCfg {
+    pub bootstrap_nodes:     Vec<String>,
+    pub client_id:           String,
+    pub connection_retry_ms: Duration,
+    pub message_timeout:     String,
+    pub api_key:             String,
+    pub api_secret_key:      String,
+}
+
+impl Default for KafkaCfg {
+    fn default() -> Self {
+        KafkaCfg { bootstrap_nodes:     vec![String::from("localhost:9092")],
+                   client_id:           String::from("http://localhost"),
+                   api_key:             String::from("CHANGEME"),
+                   api_secret_key:      String::from("CHANGEMETOO"),
+                   message_timeout:     String::from("3000"),
+                   connection_retry_ms: Duration::from_millis(300), }
     }
 }
 
--- a/components/builder-api/src/config.rs
+++ b/components/builder-api/src/config.rs
@@ -116,9 +116,7 @@ mod deserialize_into_vec {
         where D: Deserializer<'de>
     {
         let list = String::deserialize(deserializer)?;
-        let features = list.split(',')
-                           .map(|f| f.trim().to_uppercase().to_owned())
-                           .collect();
+        let features = list.split(',').map(|f| f.trim().to_uppercase()).collect();
         Ok(features)
     }
 }
@@ -440,7 +438,8 @@ mod tests {
         assert_eq!(config.api.build_targets.len(), 1);
         assert_eq!(config.api.build_targets[0], target::X86_64_LINUX);
 
-        assert_eq!(&config.api.features_enabled, "foo, bar");
+        assert_eq!(&config.api.features_enabled,
+                   &["foo".to_string(), "bar".to_string()]);
         assert_eq!(config.api.build_on_upload, false);
         assert_eq!(config.api.private_max_age, 400);
 
--- a/components/builder-api/src/config.rs
+++ b/components/builder-api/src/config.rs
@@ -200,22 +200,36 @@ impl ToSocketAddrs for HttpCfg {
 #[derive(Clone, Debug, Deserialize)]
 #[serde(default)]
 pub struct KafkaCfg {
-    pub bootstrap_nodes:     Vec<String>,
-    pub client_id:           String,
-    pub connection_retry_ms: Duration,
-    pub message_timeout:     String,
-    pub api_key:             String,
-    pub api_secret_key:      String,
+    pub bootstrap_nodes:        Vec<String>,
+    pub client_id:              String,
+    #[serde(with = "deserialize_into_duration")]
+    pub connection_retry_delay: Duration,
+    pub message_timeout:        String,
+    pub api_key:                String,
+    pub api_secret_key:         String,
 }
 
 impl Default for KafkaCfg {
     fn default() -> Self {
-        KafkaCfg { bootstrap_nodes:     vec![String::from("localhost:9092")],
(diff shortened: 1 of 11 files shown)


components/builder-api/src/config.rs
```diff
@@ -57,7 +58,8 @@ impl Default for Config {
                  ui:          UiCfg::default(),
                  memcache:    MemcacheCfg::default(),
                  jobsrv:      JobsrvCfg::default(),
-                 datastore:   DataStoreCfg::default(), }
+                 datastore:   DataStoreCfg::default(),
+                 kafka:       KafkaCfg::default(), }
```
christophermaier, line 62: FYI, since every field of the `Config` struct implements `Default`, you can just derive `Default` for `Config` and get rid of this entire implementation block.

components/builder-api/src/config.rs
```diff
@@ -210,6 +212,28 @@ impl ToSocketAddrs for HttpCfg {
     }
 }
 
+#[derive(Clone, Debug, Deserialize)]
+#[serde(default)]
+pub struct KafkaCfg {
+    pub bootstrap_nodes:     String, // "foo.bar.baz:9090,baz.fuzz.bizz:9191"
```
christophermaier, line 218: Maybe `Vec<String>` if it's possible to have more than one?

components/builder-api/src/config.rs
```diff
@@ -210,6 +212,28 @@ impl ToSocketAddrs for HttpCfg {
...
+    pub client_id:           String,
+    pub connection_retry_ms: u64,
+    pub message_timeout:     String,
```
christophermaier, line 221: For timeouts and retry limits, I've found `Duration` to be a more useful type, particularly because it encapsulates the time units within itself, so you don't have to have names like "foo_ms" and "bar_min".

components/builder-api/src/server/services/events.rs
```diff
@@ -0,0 +1,52 @@
...
+                                     .set("sasl.username", &config.api_key)
+                                     .set("sasl.password", &config.api_secret_key)
+                                     .set("security.protocol", "SASL_SSL")
+                                     .set("sasl.mechanisms", "PLAIN")
+                                     .create()
+            {
+                Ok(iprodr) => return KafkaProducer(iprodr),
+                Err(err) => error!("Error initializing kafka producer, will retry: {:?}", err),
+            }
+            thread::sleep(Duration::from_millis(config.connection_retry_ms));
```
christophermaier, line 35: It might be cleaner to implement `TryFrom<KafkaCfg> for KafkaProducer` here, and handle retries explicitly. That decouples the fundamental creation of the producer from the retry logic, rather than mixing them together.

Also, it would be good to have a limit on the number of retries; otherwise you could just spin indefinitely.

components/builder-api/src/server/mod.rs
```diff
@@ -65,7 +66,8 @@ features! {
         const Jobsrv = 0b0000_0010,
         const LegacyProject = 0b0000_0011,
         const Artifactory = 0b0000_0100,
-        const BuildDeps = 0b0000_1000
+        const BuildDeps = 0b0000_1000,
+        const EventBus = 0b0000_0110
```
smacfarlane, line 70: My understanding was features uses these as a bit mask. If that's the case wouldn't `EventBus == Jobsrv | Artifactory`?

components/builder-api/src/config.rs
eeyun: So this is something I hate in this implementation. However rdkafka-rust expects the bootstrap list as a single node. Which would mean unnecessarily doing some string parsing to put this into the shape of Vec<String> only to then reverse that work to get it back into this shape.

support/ci/builder-base-plan.sh
```diff
@@ -65,4 +65,13 @@ do_builder_prepare() {
   # Used to set the active package target for the binaries at build time
   export PLAN_PACKAGE_TARGET="$pkg_target"
   build_line "Setting PLAN_PACKAGE_TARGET=$PLAN_PACKAGE_TARGET"
+
+  # Used to allow librdkafka build scripts to execute successfully
+
+  if test -f /usr/bin/env; then
+    build_line "/usr/bin/env exists skipping symlink"
+  else
+    ln -s "/$(hab pkg path core/coreutils)/bin/env" /usr/bin/env
```
smacfarlane, line 74: I think we can remove leading `/` here :)

christophermaier (commented):
Cool to see some exploration of this!

Feel free to check out how we handled events in the Supervisor, in case there are any tricks you want to steal: https://github.com/habitat-sh/habitat/blob/master/components/sup/src/event.rs and https://github.com/habitat-sh/habitat/blob/master/components/sup/src/event :joy:

components/builder-api/src/server/services/events.rs
eeyun: That's a pretty great idea. When the feature is enabled i think we _do_ want to retry until a connection is made available.

components/builder-api/src/config.rs
eeyun: that's beautiful thanks!

components/builder-api/src/config.rs
christophermaier: The way I look at things like this is that we have our API (the thing we use every day, and that is part of our core business logic), and then we have the API of whatever libraries we happen to be using today. Ideally, our API would be as logical as we could make it; putting aside Kafka for a second, if any of us were creating any kind of data structure that called for more than one of something, I'm pretty sure we'd all reach for something like a `Vec` first. Inside of that API, we'd do whatever manipulations are necessary to interact with the libraries we're using. Today, we're using `rdkafka-rust`, which wants everything as a comma-delimited String. Tomorrow we might use something completely different that expects some other format (maybe something weird like a semicolon-delimited string :shrug: ). The take away here is that our business logic doesn't have to change if we keep a `Vec`, and you only have to tweak the translation layer between that API and the library, rather than having to tweak _everything_.

(As an aside, it's always nicer to encode logic and expectations as code, rather than comments.)

I readily acknowledge that it's super small stakes right now, but little decisions early on have a tendency to cause all kinds of issues down the road.

components/builder-api/src/server/services/events.rs
christophermaier: That could absolutely be a valid outcome of this. That particular decision would be really great to see explicitly handled in the service startup logic, though, rather than buried down here. That kind of thing is really in the domain of the overall service as a whole, rather than something fundamental to either of these types (and embedding that logic here strongly implies that it really is in the domain of the types).

components/builder-api/src/config.rs
eeyun: 100% agreement.

components/builder-api/src/server/mod.rs
eeyun: Well I might come across supremely ignorant here but: I'm not sure if features-rs is using these as a bitmask or not, I'll have to go read the macro. My understanding was that each entry in the features macro just needed a different integer expressed in binary notation. If these were equal then we would be seeing errors in the logs in live around artifactory failing to start/connect/etc.

eeyun: Having discussed a bit of this with @fnichol I think this should be fine. Also, I'm not running into any strange behavior in my functional tests when I toggle these options :)

fnichol: Yep, it's essentially a bitmask! The underlying implementation uses the [bitflags](https://docs.rs/bitflags) crate which tracks all these inside of a `usize` (I believe that's the packed primitive). So, lots and lots of potential features.

jeremymv2:
This is still a preview feature so I'm not going to add any official documentation on how to enable the `EventBus` feature. However, for the sake of reviewing the PR, to enable it add the following to `support/builder/config.sh` where we set up the `/hab/svc/builder-api/user.toml` file:

```
[api]
features_enabled = "jobsrv,eventbus"
```

christophermaier:
> [api]
> features_enabled = "jobsrv,eventbus"

Any reason why this couldn't be typed as a list / array? That is:

```toml
[api]
features_enabled = ["jobsrv", "eventbus"]
```

jeremymv2:
> Any reason why this couldn't be typed as a list / array?

Hmm, that's a great idea but could be tricky to implement since it's a breaking change. All the existing installations would have to modify their `user.toml` or whatever.

christophermaier:
@jeremymv2 Ah, good point. It should be possible to implement custom deserialization logic that could try parsing as a list first, and then falling back to a string, but that may not be something you want to delve into at the moment.

jeremymv2:
@christophermaier Yes, it turns out that I've discovered this may not be something easy to tackle in this PR and I may defer this change to another time. Especially since `api.features_enabled` is a preexisting API/type. 

```
    pub fn deserialize<'de, D>(deserializer: D) -> Result<Vec<String>, D::Error>
        where D: Deserializer<'de>
    {
        let features = match String::deserialize(deserializer) {
            Ok(list) => {
                list.split(',')
                    .map(|f| f.trim().to_uppercase().to_owned())
                    .collect()
            }
            Err(_) => {
                match Vec::deserialize(deserializer) {
                    Ok(list) => list,
                    Err(_) => vec![],
                }
            }
        };
        Ok(features)
```

```
115 |     pub fn deserialize<'de, D>(deserializer: D) -> Result<Vec<String>, D::Error>
|                                ------------ move occurs because `deserializer` has type `D`, which does not implement the `Copy` trait
...
118 |         let features = match String::deserialize(deserializer) {
|                                                  ------------ value moved here
...
125 |                 match Vec::deserialize(deserializer) {
|                                        ^^^^^^^^^^^^ value used here after move
|```

jeremymv2:
As an interim step, I've changed the underlying struct field type for `features_enabled` to a `Vec<String>` instead of `String` but still allow the preexisting `config.toml` format of `features_enabled = "jobsrv,eventbus"`. This is in https://github.com/habitat-sh/builder/pull/1533/commits/78fa3cf24749227c2481bf64e2c0c407c9155a3c

markan (approved):
This looks good to me. Let's get it merged.

smacfarlane (approved):

christophermaier:
tenor-207383283

PR #1538, opened 2020-10-20
Extend package build manifest to support existing build workers.

markan:
The build workers don't allow exact package resolution, but instead
resolve the latest from a channel. The execution ordering of the base
graph is specified only by the direct dependencies of the
package. This creates an antidependency (read before write) as it is
possible that we build the next iteration of a package before all of
the consumers of the last iteration have started; if that happens
those packages might pick up the wrong iteration. This is likely
harmless, except it makes the process + nondeterministic and hard to
debug. Constraining this will protect against this nondeterminism at
the cost of some parallelism.

Co-authored-by: Mark Anderson <mark@chef.io>
Co-authored-by: Scott Macfarlane <smacfarlane@chef.io>

### DIFF
--- a/components/builder-graph/src/graph_helpers.rs
+++ b/components/builder-graph/src/graph_helpers.rs
@@ -16,6 +16,8 @@ use petgraph::{algo::tarjan_scc,
                graphmap::DiGraphMap,
                Direction};
 
+use itertools::Itertools; // 0.8.0
+
 use std::{collections::{HashMap,
                         HashSet,
                         VecDeque},
@@ -204,7 +206,7 @@ pub fn dump_graph_raw(graph: &DiGraphMap<PackageIdentIntern, EdgeType>,
     let mut file = File::create(&path).unwrap();
 
     // iterate through nodes
-    for node in graph.nodes() {
+    for node in graph.nodes().sorted() {
         let (in_count, out_count) = count_edges(&graph, node);
         let orphaned = (in_count == 0) && (out_count == 0);
 
@@ -219,8 +221,15 @@ pub fn dump_graph_raw(graph: &DiGraphMap<PackageIdentIntern, EdgeType>,
                     EdgeType::BuildDep => bdeps.push(succ),
                     EdgeType::RuntimeDep => rdeps.push(succ),
                     EdgeType::StrongBuildDep => sdeps.push(succ),
+                    EdgeType::ExternalConstraint => {
+                        warn!("External Constraints should not appear here {:?}",
+                              edge_weight)
+                    }
                 }
             }
+            bdeps.sort();
+            rdeps.sort();
+            sdeps.sort();
             let bdeps_join = join_idents(",", &bdeps);
             let rdeps_join = join_idents(",", &rdeps);
             let sdeps_join = join_idents(",", &sdeps);
@@ -304,6 +313,7 @@ fn edgetype_to_abbv(edge: EdgeType) -> &'static str {
         EdgeType::RuntimeDep => "R",
         EdgeType::BuildDep => "B",
         EdgeType::StrongBuildDep => "S",
+        EdgeType::ExternalConstraint => "X",
     }
 }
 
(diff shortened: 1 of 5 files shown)


components/builder-graph/src/graph_helpers.rs
```diff
@@ -219,8 +221,14 @@ pub fn dump_graph_raw(graph: &DiGraphMap<PackageIdentIntern, EdgeType>,
                     EdgeType::BuildDep => bdeps.push(succ),
                     EdgeType::RuntimeDep => rdeps.push(succ),
                     EdgeType::StrongBuildDep => sdeps.push(succ),
+                    EdgeType::ExternalConstraint => {
+                        unimplemented!("External Constraints should not appear here")
```
jeremymv2, line 225: I'm curious if there is something we can do here other than insert a `panic` waiting to happen? For that reason, I don't think the pattern of utilizing `unimplemented!()`  for production code, when used in this particular manner, is going to be a good precedent to set.

Maybe we're better served if this returns an `Err` or is handled in some other non-fatal way.

From the docs:
```
Indicates unimplemented code by panicking with a message of "not implemented".

This allows your code to type-check, which is useful if you are prototyping or implementing a trait that requires multiple methods which you don't plan of using all of.
```

Where we've utilized `unimplemented!()` before was the latter case - for the trait pattern abstraction where we didn't plan on ever calling the functions but wanted the code to compile. This feels different to me since we're in a `match` and we cannot guarantee the variant won't be found, ultimately causing a service crash.

components/builder-graph/src/graph_helpers.rs
```diff
@@ -428,6 +437,9 @@ pub fn transitive_deps(graph: &DiGraphMap<PackageIdentIntern, EdgeType>,
                 EdgeType::RuntimeDep => true,
                 EdgeType::BuildDep => include_build_deps,
                 EdgeType::StrongBuildDep => include_build_deps,
+                EdgeType::ExternalConstraint => {
+                    unimplemented!("External Constraints should not appear here")
```
jeremymv2, line 441: Similar comment here as the comment for the previous `unimplemented!()`. Whether this remains, or becomes a warning message in the logs, it should probably have a unique message so we can distinguish and easily identify exactly where the `EdgeType::ExternalConstraint` violation happened.

components/builder-graph/src/graph_helpers.rs
jeremymv2: TIL: there's also a `todo!()` macro to handle prototyping https://doc.rust-lang.org/std/macro.todo.html

components/builder-graph/src/package_build_manifest_graph.rs
```diff
@@ -191,10 +201,59 @@ impl PackageBuildManifest {
...
+                    self.graph
+                        .neighbors_directed(prev_node, petgraph::EdgeDirection::Incoming)
+                        .collect();
+                for consumer in consumers {
+                    self.graph
+                        .add_edge(*node, consumer, EdgeType::ExternalConstraint);
+                }
+            }
+        }
+    }
```
jeremymv2, line 250: This is probably not an issue, however, I am curious about something.  Since we are "fixing up" the graph by adding dependencies, should we add any debugging messages to help identify how often and/or how many nodes we are adding? Otherwise, we may be making some assumptions about the graph content and to what extent our mutations are occurring. Intrigued if that would prove helpful or not.

components/builder-graph/src/graph_helpers.rs
markan: Unlike the graph dumping code above, if we hit this here we have a deeply, terribly broken graph. Ideally we'd represent this as a different type to try and forbid it totally, but due to the way we reuse some code across phases it would require some serious refactoring.

components/builder-graph/src/package_build_manifest_graph.rs
markan: Interesting thought. A nitpick/nuance is that technically we're not adding dependencies, we're making some implicit relationships explicit. Don't mind adding a logging message, but it either will be a largish number anytime we have a cycle, or zero otherwise.

components/builder-graph/src/graph_helpers.rs
jeremymv2: I see, it sounds like we actually want to crash here then because we're in an unknown state.

In that case, I think replacing `unimplemented!()` with `panic!()` is going to serve the reader better as we have no intention of eventually implementing something for this match branch arm.

markan: Agreed

jeremymv2 (approved):

christophermaier (commented):
I'm late to the party, but I had a couple observations... none are deal breakers, though!

PR #1539, opened 2020-10-26
Smma/worker scheduler 1523

markan:
This PR implements the new manifest based scheduler and some minor changes to the worker manager to accommodate it.

The previous scheduler used the database's view of the dependencies and was deeply intertwined with the database representation, and ignored the graph datastructure. As a result, it couldn't handle circular dependencies, and was impractical to retrofit. 

The new scheduler takes a in-memory representation of the graph (the manifest) and serializes it to the DB as entries in the job_graph table. These entries track each build item's dependencies and a waiting_on_count of dependencies remaining to be built.  It then uses stored procedures to atomically manage that count, and choose/take the next available build item to build.

This as implemented as a small actor style worker implemented as a Tokio async thread. The actual code is hopefully straightforward and simple. One significant change was to change the communication between the worker manager and the scheduler. The old system used a database based signaling mechanism to flag state changes of builds in progress. We chose to take a potential reliability/persistence hit in the interests of simplicity and make that a a direct message to the scheduler from the worker. 

All of this should be protected by a feature flag and be safe to commit.


jeremymv2:
@markan looking good! There are just a few very minor shellcheck/rustfmt items to make the build go green :green_heart:

components/builder-jobsrv/src/scheduler_datastore/test.rs
```diff
@@ -0,0 +1,358 @@
+#[cfg(test)]
+#[cfg(feature = "postgres_tests")]
+// cargo test --features postgres_tests to enable
+// from root
+// cargo test -p habitat_builder_jobsrv --features=postgres_tests
+// --manifest-path=components/builder-jobsrv/Cargo.toml
+mod test {
```
christophermaier, line 7: I think this will be `scheduler_datastore::test::test` because of how it's nested. You probably want to put the `test` and `feature` annotations on the declaration up in `scheduler_datastore`.

The test-only structs from `scheduler_datastore` could also just as well be defined in here.

Alternatively, all this could go into a `tests` module in `scheduler_datastore`... up to you :smile:

components/builder-jobsrv/src/scheduler_datastore/test.rs
```diff
@@ -0,0 +1,358 @@
...
+                hab_core::package::PackageTarget,
+                test_helpers::*};
+
+    use habitat_builder_db::{datastore_test,
+                             models::{jobs::{JobExecState,
+                                             JobGraphEntry,
+                                             JobStateCounts,
+                                             NewJobGraphEntry},
+                                      package::BuilderPackageTarget}};
+    use std::str::FromStr;
```
christophermaier, line 20: You could probably replace most of these with `use super::*;`, as in "normal" testing modules.

components/builder-jobsrv/src/scheduler_datastore/test.rs
```diff
@@ -0,0 +1,358 @@
...
+
+    #[test]
+    fn take_next_job_for_target() {
+        let target_platform =
+            BuilderPackageTarget(PackageTarget::from_str("x86_64-linux").unwrap());
+
+        let ds = datastore_test!(DataStore);
+        let conn = ds.get_pool().get_conn().unwrap();
+
+        let mut h = DbHelper::new(0, "x86_64-linux");
```
christophermaier, line 66: Might be useful for this to accept a `BuilderPackageTarget` as an argument, just to reduce duplication.

components/builder-jobsrv/src/scheduler_datastore/test.rs
```diff
@@ -0,0 +1,358 @@
...
+        let job_graph_entry = JobGraphEntry::create(&entry, &conn).unwrap();
+
+        assert_eq!(job_graph_entry.group_id, 0);
+        assert_eq!(job_graph_entry.job_state, JobExecState::Pending);
+    }
+
+    #[test]
+    fn take_next_job_for_target() {
+        let target_platform =
+            BuilderPackageTarget(PackageTarget::from_str("x86_64-linux").unwrap());
```
christophermaier, line 61: This might benefit from a `Default` implementation on `BuilderPackageTarget`, or (failing that), a small helper function to simplify the construction of it.

components/builder-jobsrv/src/scheduler_datastore/test.rs
```diff
@@ -0,0 +1,358 @@
...
+
+        let manifest = helpers::manifest_data_from_file();
+        for i in 1..2 {
+            make_job_graph_entries(i as i64,
+                                   JobExecState::Ready,
+                                   target_platform,
+                                   &manifest,
+                                   &conn);
+        }
+        // Test some stuff for real here
```
christophermaier, line 104: TODO?

components/builder-jobsrv/src/scheduler_datastore/test.rs
```diff
@@ -0,0 +1,358 @@
...
+        let ds = datastore_test!(DataStore);
+        let conn = ds.get_pool().get_conn().unwrap();
+
+        let _ = make_simple_graph_helper(0, &TARGET_PLATFORM, &conn);
+
+        let rdeps = JobGraphEntry::transitive_rdeps_for_id(0, &conn).unwrap();
+        assert_eq!(rdeps.len(), 0);
+
+        let mut rdeps = JobGraphEntry::transitive_rdeps_for_id(1, &conn).unwrap();
+        rdeps.sort();
```
christophermaier, line 148: Should `transitive_rdeps_for_id` return a sorted list as part of its contract?

components/builder-jobsrv/src/server/scheduler2.rs
```diff
@@ -0,0 +1,374 @@
...
+    last_message_debug: String, /* It would be cool to be able to do something like this:
+                                 * last_message:  Option<SchedulerMessage>,
+                                 * But the responders can't be copied so it's hard to keep the
+                                 * message around. */
+}
+
+type Responder<T> = oneshot::Sender<T>;
+
+#[derive(Debug)]
+#[allow(dead_code)] // TODO REMOVE
```
christophermaier, line 43: TODO

components/builder-jobsrv/src/server/scheduler.rs
```diff
@@ -404,6 +404,9 @@ impl ScheduleMgr {
         {
             // Check the deps for the project. If we don't find any dep that
             // is in our project list and needs to be built, we can dispatch the project.
+            // NOTE: get_ident().is_empty() is only true if the project has never been built
+            // Otherwise ident is going to be the FQPI of the latest ident (which channel is
+            // unclear)
```
christophermaier, line 409: Might be worth enshrining this in code directly, rather than just in a comment.

christophermaier (commented):
Haven't finished working through this, but wanted to get my comments thus far out for review.

components/builder-db/src/migrations/2020-08-XX_job_graph_experiment/up.sql
```diff
@@ -0,0 +1,248 @@
+CREATE TYPE job_exec_state AS ENUM (
```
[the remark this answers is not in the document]
markan: It hasn't caused any issues so far, but its worth renaming this to something more standard now we are close to merging.

components/builder-db/src/models/package.rs
```diff
@@ -954,7 +954,10 @@ impl Deref for BuilderPackageIdent {
          Clone,
          FromSqlRow,
          AsExpression,
-         PartialEq)]
+         PartialEq,
+         Hash,
+         Eq,
+         Copy)]
 #[sql_type = "Text"]
 pub struct BuilderPackageTarget(pub PackageTarget);
```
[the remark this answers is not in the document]
markan: Not against renaming. Probably should also rename BuilderPackageIdent; that might be best done as a separate PR.

components/builder-db/src/schema/jobs.rs
```diff
@@ -80,3 +80,23 @@ table! {
...
+        id -> BigInt,
+        group_id -> BigInt,
+        project_name -> Text,
+        job_id -> Nullable<BigInt>,
+        job_state -> JobExecStateMapping,
+        manifest_ident -> Text,
+        as_built_ident -> Nullable<Text>,
+        dependencies -> Array<BigInt>,
+        waiting_on_count -> Int4,
+        target_platform -> Text, // Should be enum
```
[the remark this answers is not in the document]
markan: Making it a postgres enum ends up being difficult to do without migrating most of the other users (in particular the  jobs table)
There's also some painful coupling elsewhere (the natural type to use would have to be wrapped). I'd like to take a separate pass at cleaning up this rather than enlarge this PR any further.

components/builder-jobsrv/src/server/mod.rs
```diff
@@ -183,34 +195,68 @@ pub async fn run(config: Config) -> Result<()> {
...
-      .bind(cfg.http.clone())
-      .unwrap()
-      .run()
-      .await
-      .map_err(Error::from)
+        // This is not what we want.  try_join! would be more appropriate so that we shut down if
+        // any of the things we're joining returns an error. However, the http_serv and
+        // scheduler_handle have different error types and I can't figure out how to resolve
+        // that scenario. We also need to handle any errors on the scheduler
+        let (http_res, _sched_res) = tokio::join!(http_serv, scheduler_handle);
```
christophermaier, line 237: :thinking: There may be some kind of wrapping / error mapping we can do to make this happen....

components/builder-db/src/models/jobs.rs
```diff
@@ -379,3 +431,356 @@ impl GroupProject {
...
+        j.wd = JobGraphEntry::count_by_state(gid, JobExecState::WaitingOnDependency, &conn)?;
+        j.rd = JobGraphEntry::count_by_state(gid, JobExecState::Ready, &conn)?;
+        j.rn = JobGraphEntry::count_by_state(gid, JobExecState::Running, &conn)?;
+        j.ct = JobGraphEntry::count_by_state(gid, JobExecState::Complete, &conn)?;
+        j.jf = JobGraphEntry::count_by_state(gid, JobExecState::JobFailed, &conn)?;
+        j.df = JobGraphEntry::count_by_state(gid, JobExecState::DependencyFailed, &conn)?;
+        j.cp = JobGraphEntry::count_by_state(gid, JobExecState::CancelPending, &conn)?;
+        j.cc = JobGraphEntry::count_by_state(gid, JobExecState::CancelComplete, &conn)?;
+        let duration_millis = start_time.elapsed().as_millis();
+        trace!("DBCall JobGraphEntry::count_all_states time: {} ms",
```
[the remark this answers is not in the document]
markan: I have the start of such a macro, but it's waiting for rust 1.47

components/builder-jobsrv/src/test_helpers.rs
```diff
@@ -0,0 +1,192 @@
...
+    h.add(conn,
+          "foo/ping/1.2.3/123",
+          &["foo/bar/1.2.3/123"],
+          JobExecState::WaitingOnDependency);
+    h.add(conn,
+          "foo/pong/1.2.3/123",
+          &["foo/baz/1.2.3/123", "foo/ping/1.2.3/123"],
+          JobExecState::WaitingOnDependency);
+
+    h
```
christophermaier, line 165: For these tests, I wonder if it would be worth creating some kind of file format (probably inspired by `dot` syntax) so that you could declare several graphs as readable test fixture files. (If it were based on `dot` syntax, you might even be able to generate visual representations of them, which would be fantastic for understanding!) Then your test helpers could just have general logic for reading these files and generating the code that represents them for tests.

components/builder-jobsrv/src/scheduler_datastore.rs
```diff
@@ -0,0 +1,229 @@
...
+    pub fn new_test() -> Self {
+        let data_store = datastore_test!(DataStore);
+        SchedulerDataStoreDb { data_store }
+    }
+
+    #[cfg(test)]
+    #[cfg(feature = "postgres_tests")]
+    pub fn get_connection_for_test(&self) -> PooledConnection<ConnectionManager<PgConnection>> {
+        self.get_connection()
+    }
```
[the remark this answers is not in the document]
markan: get_connection shouldn't have been public.

components/builder-jobsrv/src/scheduler_datastore/test.rs
markan: This is structured this way because we 1) wanted to to be in a separate file
and 2) wanted to have the cfg feature applied to it, which only seems to work right
directly in front of a mod declaration. Couldn't quite figure out how to get that to work without the double layering you see here.

components/builder-jobsrv/src/scheduler_datastore/test.rs
markan: This is fallout of the weird test::test nesting; I couldn't figure out how to use super::super::* or whatever to bring things in.

christophermaier: In `scheduler_datastore.rs` you'll want this:

```rust
#[cfg(test)]
#[cfg(feature = "postgres_tests")]
mod test;
```
Then, in `scheduler_datastore/test.rs`, you'll have this:

```rust
use super::*;

#[test]
fn do_a_thing() {
    // a thing
}
```
Only use the `super::*` if you need to import things you've already imported (or created) in `scheduler_datastore.rs`. It's exactly like you declared your test module in the `scheduler_datastore.rs` file.

components/builder-jobsrv/src/server/scheduler2.rs
```diff
@@ -0,0 +1,448 @@
...
+                 .finish() //  rework when finish_non_exhaustive is on main
+            }
+            _ => {
+                f.debug_struct("SchedulerMessage:: UNIMPLEMENTED")
+                 .field("Unknown", &"omitted".to_owned())
+                 .finish()
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
markan: A few thoughts. 
First of all, anything of type Responder isn't really very useful to render, at least as far as my current knowledge goes. It's basically a messy communications structure. 
I agree that the omitted stuff isn't really super useful, and for now I'll just drop them.

components/builder-jobsrv/src/server/handlers.rs
```diff
@@ -414,25 +440,153 @@ pub fn job_group_create(req: &RpcMessage, state: &AppState) -> Result<RpcMessage
...
+            info!("Including deps in Manifest");
+            // !(!msg.get_deps_only() || msg.get_package_only())
+            let _exclude_root = msg.get_deps_only() && !msg.get_package_only();
+
+            // TODO FIGURE OUT EXCLUDE ROOT
+            let target_graph = state.graph.read().unwrap();
+            let graph = target_graph.graph_for_target(target).unwrap();
+            let package = PackageIdentIntern::from_str(&project_name).unwrap(); // We created this, so it should never fail
+
+            let unbuildable = GraphDataStore::from_pool(state.db.clone())?;
```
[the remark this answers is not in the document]
markan: This is actually a graph_datastore which has a couple of traits one of which is the Unbuildable trait; that is what we actually use in compute_build. Once we get to the point where we can get rid of the old scheduler and clean up the datastore layer, we can eliminate this.

components/builder-jobsrv/src/server/scheduler.rs
markan: I didn't want to make any changes in the old scheduler code, but I did want to note this to clarify something I found confusing.

PR #1540, opened 2020-10-28
[REFACTOR] cloudevents API into builder-events

jeremymv2:
This removes the tight eventbus binding on Kafka and should allow us to arbitrarily switch out the event bus platform in the future, if needed. 


eeyun (approved):
This is a really good change. I'm happy to see this get a better abstraction. Though... I might suggest an experiment to see about replacing rdkafka-rs with vector

christophermaier:
> I might suggest an experiment to see about replacing rdkafka-rs with vector

@eeyun I see what you did there...

components/builder-events/src/cloud_events.rs
```diff
@@ -0,0 +1,68 @@
...
+        Ok(CloudEvent(event))
+    }
+
+    pub async fn maybe_emit(event_bus: &Option<EventBusConn>,
+                            event_key: i64,
+                            event_type: EventType,
+                            event_body: serde_json::Value) {
+        // We check if the EventBus feature is enabled and a Some variant
+        // If so, we use the underlying connection to emit an event. No-op otherwise.
+        if event_bus.as_ref().is_some() {
```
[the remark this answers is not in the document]
jeremymv2: Handled in https://github.com/habitat-sh/builder/pull/1540/commits/8b8d2f6df183f24ecb4ae6e8a7af777d1b86a2a1

components/builder-events/src/cloud_events.rs
```diff
@@ -10,7 +10,9 @@ use std::fmt;
 use url::Url;
 use uuid::Uuid;
 
-pub struct CloudEvent(pub Event);
+pub struct CloudEvent {
+    pub inner: Event,
+}
```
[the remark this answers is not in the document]
jeremymv2: Handled in https://github.com/habitat-sh/builder/pull/1540/commits/8b8d2f6df183f24ecb4ae6e8a7af777d1b86a2a1

components/builder-events/src/kafka.rs
```diff
@@ -43,11 +43,11 @@ impl TryFrom<&EventBusCfg> for KafkaProducer {
 
 #[allow(clippy::wildcard_in_or_patterns)]
 pub async fn produce(msgkey: i64, event: CloudEvent, conn: &EventBusConn) {
-    let topic = match event.0.get_type() {
+    let topic = match event.inner.get_type() {
         "PackageChannelMotion" | _ => KAFKA_DEFAULT_TOPIC_NAME,
```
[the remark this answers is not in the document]
jeremymv2: Handled in https://github.com/habitat-sh/builder/pull/1540/commits/8b8d2f6df183f24ecb4ae6e8a7af777d1b86a2a1

components/builder-events/src/connection.rs
```diff
@@ -8,6 +12,29 @@ pub enum Provider {
     Kafka,
 }
 
+impl Provider {
+    pub const KAFKA: &'static str = "kafka";
+}
```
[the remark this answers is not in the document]
jeremymv2: I added an `EventBusProvider` trait in https://github.com/habitat-sh/builder/pull/1540/commits/8b8d2f6df183f24ecb4ae6e8a7af777d1b86a2a1

components/builder-events/src/connection.rs
```diff
@@ -0,0 +1,59 @@
+use crate::kafka::KafkaProducer;
+use std::{convert::From,
+          time::Duration};
+
+#[derive(Clone, Debug, Deserialize)]
+#[non_exhaustive]
+pub enum Provider {
+    Kafka,
+}
```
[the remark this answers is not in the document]
jeremymv2: In https://github.com/habitat-sh/builder/pull/1540/commits/8b8d2f6df183f24ecb4ae6e8a7af777d1b86a2a1 I introduced a `EventBusClient` struct + `EventBusProvider` trait that I think handles your concerns. This enum now just exists for Deserialization and matching concerns when handling the config.

components/builder-events/src/connection.rs
```diff
@@ -0,0 +1,59 @@
...
+impl Default for EventBusCfg {
+    fn default() -> Self {
+        EventBusCfg { api_key:                String::from("CHANGEME"),
+                      api_secret_key:         String::from("CHANGEMETOO"),
+                      bootstrap_nodes:        vec![String::from("localhost:9092")],
+                      client_id:              String::from("http://localhost"),
+                      connection_retry_delay: Duration::from_secs(3),
+                      message_timeout_ms:     3000,
+                      provider:               Provider::Kafka, }
+    }
```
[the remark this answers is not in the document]
jeremymv2: I was thinking that this could be a blanket struct for handling _any_ event bus. Granted, it's Kafka oriented right now, but could be extended to support other message platforms. Should that scenario arise, the operator will merely set `provider = <the_new_provider>` in the `config.toml` and set whichever new fields are necessary  for that bus type.

components/builder-events/src/connection.rs
```diff
@@ -76,14 +78,34 @@ mod deserialize_into_duration {
...
+                }
+            }
+        }
     }
+
+    pub async fn publish(&self, event: BuilderEvent) { self.inner.publish(event).await; }
+}
+
+#[async_trait]
+pub trait EventBusProvider: Send {
```
christophermaier, line 109: I don't think you need the `Send` trait bound here.

components/builder-events/src/error.rs
```diff
@@ -29,6 +31,10 @@ impl From<EventBuilderError> for Error {
     fn from(err: EventBuilderError) -> Error { Error::CloudEventBuilderError(err) }
 }
 
+impl From<rdkafka::error::KafkaError> for Error {
+    fn from(err: rdkafka::error::KafkaError) -> Error { Error::KafkaError(err) }
+}
+
```
christophermaier, line 37: :thinking: To keep things general like we've been discussing implies that the error handling should be similarly general.

You could require that all your Event Bus implementors return an `EventBusError`, which could be defined generically in terms of `Box<dyn std::error::Error>`. This way, the implementors can create whatever kinds of event structure they want to, and as long as it implements `std::error::Error` (which it will), things will be fine. Importantly, _you_ won't have to keep adding new error variants here to accommodate new event backend implementations.

components/builder-events/src/kafka.rs
```diff
@@ -15,7 +16,11 @@ use rdkafka::{config::ClientConfig,
 pub const KAFKA_DEFAULT_TOPIC_NAME: &str = "builder_events";
 
 pub struct KafkaProducer {
-    pub inner: FutureProducer,
+    inner: FutureProducer,
+}
+
+impl KafkaProducer {
+    pub fn into_inner(self) -> FutureProducer { self.inner }
```
christophermaier, line 23: This doesn't appear to be used.

components/builder-events/src/kafka.rs
```diff
@@ -41,28 +46,28 @@ impl TryFrom<&EventBusCfg> for KafkaProducer {
...
 #[allow(clippy::wildcard_in_or_patterns)]
-pub async fn produce(msgkey: i64, event: CloudEvent, conn: &EventBusConn) {
-    let topic = match event.inner.get_type() {
-        "PackageChannelMotion" | _ => KAFKA_DEFAULT_TOPIC_NAME,
-    };
-    let message_record =
-        MessageRecord::from_event(event.inner).expect("error while serializing the event");
+impl EventBusProvider for KafkaProducer {
+    async fn publish(&self, event: BuilderEvent) {
+        let topic = match event.clone().into_inner().get_type() {
```
christophermaier, line 53: Why the call to `clone()` here?

components/builder-events/src/kafka.rs
```diff
@@ -41,28 +46,28 @@ impl TryFrom<&EventBusCfg> for KafkaProducer {
...
+            EventType::PACKAGECHANNELMOTION | _ => KAFKA_DEFAULT_TOPIC_NAME,
+        };
+        let routing_key = event.clone().routing_key();
```
christophermaier, line 56: And here?

components/builder-events/src/event.rs
```diff
@@ -0,0 +1,55 @@
...
+impl BuilderEvent {
+    pub fn new(event_type: EventType,
+               routing_key: Option<String>,
+               payload: serde_json::Value)
+               -> Result<Self, Error> {
+        let host = fqdn().unwrap_or_else(|| "localhost".to_string());
+        let source_url = Url::parse(&format!("https://{}", host))?;
+
+        let event = EventBuilderV10::new().id(format!("{}", Uuid::new_v4()))
+                                          .ty(format!("{}", event_type))
```
christophermaier, line 44: These bare `format!` calls imply you can just use `to_string()` calls directly, e.g. `Uuid::new_v4().to_string()`, etc.

components/builder-events/src/event.rs
```diff
@@ -0,0 +1,55 @@
...
+                                          .source(source_url)
+                                          .data("application/json", payload)
+                                          .build()?;
```
christophermaier, line 47: What's the scenario in which `build()` would fail? Since (I think?) we control all the inputs to this function, we may be able to replace `?` with `.expect("This should always work because we control all the inputs")` and simplify the signature of this `BuilderEvent::new()` function.

components/builder-events/src/event.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+impl BuilderEvent {
+    pub fn new(event_type: EventType,
+               routing_key: Option<String>,
+               payload: serde_json::Value)
+               -> Result<Self, Error> {
+        let host = fqdn().unwrap_or_else(|| "localhost".to_string());
+        let source_url = Url::parse(&format!("https://{}", host))?;
```
christophermaier, line 41: :thinking: This information will be the same for every single event creation run by the system. Is there a way that we can compute this once and cache it somewhere?

If we do that, we remove one source of fallibility in this call, which could allow us to simplify this function signature.

components/builder-events/src/event.rs
```diff
@@ -0,0 +1,55 @@
...
+#[derive(Clone)]
+pub struct BuilderEvent {
+    inner:       Event,
+    routing_key: Option<String>,
+}
+
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+#[non_exhaustive]
+pub enum EventType {
+    PackageChannelMotion,
```
christophermaier, line 19: Documentation on all of this would be great.

components/builder-api/src/server/resources/channels.rs
```diff
@@ -443,19 +444,10 @@ async fn promote_package(req: HttpRequest,
...
-                Ok(_) => {
-                    if let Some(bus) = &state.eventbus.as_ref() {
-                        let event_type = EventType::PackageChannelMotion;
-                        match BuilderEvent::new(event_type, None, json!(&auditevent)) {
-                             Ok(event) => bus.inner.publish(event).await,
-                             Err(err) => debug!("Failed to create EventBus event for {}: {}", event_type, err)
-                        }
-                    }
-                }
+                Ok(_) => publish_event!(&state, EventType::PackageChannelMotion, json!(&auditevent)),
```
christophermaier, line 447: :thinking: Is `publish_event` actually a function of `AppState`?

christophermaier: While you're at it with the macro, you could stuff the `json!` call _inside_ `publish_event!`, which would simplify this callsite.

components/builder-events/src/event.rs
```diff
@@ -53,3 +53,22 @@ impl BuilderEvent {
...
+            match BuilderEvent::new(event_type, $routing_key, $json) {
+                Ok(event) => bus.inner.send(event).await,
+                Err(err) => {
+                    debug!("Failed to create EventBus event for {}: {}",
+                           event_type, err)
+                }
+            }
+        }
+    };
+}
```
christophermaier, line 74: I think this could just be a plain old function call, rather than a macro.

components/builder-events/src/event.rs
```diff
@@ -53,3 +53,22 @@ impl BuilderEvent {
...
 }
+
+#[macro_export]
+macro_rules! publish_event {
+    ($state:expr, $type:expr, $json:expr) => {
+        publish_event!($state, $type, $json, None)
+    };
+    ($state:expr, $type:expr, $json:expr, $routing_key:expr) => {
+        if let Some(bus) = $state.eventbus.as_ref() {
+            let event_type = $type;
```
christophermaier, line 64: I'd just rename `$type` to `$event_type` in the parameter list and remove this line altogether.

components/builder-events/src/event.rs
```diff
@@ -53,3 +53,22 @@ impl BuilderEvent {
...
+            match BuilderEvent::new(event_type, $routing_key, $json) {
+                Ok(event) => bus.inner.send(event).await,
```
christophermaier, line 66: This `send` should be a method of `bus`, rather than requiring whatever's passed in to have a public `inner` member.

components/builder-api/src/server/mod.rs
```diff
@@ -102,26 +98,7 @@ impl AppState {
...
-                                thread::sleep(config.eventbus.connection_retry_delay);
-                                continue;
-                            }
-                        }
-                    }
-                }
-                // other messaging busses
-                _ => unimplemented!(),
-            }
+            app_state.eventbus = EventBusClient::new(&config.eventbus).ok();
```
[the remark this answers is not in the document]
jeremymv2: Added in https://github.com/habitat-sh/builder/pull/1540/commits/8f30d87eae34f9b5bce20b80a7ea1bf381a90714

components/builder-events/src/event.rs
jeremymv2: Modified in https://github.com/habitat-sh/builder/pull/1540/commits/dab949e9cc2965ca859a5ba57eb9c39bb158a1cb

components/builder-events/src/connection.rs
jeremymv2: Removed in https://github.com/habitat-sh/builder/pull/1540/commits/6734d5c6c3a2ccbc43cf30e84800011a359899c5

components/builder-api/src/server/resources/channels.rs
```diff
@@ -443,11 +443,18 @@ async fn promote_package(req: HttpRequest,
...
-                Ok(_) => {}
+                Ok(_) => {
+                    if let Some(bus) = &state.eventbus.as_ref() {
+                        let event_type = EventType::PackageChannelMotion;
+                        match BuilderEvent::new(event_type, None, json!(&auditevent)) {
+                             Ok(event) => bus.inner.publish(event).await,
+                             Err(err) => debug!("Failed to create EventBus event for {}: {}", event_type, err)
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
jeremymv2: Good catch! I've moved the event publishing calling site out of the audit block in https://github.com/habitat-sh/builder/pull/1540/commits/ff44d1a7bdc039886acb563df19faf66f08dbec3

I like your idea about the single logical transaction for audited events. I've created an issue for it here https://github.com/habitat-sh/builder/issues/1547

components/builder-events/src/event.rs
jeremymv2: Agreed. I handled this in https://github.com/habitat-sh/builder/pull/1540/commits/89ec48ec0e3da015458150f63ec464ee8a65a72c

components/builder-events/src/connection.rs
```diff
@@ -76,14 +78,34 @@ mod deserialize_into_duration {
     }
 }
 
-pub struct EventBusConn {
-    pub kafka:           Option<KafkaProducer>,
-    pub provider_in_use: Provider,
+pub struct EventBusClient {
+    pub inner:  Box<dyn EventBusProvider>,
+    pub config: EventBusCfg,
```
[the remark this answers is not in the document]
jeremymv2: `EventBusClient` dispensed with in https://github.com/habitat-sh/builder/pull/1540/commits/89ec48ec0e3da015458150f63ec464ee8a65a72c

components/builder-api/src/server/resources/channels.rs
jeremymv2: Agreed on both accounts. Handled in https://github.com/habitat-sh/builder/pull/1540/commits/172a3f954170aa79ba01ecca30f6117774a3e19f

components/builder-events/src/event.rs
jeremymv2: The reason I chose a macro is to simplify the calling sites when `routing_key` is not needed and is the `None` option variant.  In that case, the macro passes the `None` for you.

```macro_rules! publish_event {
    ($opt_bus:expr, $event_type:expr, $data:expr) => {
        publish_event!($opt_bus, $event_type, $data, with_routing_key: None <--- HERE )
    };
    ($opt_bus:expr, $event_type:expr, $data:expr,with_routing_key: $routing_key:expr) => {
        if let Some(bus) = $opt_bus.as_ref() {
            match BuilderEvent::new($event_type, $routing_key, json!($data)) {
                Ok(event) => bus.send(event).await,
                Err(err) => {
                    debug!("Failed to create EventBus event for {}: {}",
                           $event_type, err)
                }
            }
        }
    };
}
```

components/builder-events/src/event.rs
jeremymv2: I completely agree that documentation will be needed; and in the spirit of short exploratory development iterations and considering all this work is behind a feature flag, I'd like to defer adding documentation until a end-to-end producer->consumer flow is implemented. This will allow us to take any corrective actions and make discoveries validating or invaliding some of our current assumptions about how this architecture design will work, _then_ shore up the documentation :smiley:

components/builder-events/src/event.rs
christophermaier: Ah... I missed the `with_routing_key` bit in the match arm. 

My bigger point is that there's nothing here that _requires_ a macro to run (e.g., it's not generating any code based on what's being fed in). You could just as easily have `publish(bus, event_type, data)` and `publish_with_key(bus, event_type, data, key)` functions and achieve the same effect. If macros aren't required and there's not a strong ergonomics/usability argument for them, I think it's usually better from a readability perspective to use plain functions.

Also, this macro still leaks details at the call site. Things are fine when you're calling the non-routing key version, but when you call it with a routing key, you have to wrap it in a `Some`, even though the way you call the macro suggests that there's no optionality involved. That is, why would I (as a user, not as another function) want to call 

```rust
publish_event!(bus, type, data, with_routing_key: None)
```
?

Additionally, if you just want to pass in arbitrary objects that are destined to be turned into JSON, you can specify a `Serialize` trait bound on that parameter:

```rust
fn publish<D>(bus: ..., event_type: ..., data: D) 
    where D: Serialize
{
    # do stuff to turn data into JSON
}
```

If you do that, though, at some point in the future you might want to constrain things a bit more. As written, this will take _anything_ that implements `Serialize`, but you may (for peace of mind, comprehensibility, and general sanity preservation) want to tighten that up. You could instead imagine a world where `BuilderEvent` is a trait that has `Serialize` as a _supertrait_. In that world, the `publish` function would instead have a trait bound on `BuilderEvent` (at least conceptually; I know I'm changing things up here from what you currently have written). You still retain the ability to ensure that whatever you pass in can be turned into JSON, but you also specify that you don't want just _anything_, but something that is specifically one of your events.

components/builder-events/src/event.rs
christophermaier: For sure, it doesn't need to be absolutely complete documentation at this point. Keeping things light and agile at this point is a good idea.

Sometimes, however, I find it useful to add some documentation as I'm fleshing things out (maybe it'd be better to call them "notes" 😅). It can often help to clarify a design in my mind - I can't tell you how many times I've written some code, then started to describe what it's supposed to do in English, only to discover that the code is wrong, could be streamlined in some way, or is just plain unnecessary!

It can also be helpful as a rough guide for reviewers, too!

components/builder-api/src/server/resources/channels.rs
```diff
@@ -444,10 +444,12 @@ async fn promote_package(req: HttpRequest,
                 &auditevent,
                 &*conn,
             ) {
-                Ok(_) => publish_event!(&state, EventType::PackageChannelMotion, json!(&auditevent)),
+                Ok(_) => {}
```
christophermaier: This empty arm suggests that you could get away with something like this instead:

```rust
if let Err(e) = PackageChannelAudit::audit(&auditevent, &*conn) {
    debug!("Failed to save rank change to audit log: {}", e);
}
```

jeremymv2: Agreed. Handled in https://github.com/habitat-sh/builder/pull/1540/commits/9f5f1e41cf492cd955da60b6f382128239b3a073

components/builder-events/src/event.rs
jeremymv2: You've convinced me :smiley: I wrote some docs and added them in https://github.com/habitat-sh/builder/pull/1540/commits/7ac0699035e6f015f3b9863dfa72ef4035b9c3ca

components/builder-events/src/event.rs
jeremymv2: I've refactored the event creation and publishing API quite a bit in  https://github.com/habitat-sh/builder/pull/1540/commits/9f5f1e41cf492cd955da60b6f382128239b3a073 

`publish()` is now a method of `BuilderEvent` which feels to me a more elegant way to both create and publish the event in one line at the calling site:

```
BuilderEvent::new(EventType::PackageChannelMotion, NoKey, &auditevent)
      .publish(&state.eventproducer).await;
```

components/builder-events/src/connection.rs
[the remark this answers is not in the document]
jeremymv2: I have broken out the configuration into Bus-specific fields in the `EventBusConfig` struct in https://github.com/habitat-sh/builder/pull/1540/commits/9f5f1e41cf492cd955da60b6f382128239b3a073

This way we have specific name-spaced configuration that is 100% in alignment with the bus-type selected for use.

```
impl Default for Provider {
    fn default() -> Provider { Provider::None }
}

/// Holds the configuration defining the EventBus provider bindings to load as well
/// as the specific configuration required for each message bus binding type.
#[derive(Clone, Debug, Deserialize)]
#[serde(default)]
pub struct EventBusConfig {
    #[serde(with = "util::serde::string")]
    pub provider: Provider,
    pub kafka:    KafkaConfig,
}

impl Default for EventBusConfig {
    fn default() -> Self {
        EventBusConfig { provider: Provider::default(),
                         kafka:    KafkaConfig::default(), }
    }
}
```

We still have to have defaults for API key and API secret key to satisfy the `Deserialize` trait. However, it should be pretty obvious in the logs that either the API key or secret are incorrect and need to be configured for the user's message bus and we certainly can't know the correct values in advance.

christophermaier (approved):
I think this is looking really nice @jeremymv2! There will be more refinements to make, but those can be driven by need as this feature continues to grow. I think you've got a great foundation here.

PR #1544, opened 2020-11-05
Change default max json body size [jobsrv]

smacfarlane:
In some cases the manifest for a package can be larger than the maximum
default of 32k for the json body in rpc calls to jobsrv.  This was
previously corrected in #1366, but reappeared recently in our testing of
the scheduler.

Actix 2.0.0 added an `app_data` in addition to the `data` method that
makes the configuration available at the application level, rather than
the resource level.  Changing from the previously correct `data` to
`app_data` when setting the config value appears to again correctly set
the maximum payload size to 256K.

https://docs.rs/actix-web/2.0.0/actix_web/struct.Resource.html#method.app_data


christophermaier (approved):

markan (approved):
Nice detective work!

PR #1550, opened 2020-11-24
replace String return type for EventConsumer

jeremymv2:
This change improves the `EventConsumer` trait's `.poll()` method by returning a `BuilderEvent` instead of `String` type. The `BuilderEvent` type now becomes the contract message implementation for the `EventConsumer` and `EventPublisher` APIs. The `BuilderEvent` struct includes an `event_type` field that we can match on to flexibly deserialize the JSON `payload` field into a downstream Type of our choosing. This should allow us to handle any sort of business logic we need with these libraries.

### DIFF
--- a/components/builder-events/src/kafka.rs
+++ b/components/builder-events/src/kafka.rs
@@ -150,11 +150,10 @@ impl EventPublisher for KafkaPublisher {
         let message_record =
             MessageRecord::from_event(event).expect("error while serializing the event");
         match tracking {
-            Undelivered(ticket) => {
-                let topic = ticket.destination;
+            Undelivered(tag) => {
                 let future_record = {
-                    let r = FutureRecord::to(&topic).message_record(&message_record);
-                    if let AffinityKey::Key(key) = &ticket.affinity_key {
+                    let r = FutureRecord::to(&tag.destination).message_record(&message_record);
+                    if let AffinityKey::Key(key) = &tag.affinity_key {
                         r.key(key)
                     } else {
                         r
@@ -165,7 +164,10 @@ impl EventPublisher for KafkaPublisher {
                         error!("KafkaPublisher failed to send message to a Broker: {:?}",
                                err)
                     }
-                    Ok(_) => trace!("KafkaPublisher published event to topic: {}", topic),
+                    Ok(_) => {
+                        trace!("KafkaPublisher published event to topic: {}",
+                               tag.destination)
+                    }
                 };
             }
             Delivered => error!("KafkaPublisher will not publish an already delivered Event."),
--- a/components/builder-events/src/kafka.rs
+++ b/components/builder-events/src/kafka.rs
@@ -160,11 +160,12 @@ impl EventPublisher for KafkaPublisher {
                         r
                     }
                 };
-                if let Err(err) = self.inner.send(future_record, 0).await {
-                    error!("KafkaPublisher failed to send message to a Broker: {:?}",
-                           err)
-                } else {
-                    trace!("KafkaPublisher published event to topic: {}", topic);
+                match self.inner.send(future_record, 0).await {
+                    Err(err) => {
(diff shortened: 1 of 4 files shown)


components/builder-events/src/kafka.rs
```diff
@@ -120,22 +120,18 @@ impl EventConsumer for KafkaConsumer {
...
-                    match borrowed_message.payload_view::<str>() {
-                        Some(Ok(payload)) => Some(Ok(payload.to_string())),
-                        Some(Err(err)) => Some(Err(Error::EventError(Box::new(err)))),
-                        None => {
-                            let err: Box<dyn std::error::Error> = "No Payload".to_owned().into();
-                            Some(Err(Error::EventError(err)))
-                        }
+                Ok(message) => {
+                    match message.to_event() {
+                        Ok(event) => Some(Ok(event.into())),
```
christophermaier, line 128: The `From` implementation being used here sets the `routing_key` to `NoKey`, which implies some kind of lossy conversion. If this is on the consumer side, and the routing key is genuinely never needed (which sounds plausible), that suggests that a routing key isn't properly part of a `BuilderEvent`, but is something that is (optionally) needed for _sending_ the event.

If that's an accurate reading of the situation, it's somewhat analogous to saying that your mailing address is a fundamental part of the book you just ordered from Amazon. The book is its own independent thing, and your address is part of the process that brought it to you. That in turn suggests that perhaps there's some kind of "envelope" abstraction lurking in here. Maybe `BuilderEvent` is just a newtype wrapping a CloudEvent, and the "envelope" is a thing that looks like the current `BuilderEvent` (i.e., composes an event with an optional routing key). Then, when we're here on the receiving side of things, we can just take the event and make our `BuilderEvent` from it, and there's no routing key anywhere around.

markan: Interesting. I had understood the key to be more like a topic; e.g. a description of the content in some fashion. I figured it would be added later as the work evolved.

markan (commented):
Delighted to see the added typing work.

components/builder-events/src/kafka.rs
jeremymv2: Great feed back @markan and @christophermaier 

Please take a look at https://github.com/habitat-sh/builder/pull/1550/commits/b3e841d3505c1edcc94ffb2cd167debec1d5089e where I think I address your individual discussion points.

jeremymv2: @markan @christophermaier in https://github.com/habitat-sh/builder/pull/1550/commits/7ad8a55d53fd834c81221a9bb975d500dc8a7ab4 I've subtly changed the analogy a bit from a postal worker delivering an "envelope" to a shipper delivering a package and providing a "tracking" status. To me this seems to more accurately maintain the fidelity of what is actually occurring.

Instead of the consumer unwrapping an envelope to get to the `Event`, the message bus already handles that for us. From a consumer point of view, we don't really want to bother ourselves with unwrapping an envelope; we just want to get at the `Event` directly. When the `Event` is received it's implicit that it's been delivered. We therefore set its `tracking` state to `Delivered` in the `From` trait when transforming an `Event` into a `BuilderEvent` in the `.poll()` method.

Whenever a `BuilderEvent` object exists and is observable, we can determine by its `tracking` field if it is yet to be delivered, along with the delivery information, or alternatively, if it has already been received.

christophermaier: @jeremymv2 That ends up being pretty nice! I like the explicitness of it, which should make it much easier to both comprehend and update the code in the future.

christophermaier (approved):
Nice work @jeremymv2

components/builder-events/src/kafka.rs
jeremymv2: @christophermaier Why thank you, good Sir!  :bowing_man:

PR #1555, opened 2020-12-18
use remove_dir_all crate to clean source tree on worker

mwrock:
This fixes the Windows workers inability to clean the cloned source directory tree.


christophermaier (approved):
👍

PR #1573, opened 2021-01-18
Psajja/consolidate db

sajjaphani:
There are two database pool instances are defined.
  1. ~/builder/components/builder-db/src/pool.rs
  2. ~/builder/components/builder-db/src/diesel_pool.rs

Migrated the connection pool usage from pool defined in **pool.rs** to one defined in **diesel_pool.rs**. Further used **sql_query** function to execute raw sql queries. Also updated the corresponding errors.


components/builder-jobsrv/src/data_store.rs
```diff
@@ -43,10 +51,100 @@ use crate::error::{Error,
 
 mod test;
 
+// Holds the intermediate state of the 'jobs'
```
mwrock, line 54: Should we be using the builder-db models for these types? Perhaps even get rid of most/all of this raw sql by leveraging the diesel APIs.

mwrock: Looking at the functions that this raw sql is calling, I think the vast majority of them can be discarded in favor of using the diesel models directly or adding functions to the model that reproduce what the function does. Examples are:

```
CREATE OR REPLACE FUNCTION sync_jobs_v2() RETURNS SETOF jobs
    LANGUAGE sql STABLE
    AS $$
  SELECT * FROM jobs WHERE (scheduler_sync = false) OR (sync_count > 0);
$$;

CREATE OR REPLACE FUNCTION find_group_project_v1(gid bigint, name text) RETURNS SETOF group_projects
    LANGUAGE plpgsql STABLE
    AS $$
BEGIN
  RETURN QUERY SELECT * FROM group_projects WHERE owner_id = gid AND project_name = name;
  RETURN;
END
$$;
```

I do see a couple functions where we would likely need to keep the function but it looks like diesel provides a way to map to funcrions. See https://github.com/habitat-sh/builder/blob/master/components/builder-db/src/functions/jobs.rs for some examples. These seem to be clear straight forward in cases where the function returns `VOID` or a single value. However, I'm not sure how to set it up for functions that return a dataset. The only function I see that could be tough to map (with my limited diesel knowledge) is `next_pending_job_v2`.

markan:
Two quick thoughts (sorry I missed this earlier)
1) Have you looked at the code/tests in  eeyun/pg_derives_v2? There are overlapping concerns, and some of the tests may be useful.
2) Not sure what the current plan is, but much of the code in data_store.rs is unneeded once the new scheduler is in place.

components/builder-db/src/test.rs
```diff
@@ -75,35 +75,35 @@ macro_rules! datastore_test {
...
+                                            .execute(&conn)
+                                            .expect("Failed to create template database");
+
+            // Now that the database is recreated, set config to use that database
+            config.database = db_template.to_string();
+
+            let diesel_pool = DbPool::new(&config);
+
+            // Run any builder-db migrations
+            let _ = habitat_builder_db::migration::setup(&diesel_pool.get_conn().unwrap());
```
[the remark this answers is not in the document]
sajjaphani: The pool and the connection have been reused.

mwrock (commented):
Still have a little more to look at but wanted to get these questions/comments out for now.

components/builder-db/src/schema/jobs.rs
```diff
@@ -34,9 +34,9 @@ table! {
         id -> BigInt,
         group_state -> Text,
         project_name -> Text,
-        target -> Text,
```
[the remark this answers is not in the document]
sajjaphani: The order is necessary as we are deserializing from raw bytes:

https://github.com/sajjaphani/builder/blob/a3770cd71267cb7e2cb8be98db2706ec639f34fc/components/builder-db/src/models/jobs.rs#L489

components/builder-jobsrv/src/data_store.rs
```diff
@@ -345,48 +343,46 @@ impl DataStore {
...
     pub fn cancel_job_group(&self, group_id: u64) -> Result<()> {
-        let conn = self.pool.get()?;
-        conn.query("SELECT cancel_group_v1($1)", &[&(group_id as i64)])
-            .map_err(Error::JobGroupCancel)?;
+        debug!("DataStore: cancel_job_group");
+        let conn = self.diesel_pool.get_conn()?;
+        GroupProject::cancel_group_project(group_id as i64, &conn).map_err(Error::JobGroupCancel)?;
+        Group::cancel_job_group(group_id as i64, &conn).map_err(Error::JobGroupCancel)?;
 
         Ok(())
```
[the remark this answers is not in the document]
sajjaphani: Wrapped inside transaction

components/builder-jobsrv/src/migrations/2021-02-09-113251_jobsrv_diesel_migrate/up.sql
```diff
@@ -0,0 +1,58 @@
...
+--
+-- See http://diesel.rs/guides/all-about-inserts/
+-- https://github.com/diesel-rs/diesel/blob/master/CHANGELOG.md#added-21
+--
+SELECT diesel_manage_updated_at('groups');
+SELECT diesel_manage_updated_at('group_projects');
+SELECT diesel_manage_updated_at('jobs');
+SELECT diesel_manage_updated_at('busy_workers');
+-- SELECT diesel_manage_updated_at('origin_packages');
+-- SELECT diesel_manage_updated_at('audit_jobs');
```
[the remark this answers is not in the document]
sajjaphani: removed unused selects

components/builder-db/src/functions/jobs.rs
```diff
@@ -25,3 +28,16 @@ sql_function! {
...
+}
+
+sql_function! {
+  fn insert_group_v3(root_project: Text, project_names: Array<Text>, project_idents: Array<Text>, p_target: Text)
+  -> jobs::GroupRecord;
+}
+
+sql_function! {
+  fn pending_groups_v1(count: Integer) -> Array<jobs::GroupRecord>;
+}
```
mwrock, line 43: Your tansaction commit gave me the idea that you could convert these sql functions to rust functions by mimicking the logic inside a transaction and then eliminate the `Record` types.

sajjaphani: Even though we may convert these functions to use models, there could be other problems in doing.
One particular I could think of doing updates row-by-row can be extremely slow.

We may differ this conversion (all the sql functions) until we find it does not impose any performance penalty, preferably in a separate PR.

mwrock: yeah that is certainly a good concern. I'm fine with keeping the functions if you think that is best.

mwrock (approved):

mwrock:
@sajjaphani can you squash your commits? Then feel free to merge on your own.

PR #1574, opened 2021-02-05
minor fixes for functional tests

mwrock:
This seems to resolve 2 issues I was having running `test-builder`.

1. If you start the supervisor with `RUST_LOG` set to `debug`, there is a good chance that `hab-sup --version` will emit some debug output after the version which breaks the launcher's parsing logic. This is really a bug in the launcher that will be addressed separately and then we can add this back.
2. I noticed that the `mocha` test output was not printing to the screen which made me wonder if the tests were running at all. Oddly, the build-kite functional test runs to emit this output but I see nothing in a studio. Pointing to the path of the `mocha` binary seems to fix this.

### DIFF
--- a/.studiorc
+++ b/.studiorc
@@ -316,7 +316,7 @@ test-builder() {
   fi
 
   echo "Starting supervisor in test mode. Logs saved to ${logs}"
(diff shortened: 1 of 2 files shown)


mwrock:
closing for now. Point 1 above was fixed separately in the launcher and point 2 seems environmental.

PR #1603, opened 2021-08-05
Add makefile target for installing MacOS prerequisite packages

pozsgaic:
Makefile target can be invoked with:
$ make prereq
Target will refresh apt-get, and install packages preventing clean installation on MacOS workstation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,11 @@ ifeq ($(UNAME_S),Darwin)
 endif
 .PHONY: linux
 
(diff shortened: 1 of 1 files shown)


mwrock (approved):

mwrock:
Looks good to me but we do need to have the DCO cleared. What I would do is squash the three commits to 1 and add the `-s` in the commit which should get us over the DCO hump.

Issue #1615, opened 2021-08-27
builder not returning all releases for a version


mwrock:
This was noticed by a customer who has a package version with 132 releases. It can also be observed for `core/7zip`. Notice that `https://bldr.habitat.sh/v1/depot/pkgs/core/7zip/16.02` returns 24 releases:

```
{
range_start: 0,
range_end: 23,
total_count: 24,
data: [
{
```
However, the UI states that the total is 30:
![image](https://user-images.githubusercontent.com/655165/131176713-23bd9416-21ef-43bf-bd72-e76b4d37b5df.png)

Quwrying the DB `select ident from origin_packages where origin='core' and name='7zip' order by ident;` also returns 30 packages. So there are indeed 30 but the 6 oldest packages are missing from the UI and the rest api.

y-me-y:
Just as a side note this query returns 30 packages in the results https://bldr.habitat.sh/v1/depot/channels/core/unstable/pkgs/7zip/16.02

sajjaphani:
For version 16.04, on clicking the Linux node also shows the same Windows packages.
<img width="1127" alt="Screenshot 2021-09-23 at 3 54 35 PM" src="https://user-images.githubusercontent.com/2322779/134491723-fbd17751-c798-4ac9-a60e-700a4c24932a.png">

mwrock:
fixed by #1627

PR #1627, opened 2021-10-06
Fix package version details

sajjaphani:
After expanding one of the package versions (https://bldr.habitat.sh/#/pkgs/core/7zip -> VERSIONS), the expanded list contains fewer nodes that are shown in the summary of that particular version. This is happening because we are fetching 50 database entries from the database and then eliminating the duplicates. This results in returning fewer entries. We also do not handle pagination in this case.

Also, there is a problem in rendering the returned results in the UI. 
We should render each package (ident/platform) exactly once.
We should only render the corresponding platform in each expanded node.

More details:
https://github.com/habitat-sh/builder/issues/1615


components/builder-api/src/server/resources/pkgs.rs
```diff
@@ -176,7 +177,7 @@ fn get_packages_for_origin(req: HttpRequest,
     let origin = path.into_inner();
     let ident = PackageIdent::new(origin, String::from(""), None, None);
 
-    match do_get_packages(&req, &ident, &pagination) {
+    match do_get_packages(&req, &ident, &pagination, false) {
```
mwrock, line 180: rather than adding a `fetchall` param. Should we assume all packages if `range` == `-1`?

mwrock (commented):
I'm probably missing something but it looks like this addresses the case where there are more then 50 revisions. How would this fix https://bldr.habitat.sh/v1/depot/pkgs/core/7zip/16.02 where it has only 30 packages but just 24 are being returned.

PR #1656, opened 2021-11-09
Handle package name mismatch in connection creation

### LINKED ISSUE #1602: Habitat Builder fails to save with no meaningful feedback when specified package name does not match plan.sh
### `hab` CLI version
[Version of the habitat binary where you are encountering the issue]
1.6.351

### Platform Details
[Operating system distribution and release version. Cloud provider if running in the cloud]
MacOS Big Sur 11.4

### Scenario:
[What you are trying to achieve and you can't?]
Create a new package and associate a habitat/plan.sh with it

### Steps to Reproduce:
[If you are filing an issue what are the things we need to do in order to repro your problem? How are you using the CLI or Builder UI?]
1). Clone/fork a sample Chef github project
2)  Generate Habitat plan for project 
3)  Commit/push changes to github
4)  Go to Habitat Builder and create a package with a name that does not match the github project name
5). Go to Settings and generate a plan file specifying the plan.sh file that was generated in step 2
6).  Specify other settings for Automated Rebuild and Container (optional)
7). Try to Save the plan.  Plan will appear to save but the save does not persist the plan.  Refresh screen and observe that no plan is selected.
8). Repeat step 5 and Save the plan. 

### Expected Result:
[What are you expecting to happen as the consequence of above reproduction steps?]
The plan would save successfully or the Builder would provide a reason why the save failed

### Actual Result:
[What actually happens after the reproduction steps? Include the error output or a link to a gist if possible.]
Save fails without meaningful feedback, only the message "Failed to save plan connection".

sajjaphani:
closes #1602

When creating a new connection, if the package name does not match with the name in the plan file, it creates a record in the projects table but subsequently fails.

Upon updating the connection, it displays HTTP status text, which does not say what went wrong.

Along with handling the scenarios mentioned on the backend, this PR also handles displaying the message on UI.

### DIFF
--- a/components/builder-api/src/server/resources/projects.rs
+++ b/components/builder-api/src/server/resources/projects.rs
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Chef Software Inc. and/or applicable contributors
+// Copyright (c) 2018-2021 Chef Software Inc. and/or applicable contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -15,7 +15,8 @@
 use std::{env,
           str::FromStr};
 
-use actix_web::{http::{self,
+use actix_web::{body::Body,
+                http::{self,
                        StatusCode},
                 web::{self,
                       Data,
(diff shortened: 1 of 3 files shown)


sajjaphani:
Screenshot 2021-11-09 at 3 29 22 PM
Updated failure message.

mwrock (approved):

PR #1671, opened 2021-12-03
Event log fixes on the SaaS events page

sajjaphani:
Case 1: When the date filter changed on the Events (SaaS), results in console errors.
Fixes the dropdown change on the Events (SaaS) page.

Case 2: When the user logged in and tried to access Events (SaaS), results in console errors.
This is occurring because the user is logged into a different environment (acceptance/on-prem) and the auth same token is used for the production instance. Currently, we can only show events that are publicly accessible. 

We need to handle the login case separately. Below are the possible options:
1. Use separate auth credentials when accessing Events (SaaS) - may not be the right choice.
2. Support multi-login (on-prem and SaaS), use the token based on the context.
3. Any other options?

### DIFF
--- a/components/builder-web/app/client/depot-api.ts
+++ b/components/builder-web/app/client/depot-api.ts
@@ -313,10 +313,10 @@ export function getEvents(nextRange: number = 0, fromDate: string, toDate: strin
 }
 
 export function getSaasEvents(nextRange: number = 0, fromDate: string, toDate: string, query: string = '') {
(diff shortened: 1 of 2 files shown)


mwrock:
Perhaps another option (which is what we are currently doing) is to simply not authenticate to SAAS at all and only consume public saas events. My guess is that an on-prem user is not going to keep their ptivate packages on prem and only be interested in public (like core) packages on saas.

I think the SAAS experience would be to have the `enable_builder_events_saas` be `false` and have all of its own events come to the `Events` view.

mwrock (approved):

PR #1681, opened 2021-12-14
turn off proxy request buffering to allow for better upload streaming and fix super large uploads

### LINKED ISSUE #1663: 422 (processable entity) failures returned when uploading large (8gb) hart files
This was happening on an automate installed on prem builder. The hart file wrapped a very large software installation that was about 8GB. The hab client reported a 422 status from the server. In one installation (customer POC) this would fail on all 5 attempts and on a local lab installation we saw the 422 once and then success on a subsequent request.

Note that we had to jump through a few hoops to get past various nginx timeouts before we finally got to the 422 in builder-api. We set `proxy_send_timeout`, `proxy_read_timeout`, `proxy_connection_timeout` and `keepalive_timeout` to `600` in builder-api.  Further, we had to set `keepalive_timeout` on automate-load-balancer to 600 as well. Then we had to also set  `proxy_send_timeout` and  `proxy_read_timeout` to 600 on automate-load-balancer but those settings were not exposed in the automate configuration. So we had to manually edit the nginx.conf template in the /hab/pkgs/chef/automate-load-balancer config directory to inject them.

The 422 was coming from https://github.com/habitat-sh/builder/blob/6672b653545014f89927ce743da8bc5d6712de6a/components/builder-api/src/server/resources/pkgs.rs#L1035 where it creates and iterates a reader over the uploaded hart file. The error was:
```
Nov 22 21:12:19 ip-10-0-0-236 hab[713624]: automate-builder-api.default(O): [2021-11-22T21:12:19Z INFO  habitat_builder_api::server::resources::pkgs] Could not read the package at "/hab/svc/automate-builder-api/data/3d68ed95-d6ab-4f5e-aca4-4564e791afb8.tmp": IO(
Nov 22 21:12:19 ip-10-0-0-236 hab[713624]: automate-builder-api.default(O):         Custom {
Nov 22 21:12:19 ip-10-0-0-236 hab[713624]: automate-builder-api.default(O):             kind: Other,
Nov 22 21:12:19 ip-10-0-0-236 hab[713624]: automate-builder-api.default(O):             error: "premature eof",
Nov 22 21:12:19 ip-10-0-0-236 hab[713624]: automate-builder-api.default(O):         },
Nov 22 21:12:19 ip-10-0-0-236 hab[713624]: automate-builder-api.default(O):     )
```
There are a few places where this could possibly occur: opening the file (maybe??), removing the encryption header, iterating the tar items. It is difficult to know without adding more debug logging.

It would also be worth investigating if this only fails in an automate install or also fails in an "ala carte" on-prem builder install.

mwrock:
fixes #1663 

I noticed that the upload stream was hanging after a couple GB. Eventually the connection on the client or the nginx gateway would timeout which would close the stream leaving an incomplete temp file and ultimately resulting in a `premature eof` error while extracting the archive. I also observed that `builder-api` did not begin to read the stream until the client had sent the entire file. That seemed odd since the http 1.1 chunked encoding should allow for proper streaming. 

I eventually tried bypassing nginx and sending the upload strictly to builder-api. This resulted in a successful upload as well as proper streaming behavior.

I later discovered the [`proxy_request_buffering` setting in nginx](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_request_buffering). By default this is on and causes nginx to buffer the entire body before sending it off to builder-api. By turning it off, I got proper streaming and successful uploads. I'm still not clear why the stream was hanging, but this appears to fix the problem.

### DIFF
--- a/components/builder-api-proxy/habitat/config/nginx.conf
+++ b/components/builder-api-proxy/habitat/config/nginx.conf
@@ -117,6 +117,7 @@ http {
     proxy_read_timeout {{cfg.nginx.proxy_read_timeout}};
     proxy_connect_timeout  {{cfg.nginx.proxy_connect_timeout}};
     proxy_http_version 1.1;
+    proxy_request_buffering off;
     proxy_set_header Connection "";
 
     {{~#if cfg.server.listen_tls}}


sajjaphani (approved):
Looks good to me

PR #1685, opened 2021-12-22
spawn log ingester thread in a tokio runtime to fix jobsrv panic

mwrock:
While investigating another unrelated issue, I notices that the jobsrvr would panic when uploading build logs to S3 with:

```
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

The job server would restart and the log still made its way to S3 so this error would go largely unnoticed until we got any decent build load which would cause the job server to crash at the end of every build.

This probably came up some time in the last 6 months during a tokio upgrade.

I built and tested the binary on acceptance and validated that the panic does not occur with this fix.


sajjaphani (approved):
Looks good to me

PR #1689, opened 2022-01-05
Enhance builder info

sajjaphani:
This PR handles to add/update additional information in the following API:

Existing package metadata APIs have been incorporated to return the size of the hart file. Below is an example:
https://bldr.habitat.sh/v1/depot/channels/core/stable/pkgs/7zip/latest?target=x86_64-linux

The APIs returns the below additional property JSON response:
`{
 ....,
  size: 1569,
....
}`

Change the channels API to include the promotion timestamp: **/depot/pkgs/{origin}/{pkg}/{version}/{release}/channels**
The API returns the below JSON response:
`[
  {
    name: 'unstable',
    created_at: '2022-01-05T11:09:34.245636',
    promoted_at: '2022-01-05T11:09:34.245636'
  },
  {
    name: 'foo',
    created_at: '2022-01-05T11:09:37.385983',
    promoted_at: '2022-01-05T11:09:37.385983'
  }
]`

**Note:** UI changes will be handled in a separate PR.


mwrock:
Any reason why you would not add the size to `do_get_package` adding:

```
pkg_json["size"] = json!(req_state(&req).packages.size_of(&ident, target).await);
```
Note that the above won'r compile because `size_of` returns a `Result` but I'm just trying to convey the gist of how it might look.

sajjaphani:
> Any reason why you would not add the size to `do_get_package` adding:
> 
> ```
> pkg_json["size"] = json!(req_state(&req).packages.size_of(&ident, target).await);
> ```
> 
> Note that the above won'r compile because `size_of` returns a `Result` but I'm just trying to convey the gist of how it might look.
> 
> This would eliminate the redundant code in the extra API and also cache the size along with the rest of the metadata in memcache.

We could do that, but it requires us to change the signature of **do_get_package** and all the other functions that depend on this.

We also need to check and invalidate the cache entry for package metadata, where the size attribute is missing.

mwrock:
How would you need to change the signature? It seems like you just need ident and target which `do_get_package` has. At least by the point in the function where it is ready to fetch the metadata it has a target and fully qualified ident.

Regarding cache invalidation, we could restart the memcache services after the release to clear out the cache.

components/builder-api/src/server/services/metrics.rs
```diff
@@ -28,6 +28,7 @@ pub enum Counter {
     SingleUploadRequests,
     MultipartUploadRequests,
     DownloadRequests,
+    SizeRequests,
```
mwrock, line 31: I think we can remove the `SizeRequest` counter now because this will be part of `GetPackage`.

mwrock (changes requested):
This looks good! I just think we need to remove the extra Size counter.

mwrock:
Per our offline conversation, it does look like the on-prem upgrade process involves a supervisor restart: https://github.com/habitat-sh/on-prem-builder/blob/main/on-prem-docs/upgrading.md

sajjaphani:
> This looks good! I just think we need to remove the extra Size counter.

SizeRequest is update on call to size_of in s3.rs (AWS s3 API call), I think it will be good to have this metric.

sajjaphani:
Squashed all the commits and should be good now! @mwrock Can you take a quick look at this?

mwrock (approved):

PR #1690, opened 2022-01-05
add build script and update Makefile

pozsgaic:
Signed-off-by: pozsgaic <pozsgai@progress.com>

Added a build script and modified the Makefile to enable building from Makefile:
--   sudo make build-builder-api

Or directly using the build script
--  sudo ./build.sh builder-api

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ help:
 
 define BUILD
 build-$1: linux ## builds the $1 component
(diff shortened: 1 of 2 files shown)


pozsgaic:
We will make one further update for this effort:

-  Create an environment script that will be called from build.sh and test/run_cargo_test.sh and
    invoke that rather than cutting/pasting environment settings.

mwrock (approved):

PR #1700, opened 2022-01-24
Enhancing Builder info (UI)

### LINKED ISSUE #1624: Enhanced Bldr Package Information
This issue is for Aha -Idea :- https://chef.aha.io/ideas/ideas/APPDL-I-11

When looking at builder it would be very useful to have 2 more bits of information:
 
Date promoted – Sure the application is “stable” but when did that happen? Makes trouble shooting difficult if I am looking through and not knowing when something was promoted.
Size of the HART file - the number of times I have been asked how big the install is…only way to figure it out that I know of is to download it and look in cache.

sajjaphani:
closes #1624

This PR updates the UI for:
1. To Show the HART size
2. To show the promotion date of channels

In addition, it addresses the following fixes:
1. Choose the proper creation and updation date for package channels 
2. Make sure the **Versions** tab is highlighted when expanding the versions nodes.

### DIFF
--- a/components/builder-web/app/shared/date/date.component.ts
+++ b/components/builder-web/app/shared/date/date.component.ts
@@ -34,6 +34,6 @@ export class DateComponent {
   }
 
   get displayDate() {
-    return moment(this.date).fromNow();
+    return moment.utc(this.date).fromNow();
   }
 }
--- a/components/builder-web/app/shared/date/date.component.ts
+++ b/components/builder-web/app/shared/date/date.component.ts
@@ -1,0 +1,39 @@
+// Copyright (c) 2022 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { Component, Input } from '@angular/core';
+
+import * as moment from 'moment';
+
+@Component({
+  selector: 'hab-date',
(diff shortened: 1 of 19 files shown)


components/builder-web/app/shared/date/date.component.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    if (this.date && this.date.trim() !== '') {
+      tip = this.date;
+    }
+
+    return moment(this.date).format('ddd DD-MMM-YYYY, hh:mm A');
+  }
+
+  get displayDate() {
+    return moment(this.date).fromNow();
```
mwrock, line 37: Can we make sure the time comparison is purely GMT based? The date is GMT but Now is the local time zone which gives an inaccurate time.

mwrock (changes requested):
This looks good. I think we just need to fix the time calculation as noted. One other bug I found but I don't think it is related to this front end PR but in the back end is that the file size sometimes comes as `0` from the pkg release endpoint but is populated in the `latest` endpoint. I have an example of this on my on-prem ec2 you can see:

http://ec2-44-234-8-161.us-west-2.compute.amazonaws.com/v1/depot/pkgs/core/7zip/latest?target=x86_64-windows
http://ec2-44-234-8-161.us-west-2.compute.amazonaws.com/v1/depot/pkgs/core/7zip/16.04/20190523094208?range=0

They are both the same release but the first link has the size and the second link does not.

components/builder-web/app/shared/date/date.component.ts
sajjaphani: Moment offers various parsing options. moment(date) will parse the date in the context of user's local time. There is another variant monent.utc(date) to part as a UTC date. I think we can use this.

https://momentjs.com/guides/#/parsing/

mwrock (approved):

mwrock:
I did discover why the size is coming back 0. The release endpoint is not sending a `target` and it defaults to linux. Because this is a Windows package, it cannot find it. I think the fix for this is to pass `pkg.target` to size_of instead of target. I can play with that later if you are not able to get to this.

sajjaphani:
> size_of

Git it. I think we also need to use the pkg.target when calling channels_for_package_ident.

I will take this as a separate PR.

PR #1706, opened 2022-02-08
Fix picking the pkg target for fully qualified ident

sajjaphani:
When we do not supply the target parameter for a package, it will default to X86_64_Linux. However, if the input ident is fully qualified, the builder determines the correct target when finding a package.

Sample route:
<builder-url>/v1/depot/pkgs/neurosis/testapp/0.1.4/20181115124506

After identifying the package, the target property from the newly found package should be used instead of the default.


mwrock (approved):
This looks good. I think you just need a simple formatting change here.

PR #1726, opened 2022-04-22
Worker disconnects

pozsgaic:
Draft


components/builder-jobsrv/src/server/worker_manager.rs
```diff
@@ -865,15 +848,38 @@ impl WorkerMgr {
...
+            | jobsrv::JobState::CancelPending
+            | jobsrv::JobState::CancelProcessing => jobsrv::WorkerState::Busy,
+            jobsrv::JobState::Rejected
+            | jobsrv::JobState::CancelComplete
+            | jobsrv::JobState::Complete
+            | jobsrv::JobState::Failed => jobsrv::WorkerState::Ready,
+        };
+        heartbeat.set_state(worker_state);
+        heartbeat.set_target(job.get_target().to_string());
+        self.process_heartbeat(&heartbeat)?;
```
mwrock, line 881: maybe have `process_job_status` return a `Result<Heartbeat>` and then have the caller call `process_heartbeat`. This feels a little tangled.

components/builder-worker/src/heartbeat.rs
```diff
@@ -12,224 +12,16 @@
...
-    }
-
-    // Set internal state to `PulseState::Pulse` and notify client OK
-    fn resume(&mut self) {
-        debug!("heartbeat resumed");
-        self.state = PulseState::Pulse;
-        self.cli_sock.send("", 0).unwrap();
-    }
-}
+pub fn worker_os() -> proto::Os { proto::Os::Darwin }
```
mwrock, line 27: if we are gutting the heartbeatcli, I don't think we need a separate file for what we are keeping.

components/builder-worker/src/server.rs
```diff
@@ -177,17 +191,21 @@ impl Server {
...
 
     fn set_ready(&mut self) -> Result<()> {
-        self.hb_cli.set_ready()?;
         self.state = State::Ready;
         Ok(())
     }
 
+    // Broadcast to subscribers the HeartbeatMgr health and state
+    fn send_heartbeat(&mut self) -> Result<()> {
+        self.fe_sock.send(&message::encode(&self.heartbeat)?, 0)?;
```
mwrock, line 205: do we need a separate function for this? I think we are just calling it once right?

components/builder-worker/src/server.rs
```diff
@@ -177,17 +191,21 @@ impl Server {
     }
 
     fn set_busy(&mut self) -> Result<()> {
-        self.hb_cli.set_busy()?;
         self.state = State::Busy;
```
mwrock, line 194: I'd just move `self.state = State::Busy` to where we are calling `set_busy` and remove the function.

components/builder-worker/src/server.rs
```diff
@@ -177,17 +191,21 @@ impl Server {
...
         Ok(())
     }
 
     fn set_ready(&mut self) -> Result<()> {
-        self.hb_cli.set_ready()?;
         self.state = State::Ready;
         Ok(())
```
mwrock, line 200: Same comment as above but with `self.state = State::Ready;`

components/builder-worker/src/server.rs
```diff
@@ -145,6 +149,16 @@ impl Server {
...
             }
+
+            // If we do not receive a new job or have a job status from the runner,
+            // then send out a heartbeat.
+            if !runner_msg && !fe_msg {
+                match self.state {
+                    State::Ready => self.heartbeat.set_state(jobsrv::WorkerState::Ready),
+                    State::Busy => self.heartbeat.set_state(jobsrv::WorkerState::Busy),
+                }
+                self.send_heartbeat()?;
```
mwrock, line 160: I'm thinking this entire PR could probably be simplified quite a bit if you left the heartbeating be as is and here just send an empty message to the jobsrv.

mwrock (changes requested):
There are a few comments here but overall I think you could leave all the heartbeating alone and just have the worker send an empty message after each poll. That should be enough to keep the connection alive.

pozsgaic:
The [polling](https://github.com/habitat-sh/builder/blob/main/components/builder-worker/src/server.rs#L107) on the builder-worker does not timeout and blocks until it gets a request from builder-jobsrv or a job status update from the job runner.  As currently implemented it will not send periodic data to builder-jobsrv job port, and we would continue to see this behavior.

mwrock:
right. yes I definitely think you should keep the timeout.

mwrock (changes requested):
I do think we should separate the test into a separate PR. Also I'm wondering if it would be better to base these tests on job status rather than the data in busy_workers. The reason is that busy_workers is sort of a low level implementation. It seems like what we are concerned with here is that if we submit a job, it gets picked up by a worker and if that worker dies, adding a job should result in the job being queued indefinitely. Then after starting the worker back up, we should see the job go into InProgress or completed. So it seems like polling the state of the job via the API would be a more user oriented way of going about this.

components/builder-jobsrv/src/server/worker_manager.rs
```diff
@@ -865,11 +865,24 @@ impl WorkerMgr {
 
     fn process_job_status(&mut self) -> Result<()> {
         self.rq_sock.recv(&mut self.msg, 0)?;
+
+        let worker_name = match self.msg.as_str() {
```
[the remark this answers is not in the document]
pozsgaic: I had to do it a little differently but still better than the match syntax.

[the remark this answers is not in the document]
pozsgaic: I removed the test files and will create a separate PR for the testing

mwrock (approved):
just need to fix clippy

PR #1735, opened 2022-05-24
add HAB_FUNC_TEST end to end overrides for builder-worker

pozsgaic:
Signed-off-by: pozsgaic <pozsgai@progress.com>

### DIFF
--- a/components/builder-worker/src/runner/mod.rs
+++ b/components/builder-worker/src/runner/mod.rs
@@ -315,9 +315,7 @@ impl Runner {
         self.check_cancel(tx).await?;
         let mut section = streamer.start_section(Section::PublishPackage)?;
 
-        if env::var_os("HAB_FUNC_TEST").is_some() {
-            // Skip post process
-        } else {
+        if env::var_os("HAB_FUNC_TEST").is_none() {
             match post_process(&mut archive,
                                &self.workspace,
                                &self.config,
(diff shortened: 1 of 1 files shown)


components/builder-worker/src/runner/mod.rs
```diff
@@ -399,6 +404,23 @@ impl Runner {
...
+                    .await;
+            if res.is_err() {
+                debug!("Failed to fetch origin secret key, err={:?}, path={:?}",
+                       res,
+                       self.workspace.key_path());
+            };
+
+            return res;
+        }
+
```
mwrock, line 423: I think you can avoid all of this repetitive code by creating a `token` variable and populating it with either bobo or the `bldr_token` depending on `HAB_FUNC_TEST`

components/builder-worker/src/runner/mod.rs
```diff
@@ -315,6 +315,11 @@ impl Runner {
         self.check_cancel(tx).await?;
         let mut section = streamer.start_section(Section::PublishPackage)?;
 
+        if env::var_os("HAB_FUNC_TEST").is_some() {
+            section.end()?;
+            return Ok(());
+        }
+
```
mwrock, line 322: rather than repeating these 2 lines, you can wrap the `match` block below in a `if env::var_os("HAB_FUNC_TEST").is_none()`.

mwrock (changes requested):

components/builder-worker/src/runner/mod.rs
pozsgaic: updated

components/builder-worker/src/runner/mod.rs
pozsgaic: updated

mwrock (changes requested):

PR #1736, opened 2022-05-25
Fix activating 'Build Jobs' on package page

sajjaphani:
If we navigate to the 'Build Jobs' or 'Settings' page on the package page, the corresponding tab is not active. This change will fix the tab activation.

This 'routerLinkActiveOptions' was removed for the versions tab to be active even after selecting a particular version from the list. The same is true for Build Jobs.

We will investigate how to handle the child's routes to be active later.

### DIFF
--- a/components/builder-web/app/package/package/package.component.html
+++ b/components/builder-web/app/package/package/package.component.html
@@ -19,6 +19,7 @@
     <a
       mat-tab-link
       routerLink="./"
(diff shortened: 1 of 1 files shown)


mwrock (approved):
Should this be added to the settings link too?

sajjaphani:
Not required on the Settings!

PR #1780, opened 2023-04-13
replaced icons with text

jasonheath:


components/builder-web/app/package/package-sidebar/_package-sidebar.component.scss
```diff
@@ -9,5 +9,10 @@
     hab-icon {
       float: right;
     }
+
+    .wbs { // for wee bit smaller
+      font-size: smaller;
```
mwrock, line 14: just looking around and it seems the standard used in this code base is t ouse precise pixel sizes. There seem to be some helper variables defined for this in _base.scss. For example `$small-font-size`. So it might be worth considering one of those variables over `smaller`.

components/builder-web/app/package/package-sidebar/package-sidebar.component.ts
```diff
@@ -43,7 +43,7 @@ export class PackageSidebarComponent {
 
   get buildButtonLabel() {
     return this.building ? 'Build pending' :
-           this.isNewProject ? 'Build latest versions' : 'Build latest version';
+      this.isNewProject ? 'Build latest versions' : `Build latest ${this.platform.name} version`;
```
mwrock, line 46: do we need this? It does not appear to be replacing any icon but is just additional text.

components/builder-web/app/util.ts
```diff
@@ -118,32 +118,32 @@ export function isEarlyAccess() {
...
+    name: 'macOS (x86_64)',
+    title: 'macOS (x86_64)',
     param: 'mac'
   },
   {
     id: 'aarch64-darwin',
-    name: 'macOS M1',
-    title: 'macOS (aarch64)',
+    name: 'macOS (ARM64)',
+    title: 'macOS (ARM64)',
```
mwrock, line 146: Shouldn't these remain `aarch64`? Just wondering where `ARM64` is coming from.

components/builder-web/app/util.ts
```diff
@@ -118,32 +118,32 @@ export function isEarlyAccess() {
 export const targets = [
   {
     id: 'x86_64-linux',
-    name: 'Linux',
-    title: 'Linux',
+    name: 'Linux (x86_64)',
+    title: 'Linux (x86_64)',
     param: 'linux'
   },
   {
```
mwrock, line 125: We should add an entry for `aarch64-linux`.

package-lock.json
```diff
@@ -0,0 +1,3 @@
+{
```
mwrock, line 1: do we want to add this new file?

components/builder-web/package-lock.json
```diff
@@ -9,98 +9,98 @@
       "resolved": "https://registry.npmjs.org/@angular/animations/-/animations-6.1.10.tgz",
       "integrity": "sha512-dd/lq7kw3uwfHPICan8psu2nthuUpp7PvMLuNIm0XxObZ4oNs0ls6uxKEDPnEkRKoGdiJpvmsyzZZN9ACMPEAA==",
       "requires": {
-        "tslib": "^1.9.0"
```
mwrock, line 12: do we want to be making all these updates?

mwrock (changes requested):
I still need to build and preview but just adding these notes for now.

I did a global search for `hab-platform-icon` and found a few stragglers that we probably want to delete.

mwrock:
I built and loaded this into an on-prem builder. It looks good. My comments in the review still stand. The biggest thing I ran into was the missing aarch64-linux item in the `targets` const of util.ts. My builder had ARM packages and the missing entry was throwing javascript errors and distorting the page.

components/builder-web/app/package/package-sidebar/_package-sidebar.component.scss
jasonheath: I dug back into this and experimented.  I went with smaller because it was relative sizing instead of absolute and while I have limited experience implementing responsive layouts that felt better to me based on what I think I know in a responsive design situation.  I also studied the sass bits you pointed out and looked at things with Firefox's developer tools. Using smaller the font size becomes 13.333px.  I could break towards explicitly specifying 13px but given my experiments and research today I think I'd like to maintain this sizing ratio if something else changed in the future. Unless you feel strongly about it, my inclination is to keep it as is at this point.

components/builder-web/app/util.ts
jasonheath: Added an entry in line with the others that I think should work.  You'll see it when I push again.

package-lock.json
jasonheath: I thought so when I added it but it ought to be pulled and I'll do so.

components/builder-web/package-lock.json
jasonheath: Will be rolled back with the next push.

components/builder-web/app/package/package-sidebar/_package-sidebar.component.scss
mwrock: I'm not incredibly concerned but it seemed worth calling out that absolute sizes are used everywhere else. I'll leave this up to you.

components/builder-web/app/util.ts
mwrock: Is there a reason why this is `aarch64` and the M1 is `ARM64`?

jasonheath: I went with ARM64 in the M1 case because that's I believe that is what Apple uses so I went with that there. Based on previously looking at things like [Rust's platform support docs](https://doc.rust-lang.org/nightly/rustc/platform-support.html) I thought aarch64 was the commonly used term pretty much everywhere else.  Your question made me go dig again and based on [this Raymond Chen blog](https://devblogs.microsoft.com/oldnewthing/20220726-00/?p=106898) I guess Windows actually goes with ARM64 too? I'm certainly not gonna argue with Raymond Chen anyway.  So the idea was to use the nomenclature that I thought a given community used.  If I were going to change something I would probably break this way but I'm really tempted to keep the 'macOS (ARM64)' as I think that's what's common in Apple circles.

```
{
  id: 'x86_64-darwin',
  name: 'macOS Intel',
  title: 'macOS (x86_64)',
  param: 'mac'
},
{
  id: 'aarch64-darwin',
  name: 'macOS M1',
  title: 'macOS (aarch64)',
  param: 'mac_m1'
}
```

mwrock: ok if arm64 is ubiquitous for apple then that's cool.

jasonheath: Yeah, I think it is.  I went and did some more digging online and everywhere the architecture seems to be discussed ARM64 is being used.  However, in thinking about this I do think I'm going to make the mac entries match what's in the above comment which I've edited since the original post.  I think those entries would best reflect what's in commonly used based on just looking around a bit.

jasonheath:
Got rid of the hab-platform-icon references, apologies again for spacing on that feedback after vacation.  I think our outstanding items are those references and a confirmed resolution to the mappings in util.ts.

mwrock (approved):

PR #1791, opened 2023-11-07
fix UI issues

sajjaphani:
There are a few primary UI issues observed, including UI navigation issues and problems with updating the promotion timestamp.

You can easily observe these issues by navigating to an origin page (e.g., `https://bldr.habitat.sh/#/pkgs/core`) and selecting a package. From there, go to the versions tab. In the versions tab, expand the versions and choose one to access the details page. Then, return to the versions page by clicking on the version number at the top of the screen and select a different version. This can be repeated multiple times. You may notice that the package promotion details show 'a few seconds ago,' or it may even redirect you to the home page (`https://bldr.habitat.sh/#/pkgs/core`).

The primary reason for this is the way we handle errors and also the delays in the request to obtain some details and the response received.

A better approach to handling errors is to render error pages or components when encountering such scenarios. Since this requires some UX work, we can address this later.


mwrock (commented):
Any suggestions for how to repro the errors on a new builder setup? I setup a new builder and added some linux packages and promoted to stable. I consistently get the correct promotion timestamps. I can see a bunch of 404s for the different platforms so I know there are errors occurring but I can't reproduce the "few seconds ago" issue in order to validate this fix.

mwrock:
Oh I am able to repro if I promote to custom channels (not stable).

mwrock:
I built your changes to builder-api-proxy and installed/reloaded them in my builder and I am still getting the "a few seconds ago" timestamps.

sajjaphani:
> I built your changes to builder-api-proxy and installed/reloaded them in my builder and I am still getting the "a few seconds ago" timestamps.

Is the issue visible to you following the steps outlined in the description?

mwrock:
It seems to be a little different. When I initially land on a details page, the timestamps are correct, if I go back to the versions and then go to a different version they are still correct. However, if I refresh the page, the timestamps reflect "a few seconds ago". I see this on production builder too.

sajjaphani:
This pull request (PR) also resolves another issue: when there are no Linux packages, the UI fails to display the latest package for other platforms. For instance, if only Windows packages exist for a specific origin/name, the latest page erroneously states, 'There are no available Habitat artifacts (.hart files) for this origin/package.'

mwrock:
There still seems to be something wonky with the non-stable/unstable promotion timestamps. Initially landing on the details page everything is fine, but if I refresh the page, the custom channels reflect "a few seconds ago."

mwrock:
Even with the last commit, I'm still seeing the "a few seconds ago" timestamp on custom channels if I either refresh th details page or navigate directly to a details URL. I have done some digging and  I think I might see why. It looks like [`this._channels`](https://github.com/habitat-sh/builder/blob/main/components/builder-web/app/package/package-detail/package-detail.component.ts#L138C18-L138C32) has the channels of the "latest" package which is different from the detail package and the latest package has only been promoted to stable and unstable.

I'm guessing this somehow has its root in the [constructor](https://github.com/habitat-sh/builder/blob/main/components/builder-web/app/package/package/package.component.ts#L111) of `PackageComponent` when it calls `fetchLatest()` which in turn eventually sets the current package channels of the store to the channels of the latest package in https://github.com/habitat-sh/builder/blob/main/components/builder-web/app/actions/packages.ts#L153.

I'm still not sure how to cleanly fix this until I maybe better understand rxjs.

mwrock:
ok this has been working. I've been digging into to some other weirdness but I am sure it is related to nginx caching where after promoting to a channel, the details page does not include the new channel but the versions page does. It looks like nginx is caching the builder-api get-package response. At any rate I think that is unrelated to this PR.

approving. There were a lot of changes made in this PR that did not seem to result in behavioral changes, I'm just not familiar enough with this codebase to authoritatively scrutinize that so if you think anything should be reverted please go ahead, but it looks like things are behaving.

mwrock (approved):

PR #1797, opened 2023-11-29
CHEF-6081: Added testing version of OneTrust script.

agmathur:
Added the testing version of One Trust script in both development file(components/builder-web/index.html) and in deployment file(components/builder-api-proxy/habitat/config/index.html) along with the required GTM changes.

Have used the same parameter 'cfg.hosted' to load One Trust and GTM tags, if its set, so that these scripts are only run if it is a SAS deployment. 

We will need to host these changes in our acceptance environment ( https://bldr.acceptance.habitat.sh/#/pkgs/core ) and share it with relevant stakeholders for testing. Once the testing is completed, team will share the production version of One Trust script which we will need to update in our code base.

https://chefio.atlassian.net/browse/CHEF-6081

PS: I was not able to test the deployment file changes on local.

### DIFF
--- a/components/builder-api-proxy/habitat/config/index.html
+++ b/components/builder-api-proxy/habitat/config/index.html
@@ -2,8 +2,39 @@
 <html prefix="og: http://ogp.me/ns#">
   <head>
     {{#if cfg.hosted}}
+      <script type="text/javascript">
+        var oneTrustHelper = (function () {
+            function evalGTMScript() {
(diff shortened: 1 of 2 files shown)


sajjaphani (approved):
This looks fine to me. I've verified that the changes have been reflected properly.

Screenshot 2023-11-29 at 6 29 31 PM

PR #1798, opened 2023-12-06
CHEF-6081: Added production version of OneTrust script.

agmathur:
Added the production version of One Trust script in deployment file(components/builder-api-proxy/habitat/config/index.html). Have used the same parameter 'cfg.hosted' to load One Trust and GTM tags, if its set, so that these scripts are only run if it is a SAS deployment.

We will need to host these changes in our acceptance environment ( https://bldr.acceptance.habitat.sh/#/pkgs/core ) and share it with relevant stakeholders for testing. Once the testing is completed, we can do the deployment in the production environment.

https://chefio.atlassian.net/browse/CHEF-6081

PS: I was not able to test the deployment file changes on local.

Screenshot 2023-12-06 at 7 46 17 PM
Screenshot 2023-12-06 at 7 46 09 PM

### DIFF
--- a/components/builder-api-proxy/habitat/config/index.html
+++ b/components/builder-api-proxy/habitat/config/index.html
@@ -28,7 +28,7 @@
         })();
       </script>
       <!-- OneTrust Cookies Consent Notice start for bldr.habitat.sh -->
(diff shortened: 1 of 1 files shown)


sajjaphani (approved):
Looking good!

PR #1827, opened 2024-07-31
`builder-api-proxy` dependencies updated for LTS-2024 channel

agadgil-progress:
Updated following dependencies -

1. NodeJS to version - 18.12.1 (available on LTS-2024)
2. `builder-web` dependencies - `webpack`, `node-sass` etc.
3. Removed not required `gcc`, `make` and `python2` from the `plan.sh` dependency

Now at-least locally builds with `LTS-2024` channel and basic `make` and `make test` works locally. 

Checking mileage with CI.


agadgil-progress:
@rahulgoel1 @sajjaphani : It seems there is no `verify` job for `builder-api-proxy` component. As a result the build is successful. 

If you think it's worthwhile adding the job, then I can add the job for that.

.expeditor/verify.pipeline.yml
```diff
@@ -190,4 +190,23 @@ steps:
         docker:
           privileged: true
           environment:
-            - HAB_ORIGIN=ci
\ No newline at end of file
+            - HAB_ORIGIN=ci
```
mwrock, line 193: Is this needed? It only affects the builder-graph build which should be unrelated to all of this.

.expeditor/verify.pipeline.yml
```diff
@@ -190,4 +190,23 @@ steps:
...
+        docker:
+          privileged: true
+          environment:
+            - HAB_ORIGIN=habitat
+            - HAB_INTERNAL_BLDR_CHANNEL=LTS-2024
+            - HAB_BLDR_CHANNEL=LTS-2024
+            # TODO : This needs to be moved to LTS-2024 as well in future. Right now
+            # This is done because some of our `hab` packages like `hab-studio` are
+            # not available on `LTS-2024` yet.
+            - HAB_FALLBACK_CHANNEL=stable
```
mwrock, line 212: I don't think the fallback channel is going to help here. `stable` is already the default but this variable is only used for a hab build and not installing missing hab components before entering the studio which is the case here.

I think (hope) the solution here is to leave `HAB_BLDR_CHANNEL` alone which will get the studio from stable. Then when it enters the studio and starts the build it will use `HAB_INTERNAL_BLDR_CHANNEL` to set `HAB_BLDR_CHANNEL` inside of the studio.

.expeditor/verify.pipeline.yml
agadgil-progress: This is not exactly a change introduced by this commit, this is shown in diff because of the added newline above it. :-)

.expeditor/verify.pipeline.yml
agadgil-progress: @mwrock : Wondering if it's a better idea to pass `HAB_STUDIO_SECRET_HAB_BLDR_CHANNEL` and `HAB_STUDIO_SECRET_HAB_FALLBACK_CHANNEL` (both set to LTS-2024) `env` variables? This makes sure inside the `studio` accidentally `stable` does not get used at all?

agadgil-progress: I am able to make progress in the build with the two env variables above, still seeing some issues, but don't look related to `hab` or `studio`

mwrock: Oh man I'm sorry. You are absolutely correct. `HAB_STUDIO_SECRET_HAB_BLDR_CHANNEL` was the correct variable. I'm not sure setting `HAB_STUDIO_SECRET_HAB_FALLBACK_CHANNEL` to stable will do anything because I think that is the default anyways.

agadgil-progress: Nope: I meant setting `HAB_STUDIO_SECRET_HAB_FALLBACK_CHANNEL` to `LTS-2024` so that, we know inside studio `stable` is not accidentally used somehow.

mwrock:
@agadgil-progress you might consider adding `export NODE_OPTIONS="--dns-result-order=ipv4first"` to the build and se if that resolves the network problem with fetching from npm.

mwrock (approved):

rahulgoel1:
@agadgil-progress  should we merge this PR ?

agadgil-progress:
> @agadgil-progress should we merge this PR ?

Yes, I had it planned yesterday, there is one small change I have to do because LTS-2024 channel upgraded `node` so will need to quickly test it once before merging otherwise the build on main will fail. Will close this today.

agadgil-progress:
Confirmed, this builds against `LTS-2024` channel and merged.

PR #1828, opened 2024-08-01
updated rust version to 1.79.0

sougata-progress:
Story link : CHEF-14638

### DIFF
--- a/components/builder-api/src/server/error.rs
+++ b/components/builder-api/src/server/error.rs
@@ -12,16 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use actix_web::{self, http::StatusCode, HttpResponse, ResponseError};
+use actix_web::{self,
+                http::StatusCode,
+                HttpResponse,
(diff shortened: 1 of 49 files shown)


components/builder-api/src/server/error.rs
```diff
@@ -235,5 +235,5 @@ impl From<string::FromUtf8Error> for Error {
 }
 
 impl From<actix_web::error::BlockingError> for Error {
-    fn from(err: actix_web::error::BlockingError) -> Error { err.into() }
+    fn from(_: actix_web::error::BlockingError) -> Error { Error::System }
```
mwrock, line 238: I'm not sure why this `from` was implemented differently before, but can we follow the pattern we use for the other error types and create an `Error:BlockingError(actix_web::error::BlockingError)`? This way we don't lose the original error.

mwrock (changes requested):
Looks good. Just one comment.

sougata-progress:
> Looks good. Just one comment.
Changes have been pushed

PR #1833, opened 2024-08-05
Removed connected_plans from backend

sougata-progress:
story : CHEF-12418


sougata-progress:
hi @mwrock i have removed the logic for the functions i could find, which i think are related to connected plans. please let me know if there are any other functions that are related to this ex. create_job, etc, as i couldn't figure out if the other functions are needed or not. 
another thing, do i need to run make lint and fmt on this pr as well?

agadgil-progress (commented):
In addition to the comments mentioned, you may want to look at the following files and possibly update/remove 

1. `test/builder-api/src/origin.js` - Fix 'project' related test cases. 
2. possibly comment out a few tests in `test/builder-api/project.js` or may be comment out this file in `test/builder-api/api.js` with a `//TODO` added 
3. Likely following functions in `app/actions/gitHub.ts` are not required as well - (because they were being used only in the `project.ts`) `*githubInstallations` and `*githubRepositories`. This might be a good time to remove those.

test/builder-api/src/origins.js
```diff
@@ -418,118 +404,82 @@ describe("Related Origin API functions", function () {
...
-        .accept("application/json")
-        .set("Authorization", global.boboBearer)
-        .expect(204)
-        .end(function (err, res) {
-          expect(res.text).to.be.empty;
-          done(err);
-        });
-    });
-
-    it("succeeds", function (done) {
```
mwrock, line 447: you want to keep this test. Otherwise the origin is not deleted and that would explain your test failure in the build.

mwrock (changes requested):

sougata-progress:
hi @mwrock I have made the requested changes

mwrock (approved):

PR #1836, opened 2024-08-21
Fixed unauthorized access token deletion bug

sougata-progress:
story : CHEF-15004

### DIFF
--- a/components/builder-api/src/server/resources/profile.rs
+++ b/components/builder-api/src/server/resources/profile.rs
@@ -7,7 +7,8 @@ use crate::{bldr_core,
                      framework::headers,
                      helpers::req_state,
                      AppState}};
-use actix_web::{body::{BoxBody,to_bytes},
+use actix_web::{body::{to_bytes,
+                       BoxBody},
                 http::{self,
                        StatusCode},
                 web::{self,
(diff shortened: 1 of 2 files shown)


components/builder-api/src/server/resources/profile.rs
```diff
@@ -170,6 +172,61 @@ async fn revoke_access_token(req: HttpRequest,
         Err(err) => return err.into(),
     };
 
+    let get_tokens_req = req.clone();
+
+    let tokens_response = get_access_tokens(get_tokens_req).await;
```
mwrock, line 177: If you look below on line 235, this function already gets the access tokens which are the same retrieved by `get_access_tokens`. So you should be able to remove lines 175 - 203.

components/builder-api/src/server/resources/profile.rs
```diff
@@ -170,6 +172,61 @@ async fn revoke_access_token(req: HttpRequest,
...
+    let access_tokens: Vec<Value> =
+        match access_tokens_json.get("tokens").and_then(|t| t.as_array()) {
+            Some(tokens) => tokens.clone(),
+            None => {
+                let body = Bytes::from_static(b"Tokens not present.");
+                return HttpResponse::with_body(StatusCode::NO_CONTENT, BoxBody::new(body));
+            }
+        };
+
+    let valid_token_id = match access_tokens.first() {
```
mwrock, line 205: As mentioned above, we are already getting the access tokens in this function so just move this below the call to `AccountToken::list`. Rather than just checking the first token, you should iterate over each token.

mwrock (changes requested):
In addition to my comments, we should add a functional test(s) to validate this.

mwrock (approved):

PR #1839, opened 2024-09-03
Fixed OAuth error in github login for builder

sougata-progress:
story : CHEF-15427

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1369,7 +1369,7 @@ dependencies = [
  "serde",
  "serde_derive",
  "serde_json",
- "sha2",
+ "sha2 0.10.6",
  "tempfile",
  "toml",
  "url",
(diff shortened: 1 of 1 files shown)


components/builder-api/src/server/mod.rs
```diff
@@ -211,3 +211,9 @@ pub async fn run(config: Config) -> error::Result<()> {
     };
     Ok(srv.run().await?)
 }
```
mwrock, line 213: I think there is a better way here, but I don't have that solution off the top of my head now. I'm guessing this is what clippy is complaining about. If you revert the `bitflags` bump, that should eliminate the need for this clone.

mwrock (commented):
Can you mention what the actual fix was? Was it connected to sha2?

sougata-progress:
The issue was with the sha2 crate version. 0.9.8 is yanked, and that paired with rust version 1.79 was creating the seg fault issue

mwrock (approved):

PR #1842, opened 2024-09-10
Eula popup added

AadeshNichite:
Eula popup added while sign in and sign up

Related Tasks:
https://chefio.atlassian.net/browse/CHEF-13951

https://github.com/user-attachments/assets/f0ec8a48-a921-4511-aedc-b0d82e11d7b0


sajjaphani:
@AadeshNichite If a user accepts the EULA during either sign-in or sign-up, do we need to show it again for the other option? Can you check with Product?

AadeshNichite:
@sajjaphani I had a discussion with product regarding when to show Popup. We have to show popup at the time of sign in only.

sajjaphani (approved):

PR #1845, opened 2024-09-16
Release SAAS builder built against LTS-2024 dependencies

sougata-progress:
story : https://chefio.atlassian.net/browse/CHEF-15035

### DIFF
--- a/.github/actions/hab-pkg-build-and-upload-linux/action.yaml
+++ b/.github/actions/hab-pkg-build-and-upload-linux/action.yaml
@@ -15,6 +15,8 @@ runs:
       shell: bash
       env:
         BLDR_COMPONENT: ${{ inputs.bldr-component }}
+        HAB_BLDR_CHANNEL: LTS-2024
+        HAB_FALLBACK_CHANNEL: LTS-2024
       run: |
         hab pkg build $BLDR_COMPONENT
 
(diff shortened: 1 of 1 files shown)


mwrock (changes requested):
The key changes here looks like they are directed for the local dev experience via `make`. That makes sense to keep but the key intent here is what our CI builds. [This GH action](https://github.com/habitat-sh/builder/blob/main/.github/actions/hab-pkg-build-and-upload-linux/action.yaml ) controls how components are built on merge. These components need to be build against LTS.

This also doesn't seem to impact the verify pipeline. we want to make sure that pipeline uses LTS.

.github/actions/hab-pkg-build-and-upload-linux/action.yaml
```diff
@@ -24,7 +24,7 @@ runs:
         HAB_AUTH_TOKEN: ${{ inputs.hab-auth-token }}
       run: |
         source results/last_build.env
-        sudo hab pkg install --auth $HAB_AUTH_TOKEN results/$pkg_artifact
+        sudo hab pkg install --auth $HAB_AUTH_TOKEN results/$pkg_artifact --channel LTS-2024
```
mwrock, line 27: I don't think this is going to do anything. Because it is installing from a local hart file, it won't pull from a channel.

mwrock (changes requested):
The `.github\actions\hab-pkg-build-and-upload-linux\action.yaml` will need to set `HAB_BLDR_CHANNEL` and `HAB_FALLBACK_CHANNEL` to `LTS-2024` for the `hab pkg build` step.

mwrock (approved):

PR #1860, opened 2024-12-09
minio migration via hooks

jasonheath:


mwrock:
@jasonheath I didn't get too far with this:
* I was on a on-prem bldr setup via the ./install.sh script.
* I promoted all of the LTS-2024 packages on builder to a new channel `LTS-TEST`. 
* I built builder-minio from this PR in the `habitat` origin and uploaded it to the `LTS-TEST` channel.
* I then ran `sudo ./uninstall.sh` on my builder
* I edited my `bldr.env` file to use the LTS-TEST channel
* I ran `./install.sh`

The install completed but builder-minio was in the `down` state. I ran `journalctl -n 50 -fu hab-sup` and see the following:
Dec 10 21:19:58 ip-172-31-37-92 hab[490708]: builder-minio hook[install]:(HK): CHECKING dependencies
Dec 10 21:19:58 ip-172-31-37-92 hab[490708]: builder-minio hook[install]:(HK): core/aws-cli not installed, installing
Dec 10 21:19:58 ip-172-31-37-92 hab[490708]: builder-minio hook[install]:(HK): sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
Dec 10 21:19:58 ip-172-31-37-92 sudo[490935]: pam_unix(sudo:auth): conversation failed
Dec 10 21:19:58 ip-172-31-37-92 sudo[490935]: pam_unix(sudo:auth): auth could not identify password for [hab]
Dec 10 21:19:58 ip-172-31-37-92 hab[490708]: builder-minio hook[install]:(HK): sudo: a password is required
Dec 10 21:19:58 ip-172-31-37-92 hab[490708]: builder-minio hook[install]:(HK): ERROR: install of core/aws-cli FAILED
Dec 10 21:19:58 ip-172-31-37-92 hab[490708]: builder-minio(HK): Installation failed! 'install' exited with status code 4
Dec 10 21:19:58 ip-172-31-37-92 hab[490708]: hab-sup(MR): Failed to run install hook for habitat/builder-minio, habitat/builder-minio/9779/20241210205901 install hook failed: exited with status 4

The prerequisites check could be avoided as long as the prerequisite packages are in the pkg dependencies.

components/builder-minio/habitat/config_install/minio-migration.sh
```diff
@@ -0,0 +1,116 @@
...
+}
+
+function _curl() {
+  _hab_exec 'core/curl' curl "$@"
+}
+
+function _jq() {
+  _hab_exec 'core/jq-static' jq "$@"
+}
+
```
mwrock, line 20: As long as aws, curl, and jq are runtime deps, you don't need `hab pkg exec`. You can call them directly because they will be on the path.

components/builder-minio/habitat/config_install/minio-migration.sh
```diff
@@ -0,0 +1,116 @@
...
+      sleep .5
+    fi
+  done
+  return 1
+}
+
+function minio_stop() {
+  for ((n = 0; n < 20; n++)); do
+    if pgrep minio >/dev/null; then
+      pkill minio >/dev/null
```
mwrock, line 73: this might be emitting to stderr

components/builder-minio/habitat/config_install/minio-migration.sh
```diff
@@ -0,0 +1,116 @@
...
+      return 0
+    fi
+  done
+}
+
+function config_environment_for_migration() {
+  if [[ ! $1 =~ ^[0-9]+$ ]]; then
+    echo "ERROR: Invalid timestamp"
+  fi
+  export MINIO_ENDPOINT="{{cfg.env.MINIO_ENDPOINT}}"
```
mwrock, line 85: maybe use `https://localhost:{{cfg.bind_address}}` like the other hooks do.

components/builder-minio/habitat/config_install/minio-migration.sh
```diff
@@ -0,0 +1,116 @@
...
+  export AWS_ACCESS_KEY_ID="{{cfg.env.MINIO_ACCESS_KEY}}"
+  export AWS_SECRET_ACCESS_KEY="{{cfg.env.MINIO_SECRET_KEY}}"
+  export s3_url="s3://{{cfg.env.MINIO_BUCKET}}"
+  export MINIO_BUCKET="{{cfg.env.MINIO_BUCKET}}"
```
mwrock, line 89: would the two above be different from `{{cfg.bucket_name}}`?

components/builder-minio/habitat/default.toml
```diff
@@ -35,8 +35,14 @@ bind_port = 9000
 # Setup additional environment variables to pass to minio
 # Doc: https://docs.min.io/docs/minio-server-configuration-guide.html
 [env]
+MINIO_ENDPOINT="http://localhost:9000"
+MINIO_BUCKET="habitat-builder-artifact-store.local"
```
mwrock, line 39: see above. You may not need these two.

components/builder-minio/habitat/hooks/install
```diff
@@ -0,0 +1,100 @@
...
+  minio_old_summary=$(summarize_old_minio_bucket_objects "$timestamp")
+
+  download_bucket_objects
+
+  minio_stop
+
+  MIGRATION_BACKUP_DIRECTORY=$(mktemp -d -t minio-data-backup-"$timestamp"-XXXXXXXXXX)
+  cp -r /hab/svc/builder-minio/data "$MIGRATION_BACKUP_DIRECTORY"
+  GLOBIGNORE=".:.."
+  for x in /hab/svc/builder-minio/data/*; do
```
mwrock, line 58: Inform user here that this might take some time.

components/builder-minio/habitat/hooks/install
```diff
@@ -0,0 +1,100 @@
...
+  upload_bucket_objects
+
+  minio_new_summary=$(summarize_new_minio_bucket_objects "$timestamp")
+
+  if diff <(tail -n 2 "$minio_old_summary") <(tail -n 2 "$minio_new_summary"); then
+    echo "Migration was successful"
+    echo "A copy of the original minio backend has been left in $MIGRATION_BACKUP_DIRECTORY and should be removed manually"
+    rm -rf "$WAYPOINT"
+    echo "Old Minio bucket summary: $minio_old_summary"
+    echo "New Minio bucket summary: $minio_new_summary"
```
mwrock, line 86: What does the summary include? Is it a list of all packages migrated? That could be really long. Maybe save the lists to files and store alongside the backup?

components/builder-minio/habitat/config_install/minio-migration.sh
jasonheath: Removed wrappers and called directly.

components/builder-minio/habitat/config_install/minio-migration.sh
jasonheath: Now redirecting both with &> instead of >

components/builder-minio/habitat/hooks/install
jasonheath: The summaries are what you see in Example 5 [here](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3/ls.html#examples) (I don't see a direct link to the #5, just all of the examples).  Both are written to files in the alongside everything else. Here we're just comparing the actual summaries on the last two lines of those files because of the embedded timestamps.

components/builder-minio/habitat/hooks/install
jasonheath: Added messaging.

components/builder-minio/habitat/config_install/minio-migration.sh
jasonheath: Since this will execute in the context of bldr.env and our users might have changed that there shouldn't we keep it?  I was going to change it and align with the other hooks but that was nagging at me. My counter to myself on that is if people had been changing it then why is hook-helper.sh broken?  I think the answer to that question and what answer your question posed in these three items boils down to the user.toml files created [here](https://github.com/habitat-sh/on-prem-builder/blob/d5701f5c5fbc8948cde2ef00e1207e0e3cf1d2fb/scripts/provision.sh#L70) and [here](https://github.com/habitat-sh/on-prem-builder/blob/d5701f5c5fbc8948cde2ef00e1207e0e3cf1d2fb/scripts/provision.sh#L98).  I think that explains the difference in the bucket name (next comment) and and why we should look to the environment.

components/builder-minio/habitat/config_install/minio-migration.sh
jasonheath: In our hooks we have that habitat-builder-artifact-store.default but in bldr.env.sample and the scripting of on-prem-builder we have habitat-builder-artifact-store.local. I think the user.toml files probably explain why this isn't biting us today.

components/builder-minio/habitat/default.toml
jasonheath: I think in light of the user.toml files and the default values of bldr.env.sample this is probably the way to go.  However, all three of these probably break the same.

components/builder-minio/habitat/config_install/minio-migration.sh
mwrock: I didn't notice that the `bucket_name` in the default.toml is slightly different than what is written to the user.toml. However, the hooks just use what is in `bucket_name` so I'd think you would want to use whatever value is stored in the `bucket_name` configuration setting because that is where the artifacts will be.

components/builder-minio/habitat/config_install/minio-migration.sh
mwrock: The `MINIO_ENDPOINT` in bldr.env is only used if they are using aws S3 or their own external s3. It's not used if they are using our packaged minio.

components/builder-minio/habitat/config_install/minio-migration.sh
jasonheath: I'm not sure about this one in particular. In my on prem builder the bucket name is habitat-builder-artifact-store.local so I'm really feeling like using the info from the bldr.env is the way to go, at least for here.  I lean that way for all of it but when it comes to the bucket name I think we have to use what's coming from bldr.env and while I feel the same about the minio endpoint stuff I would want to review a bit more yet.

mwrock: I guess what I'm trying to say is that if a user has used a bldr.env to set up the user.toml, then `{{cfg.bucket_name}}` will equal `habitat-builder-artifact-store.local` and thus there is no need to introduce a separate setting.

jasonheath: OK, you seem certain and I'm confidence level varies based on which exact part of all this we're talking about.  I'll just try to make everything here way you've asked for.

components/builder-minio/habitat/config_install/minio-migration.sh
jasonheath: Leaving the &> on these but the terminating output we had noticed was really coming from the minio server process so I've gotten that properly redirecting now too.

components/builder-minio/habitat/default.toml
jasonheath: Done.

components/builder-minio/habitat/config_install/minio-migration.sh
jasonheath: Done.

components/builder-minio/habitat/config_install/minio-migration.sh
jasonheath: Done.

mwrock (approved):
Looks good!

PR #1862, opened 2024-12-26
Clean up via shellcheck, shfmt, and markdownlint

jasonheath:
Needed to to trigger the github build action for minio and realized it was an opportunity to take these actions.


mwrock (approved):
Also note you can always trigger the action from the github workflow UI at https://github.com/habitat-sh/builder/actions/workflows/adhoc-hab-pkg-build-upload.yaml and clicking "Run Workflow."

PR #1866, opened 2025-01-29
Remove Obsolete / Dead Code

jasonheath:
This removes obsolete components and dead code.

Code removal was equivalent to the following:
- git rm -r components/builder-graph
- git rm -r components/builder-jobsrv
- git rm -r components/builder-worker
- git rm -r components/builder-api/habitat-dev
- git rm -r components/builder-api-proxy/habitat-dev

After that the top level Cargo.toml was updated to remove the graph, jobsrv, and worker components which dropped a lot of dependencies and restored basic cargo functionality.  Then broad review of the code base started in order to remove or modify items as appropriate for the removal of these items as well as anything else that was identified as obsolete.


jasonheath:
The commits so far all shake out from the initial mandate of dropping the graph, jobsrv, and worker components and the code still compiles via cargo build.  I've intentionally stayed out of the rust code so far but some amount of review is due there to cut out dead code.  The idea be to commit now and see how we far under buildkite at this point.

Beyond changes that shake out from this first push and the subsequent rust changes in the remaining components at least the following items feel as work to be done.

- Both DEVELOPING.md and RELEASE.md deserve review and revision.
- Does anything need to be dropped from the DB?
- Does anything need to be removed from the Builder HTTP API?
- What follow on may be appropriate for on-prem-builder?
- Terraform needs review and updated.

The first four feel like appropriate considerations for this current round of work.

As for on-prem-builder follow on we've deleted a couple of scripts here that I think are mirrored there. Removing those, reviewing for similar, and formally testing that our scripting works as expected may be it. There is one reference to jobsrv in provision.sh that I see but I haven't looked closely at it yet.  This work feels small.

Updating our Terraform scripts is probably best suited as a planned item for a future PI as it may involve another team or be handed off to another team altogether. This work feels large or x-large.

sajjaphani:
@jasonheath The builder-graph component builds a dependency graph, which is then used by the bldr-jobsrv component during the rdeps search and also when deleting a package to ensure there are no rdeps on that package. How do we plan to account for this removal?

PR #1867, opened 2025-02-19
Updated LTS to  base channel support for builder

sougata-progress:
[CHEF-18456](https://progresssoftware.atlassian.net/browse/CHEF-18456)

[CHEF-18456]: https://progresssoftware.atlassian.net/browse/CHEF-18456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/builder-web/app/package/package-sidebar/package-sidebar.component.html
```diff
@@ -23,17 +23,18 @@ <h3>Settings</h3>
...
   </section>
+
   <h3>Latest</h3>
   <div class="sidebar-section" style="margin-left: 12px;">
-    <section class="latest-stable" *ngIf="isLTSChannelExist">
-      <h3 style="margin-bottom: 0px;"> {{ latestLTS  }}</h3>
-      <div *ngIf="currentLts">
-        <a [routerLink]="['./', currentLts.ident.version, currentLts.ident.release]">
-          {{ currentLts.ident.version }}/{{ currentLts.ident.release }}
+    <section class="latest-base" *ngIf="isBaseChannelExist">
```
mwrock, line 29: this should remain `latest-stable`. There wasn't any change to the css class.

components/builder-api-proxy/habitat/default.toml
```diff
@@ -19,7 +19,7 @@ enable_builder_events     = false
 # Enable/Disable the builder events from SaaS
 # The 'enable_builder_events' property also needs to be set to enable SaaS events.
 enable_builder_events_saas = false
-enable_lts = false
+enable_base = true
```
mwrock, line 22: we don't want this enabled by default

components/builder-web/app/util.ts
```diff
@@ -226,4 +230,4 @@ export function labelForJobState(state) {
   }[state.toLowerCase()];
 }
 
-export const latestLTS = 'LTS-2024';
+export const latestBase = 'base';
```
mwrock, line 233: it would be nice if this `base` string was configurable. That way if you wanted to test with another channel or we decided to change the channel name, we don't need to rebuild.

mwrock (changes requested):

components/builder-web/app/util.ts
sougata-progress: @mwrock can you suggest a way how i can do this?

mwrock (approved):

PR #1875, opened 2025-03-05
Jah/rm components and plans obe

jasonheath:


components/builder-core/src/metrics.rs
```diff
@@ -50,17 +49,6 @@ pub trait CounterMetric: Metric {
     }
 }
 
-pub trait GaugeMetric: Metric {
```
jasonheath, line 53: Should GaugeMetric be restored?

components/builder-db/src/models/jobs.rs
```diff
@@ -1,1120 +0,0 @@
-use super::db_id_format;
```
jasonheath, line 1: Similar to above, drop these db items now or wait for a later time and a different PR.

components/builder-db/src/schema/jobs.rs
```diff
@@ -1,102 +0,0 @@
-table! {
```
jasonheath, line 1: Similar to above, drop these db items now or wait for a later time and a different PR.

components/builder-jobsrv/src/migrations/00000000000000_diesel_initial_setup/down.sql
```diff
@@ -1,2 +0,0 @@
-DROP FUNCTION IF EXISTS diesel_manage_updated_at(_tbl regclass);
```
jasonheath, line 1: Similar to above, drop these db items now or wait for a later time and a different PR. In this case this comment applies to all of the db related material in builder-jobsrv component.

test/builder-api/src/misc.js
```diff
@@ -39,7 +39,8 @@ describe('Miscellanenous API', function () {
   });
 
   describe('Retrieving reverse dependencies', function () {
-    it('returns all reverse dependencies for an origin and package name', function (done) {
+    // JAH: ensure this works after rdep refactoring
```
jasonheath, line 42: A general note for reviewers.  Anywhere you see my initials (JAH) that's something I intend to resolve appropriately before finalizing a PR.  Not saying one will never be committed but if you see JAH somewhere its a sign that there's something about that spot I'm still working on somehow.

components/builder-api/doc/api.raml
```diff
@@ -2668,12 +2313,13 @@ types:
...
           description: Unauthorized to perform this action
         '500':
-          description: Internal server erro
+          description: Internal server error
       securedBy:
         - oauth_2_0
     put:
       description: Update a package settings entry
       responses:
+      description: Update a package settings entry
```
jasonheath, line 2322: I'm guessing this slip through on the rebase somehow.  Will need to delete.

components/builder-api/src/server/resources/pkgs.rs
```diff
@@ -1634,82 +1368,3 @@ fn download_response_for_archive(archive: &PackageArchive,
         .append_header((http::header::CACHE_CONTROL, cache_hdr))
         .streaming(rx_body.map(|s| Ok::<_, Infallible>(s)))
 }
-
-async fn has_circular_deps(req: &HttpRequest,
```
jasonheath, line 1638: I've removed this as our discussions but I'm still concerned that we may need it. What will be in place to prevent the creation of circular dependencies in the future? Something in hab autobuild?

test/builder-api/src/misc.js
jasonheath: This has been addressed. Still a good note about my use of my initials.

components/builder-db/src/models/jobs.rs
jasonheath: DB cleanup in terms of dropping unused items needs to happen in some future PR as this one is huge.

components/builder-db/src/functions/jobs.rs
```diff
@@ -1,43 +0,0 @@
-use diesel::sql_types::{Array,
```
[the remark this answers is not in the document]
jasonheath: DB cleanup in terms of dropping unused items needs to happen in some future PR as this one is huge.

components/builder-api/doc/api.raml
jasonheath: Removed.

components/builder-api/src/server/services/github.rs
```diff
@@ -1,22 +1,14 @@
...
-            protocol::jobsrv::{JobGroup,
-                               JobGroupSpec,
-                               JobGroupTrigger},
             server::{authorize::authorize_session,
                      error::{Error,
                              Result},
-                     feat,
-                     framework::{headers,
-                                 middleware::route_message},
+                     framework::headers,
```
mwrock, line 11: you should be able to remove this whole file. We no longer respond to github events.

components/builder-core/src/metrics.rs
mwrock: probably not if we don't use it anymore.

components/builder-protocol/protocols/jobsrv.proto
```diff
@@ -1,108 +1,7 @@
 syntax = "proto2";
 package jobsrv;
-import "net.proto";
```
mwrock, line 3: do we still need this file?

components/builder-protocol/src/jobsrv.rs
```diff
@@ -30,175 +28,6 @@ pub use crate::message::{jobsrv::*,
 
 pub const GITHUB_PUSH_NOTIFY_ID: u64 = 23;
```
mwrock, line 30: do we need this file?

mwrock (changes requested):
This looks mostly good. Just a few comments about some odds and ends.

sajjaphani (commented):
https://github.com/habitat-sh/builder/blob/43c75a042f277498d48fa30ee3ac7080d19d6a04/components/builder-api/src/config.rs#L48

I think the jobsrv config and its struct, JobsrvCfg, are no longer required.

.envrc
sajjaphani, line 1: https://github.com/habitat-sh/builder/blob/43c75a042f277498d48fa30ee3ac7080d19d6a04/components/builder-api/src/server/mod.rs#L78

Do we need this and the RpcClient?

components/builder-core/src/metrics.rs
jasonheath: It did seem to be used anywhere and removal hasn't caused any issues I'm aware of but I was worried that I might be missing something about the stats.  Consider it gone.

jasonheath:
> Can we consider changing the old handlebars syntax `{{~` `~}}` in the nginx.conf file of builder-api-proxy?

Absolutely.  I almost did so but stopped myself because it wasn't hurting anything like an instance of '{{~' in a toml file somewhere but I'm glad to get rid of it.

jasonheath:
@mwrock, @sajjaphani: I think I've addressed everything you've raised. 👍🏻 and 👎🏻 are used to indicate agreement and action based on how your feedback was phrased.  I also rebased the branch to take in the latest changes on main.  I think the one unaddressed item from either of you is my question to @sajjaphani about the `{{~` and `}}` syntax.  Depending on feedback from Matt and Rahul I might defer to that another PR.  I also looked for unused items having pub visibility using some scripts I've written along the way and was able to remove a few last items.  See commit 6b30775a1c6ad90f6b576f822580ed11c082b629.

mwrock (approved):
Just have a merge conflict to clear up. Other than that I'm good,

rahulgoel1:
@jasonheath  can you pls clear up the merge conflicts and merge this PR pls.

jasonheath:
I've rebased and addressed conflicts.  @mwrock has approved.  @sajjaphani and I had a discussion about nginx.conf in Teams a few days ago.  If we take action on nginx.conf it will be as a future PR.  Merging.  CC: @rahulgoel1

PR #1879, opened 2025-03-25
ability to restrict origin creation to authorized users

mwrock:
This allows a builder instance to configure a list of usernames in the builder-api user.toml:

```
[api]
allowed_users_for_origin_create = ['bobo', 'mystique', 'wesker', 'lkennedy']
```

If this list is specified and not empty, a user MUST be in this list in order to create any origin.

### DIFF
--- a/components/builder-api-proxy/habitat/config/habitat.conf.js
+++ b/components/builder-api-proxy/habitat/config/habitat.conf.js
@@ -2,7 +2,6 @@ habitatConfig({
     company_id: "{{cfg.analytics.company_id}}",
     company_name: "{{cfg.analytics.company_name}}",
     cookie_domain: "{{cfg.cookie_domain}}",
-    demo_app_url: "{{cfg.demo_app_url}}",
     docs_url: "{{cfg.docs_url}}",
     enable_builder: {{ cfg.enable_builder }},
     enable_visibility: {{ cfg.enable_visibility }},
--- a/test/builder-api/src/origins.js
+++ b/test/builder-api/src/origins.js
@@ -111,6 +111,20 @@ describe("Origin API", function () {
     });
   });
 
+  describe("Create origin when not allowed", function () {
+    it("requires the user to be allowed", function (done) {
+      request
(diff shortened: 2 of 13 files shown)


sajjaphani (commented):
https://github.com/habitat-sh/builder/blob/2fb1d9d781966ac969436ab7a7ef5b18a8a5035c/components/builder-web/app/origin/origins-page/_origins-page.component.scss#L1

You can add the following style as a child of .origins-page-component:
```css
.error {
  color: map-get($form-colors, error);
}
```

components/builder-api-proxy/habitat/config/habitat.conf.js
```diff
@@ -2,7 +2,6 @@ habitatConfig({
     company_id: "{{cfg.analytics.company_id}}",
     company_name: "{{cfg.analytics.company_name}}",
     cookie_domain: "{{cfg.cookie_domain}}",
-    demo_app_url: "{{cfg.demo_app_url}}",
```
jasonheath, line 5: How does the demo app URL removal fit into restricting origin creation? It does strike me as something that needs to be cleaned up and removed though.

test/builder-api/src/origins.js
```diff
@@ -111,6 +111,20 @@ describe("Origin API", function () {
     });
   });
 
+  describe("Create origin when not allowed", function () {
```
jasonheath, line 114: Should we also have a test that proves the ability for allow listed users to be able to create an origin?

components/builder-api-proxy/habitat/config/habitat.conf.js
mwrock: It was displayed on the "My Origins" page if you had no origins and the mockup removed it. I looked at the url we had and it just redirected to the chef habitat home page so I just stripped it from the code since it was only in a few spots.

jasonheath (approved):

sajjaphani (approved):

PR #1883, opened 2025-04-21
license key

sougata-progress:
[CHEF-20710](https://progresssoftware.atlassian.net/browse/CHEF-20710)

[CHEF-20710]: https://progresssoftware.atlassian.net/browse/CHEF-20710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/builder-web/app/profile/profile/profile.component.html
```diff
@@ -88,6 +88,40 @@ <h3>Personal Access Token</h3>
         </ng-container>
       </section>
       <hr>
+      <section>
```
mwrock, line 91: we only want this on SAAS.

mwrock (changes requested):
Here are a few comments for you to work on. Additionally, consider what happens if someone edits or removes the license in the text box? It seems the data model only accounts for inserts.

components/builder-api/src/server/resources/profile.rs
```diff
@@ -206,6 +215,96 @@ async fn revoke_access_token(req: HttpRequest,
...
+                     -> HttpResponse {
+    let conn = match state.db.get_conn().map_err(Error::DbError) {
+        Ok(conn_ref) => conn_ref,
+        Err(err) => return err.into(),
+    };
+
+    match authorize_session(&req, None, None) {
+        Ok(_session) => {
+            let license_url = format!(
+                "http://licensing-acceptance.chef.co/License/download?licenseId={}&version=2",
```
mwrock, line 231: we should make the licensing url configurable.

components/builder-web/app/profile/profile/profile.component.html
mwrock: I went looking to see how this is done and ther are a couple spots where we display something only on saas. Look at the habitat.conf.jsfile and you will see two settings that are based on `cfg.hosted`.  You could add a thirs here for displaying the license. Or maybe it would be better just to have one setting that is `is_saas_builder` and consolidate all three usages under this.

mwrock (changes requested):

components/builder-web/app/profile/profile/profile.component.html
```diff
@@ -88,6 +88,42 @@ <h3>Personal Access Token</h3>
         </ng-container>
       </section>
       <hr>
+      {{#if cfg.hosted}}
```
sougata-progress, line 91: This flag is not working

mwrock: I just pushed a commit to illustrate what I was getting at. I have not tested it yet.

components/builder-api/habitat/default.toml
```diff
@@ -3,6 +3,7 @@ log_level = "info"
 [api]
 targets = ["x86_64-linux", "x86_64-linux-kernel2", "x86_64-windows"]
 saas_bldr_url = "https://bldr.habitat.sh"
+license_validation_url = "http://licensing-acceptance.chef.co/License/download"
```
sajjaphani, line 6: license_server_url

components/builder-api/src/server/resources/profile.rs
```diff
@@ -206,6 +217,142 @@ async fn revoke_access_token(req: HttpRequest,
...
+                    Ok(resp) => resp,
+                    Err(err) => {
+                        return HttpResponse::InternalServerError().body(format!("License API \
+                                                                                 error: {}",
+                                                                                err))
+                    }
+                };
+
+            if !response.status().is_success() {
+                return HttpResponse::BadRequest().body("Invalid license key or failed to fetch \
```
sajjaphani, line 249: Do not return a generic error, as it can hide important details like license expiration or invalid license. Instead, return the actual response from the server using:

return HttpResponse::build(status).body(body);
or HttpResponse::with_body....

components/builder-web/app/profile/profile/profile.component.html
sougata-progress: tested the changes. LGTM

mwrock (changes requested):

components/builder-db/src/migrations/2025-04-16-111111_add_license_keys_table/up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE SEQUENCE IF NOT EXISTS license_keys_id_seq;
```
mwrock, line 1: You will need a corresponding `down.sql` file that drops this table.

components/builder-api/src/server/resources/profile.rs
```diff
@@ -268,18 +271,18 @@ async fn set_license(req: HttpRequest,
...
+            let expiration_str =
+                json["entitlements"].as_array().and_then(|entitlements| {
+                                                   entitlements.iter().find_map(|ent| {
+                        let end_str = ent.get("period")?.get("end")?.as_str()?;
+                        match NaiveDate::parse_from_str(end_str, "%Y-%m-%d") {
+                            Ok(end_date) if end_date >= today => Some(end_str),
+                            _ => None
+                        }
+                    })
+                                               });
```
mwrock, line 285: We should throw an error message stating that all entitlements are expired rather than stating we were unable to determine the expiration date.

mwrock (changes requested):
I've left a comment about the expiration error. The only other thing here that feels off is the fact that we are storing the expiration as a string. It seems like we should make it a proper date format. Is there a specific reason it needs to be a string?

components/builder-api/src/server/resources/profile.rs
```diff
@@ -273,22 +273,21 @@ async fn set_license(req: HttpRequest,
...
                     })
                                                });
 
-            let expiration_date = match expiration_str {
+            let expiration_date = match expiration_date {
                 Some(date) => date,
                 None => {
-                    return HttpResponse::BadRequest().body("Unable to determine license \
-                                                            expiration date.");
+                    return HttpResponse::BadRequest().body("All entitlements are expired.");
```
sougata-progress, line 290: @mwrock here

components/builder-api/src/server/resources/profile.rs
```diff
@@ -259,7 +259,7 @@ async fn set_license(req: HttpRequest,
                 };
 
             if !status.is_success() {
-                return HttpResponse::build(status).body(body);
+                return HttpResponse::build(status).body("Invalid license key.");
```
mwrock, line 262: Actually the way it was before is better. We should emit the error from the license server and not swallow it. If the user inputs a invalid key, from what I have seen, the api will return "License not found" which seems correct. However there may be some other circumstance where some other kind of error is thrown and we would want to have visibility into that.

mwrock (approved):

PR #1891, opened 2025-05-09
Added license validation for pkg download

sougata-progress:
[CHEF-20135](https://progresssoftware.atlassian.net/browse/CHEF-20135)

[CHEF-20135]: https://progresssoftware.atlassian.net/browse/CHEF-20135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/builder-api/src/config.rs
+++ b/components/builder-api/src/config.rs
@@ -117,8 +117,7 @@ pub struct ApiCfg {
     pub suppress_autobuild_origins: Vec<String>,
     pub allowed_users_for_origin_create: Vec<String>,
     pub license_server_url: String,
-    pub restricted_channels: Vec<String>,
-    pub excluded_channels: Vec<String>,
+    pub unrestricted_channels: Vec<String>,
 }
 
 mod deserialize_into_vec {
@@ -153,8 +152,7 @@ impl Default for ApiCfg {
                  license_server_url: "http://licensing-acceptance.chef.co".to_string(),
                  suppress_autobuild_origins: vec![],
                  allowed_users_for_origin_create: vec![],
-                 restricted_channels: vec![],
-                 excluded_channels: vec![] }
+                 unrestricted_channels: vec![] }
     }
 }
 
--- a/components/builder-api/src/config.rs
+++ b/components/builder-api/src/config.rs
@@ -154,7 +154,7 @@ impl Default for ApiCfg {
                  suppress_autobuild_origins: vec![],
                  allowed_users_for_origin_create: vec![],
                  restricted_channels: vec!["base-2025".to_string()],
-                 excluded_channels: vec!["lts-2024".to_string()] }
+                 excluded_channels: vec!["LTS-2024".to_string()] }
     }
 }
 
--- a/components/builder-api/src/config.rs
+++ b/components/builder-api/src/config.rs
@@ -153,8 +153,8 @@ impl Default for ApiCfg {
                  license_server_url: "http://licensing-acceptance.chef.co".to_string(),
                  suppress_autobuild_origins: vec![],
                  allowed_users_for_origin_create: vec![],
-                 restricted_channels: vec!["base-2025".to_string()],
-                 excluded_channels: vec!["LTS-2024".to_string()] }
+                 restricted_channels: vec![],
+                 excluded_channels: vec![] }
     }
 }
 
--- a/components/builder-api/src/config.rs
+++ b/components/builder-api/src/config.rs
@@ -118,6 +118,8 @@ pub struct ApiCfg {
     pub allowed_users_for_origin_create: Vec<String>,
     pub license_server_url: String,
     pub unrestricted_channels: Vec<String>,
+    pub partially_unrestricted_channels: Vec<String>,
+    pub restricted_if_present: Vec<String>,
 }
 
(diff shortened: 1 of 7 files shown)


components/builder-api/src/server/resources/pkgs.rs
```diff
@@ -420,6 +423,142 @@ async fn download_package(req: HttpRequest,
                        &conn)
     {
         Ok(package) => {
+            let channels = match channels_for_package_ident(&req, &package.ident, target, &conn) {
+                Ok(channels) => channels,
+                Err(_) => {
+                    return HttpResponse::InternalServerError().body("Failed to determine \
```
mwrock, line 429: we will want to log the actual error.

components/builder-api/src/server/resources/pkgs.rs
```diff
@@ -420,6 +423,142 @@ async fn download_package(req: HttpRequest,
...
+                                return HttpResponse::InternalServerError().body("License validation error.");
+                            }
+                        }
+                    }
+                    None => {
+                        return HttpResponse::Unauthorized().body("Authentication required.");
+                    }
+                }
+            }
+
```
mwrock, line 561: we now have two locations where we fetch the license and check the expiration: here and in the license input. Can we consolidate the logic?

components/builder-api/src/config.rs
```diff
@@ -150,7 +152,9 @@ impl Default for ApiCfg {
                  saas_bldr_url: "https://bldr.habitat.sh".to_string(),
                  license_server_url: "http://licensing-acceptance.chef.co".to_string(),
                  suppress_autobuild_origins: vec![],
-                 allowed_users_for_origin_create: vec![] }
+                 allowed_users_for_origin_create: vec![],
+                 restricted_channels: vec!["base-2025".to_string()],
+                 excluded_channels: vec!["LTS-2024".to_string()] }
```
mwrock, line 157: lets default to empty strings here and let SAAS handle this via a user.toml.

components/builder-api/src/server/resources/pkgs.rs
```diff
@@ -45,7 +47,8 @@ use crate::{bldr_core::metrics::CounterMetric,
                                req_state,
                                Pagination,
                                Target},
-                     resources::channels::channels_for_package_ident,
+                     resources::{channels::channels_for_package_ident,
+                                 profile::fetch_license_expiration},
```
mwrock, line 51: we try not to create public functions in the resources to be accessed across other resources. Consider putting these functions in helpers.rs.

components/builder-api/src/server/resources/pkgs.rs
```diff
@@ -420,6 +421,90 @@ async fn download_package(req: HttpRequest,
...
+            let channels = match channels_for_package_ident(&req, &package.ident, target, &conn) {
+                Ok(channels) => channels,
+                Err(err) => {
+                    return HttpResponse::InternalServerError()
+                        .body(format!("Failed to determine package channels: {}", err));
+                }
+            };
+
+            let should_restrict = if let Some(chs) = channels.as_ref() {
+                let chs_lower: Vec<String> = chs.iter().map(|c| c.to_lowercase()).collect();
```
mwrock, line 433: I'm pretty sure channel names are case sensitive so we should not compare with case insensitivity.

components/builder-api/src/server/resources/pkgs.rs
```diff
@@ -420,6 +421,90 @@ async fn download_package(req: HttpRequest,
...
+                         .iter()
+                         .any(|c| chs_lower.contains(&c.to_lowercase()));
+
+                let in_restricted_if_present = state.config
+                                                    .api
+                                                    .restricted_if_present
+                                                    .iter()
+                                                    .any(|c| chs_lower.contains(&c.to_lowercase()));
+
+                !(in_unrestricted_channels
```
mwrock, line 454: can we shift things around so that we can avoid this "double negative" evaluation? I find this makes the logic much more confusing.

components/builder-api/src/server/resources/pkgs.rs
```diff
@@ -420,6 +421,90 @@ async fn download_package(req: HttpRequest,
...
+                  || (in_partially_unrestricted_channels && !in_restricted_if_present))
+            } else {
+                true
+            };
+
+            if should_restrict {
```
mwrock, line 460: We only want to restrict on SAAS builder and not private on-prem builders. If `restricted_if_present` is empty, lets default to everything being unrestricted.

mwrock (changes requested):

components/builder-api/src/server/resources/pkgs.rs
sougata-progress: I updated this to a more readable logic, but clippy is complaining that the logic can be reduced. What should i do ?
https://buildkite.com/chef-oss/habitat-sh-builder-main-verify/builds/2694#0196eb25-3fe6-42bf-9c4d-2750aff62911

mwrock (approved):
looks good. assuming the build goes green, feel free to merge.

PR #1898, opened 2025-05-27
diesel upgrade

sougata-progress:
[CHEF-20571](https://progresssoftware.atlassian.net/browse/CHEF-20571)

[CHEF-20571]: https://progresssoftware.atlassian.net/browse/CHEF-20571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/builder-db/src/models/pagination.rs
+++ b/components/builder-db/src/models/pagination.rs
@@ -38,15 +38,6 @@ impl<T> Paginated<T> {
         let total_pages = (total as f64 / per_page as f64).ceil() as i64;
         Ok((records, total_pages))
     }
-
-    pub fn load_and_count_records<U>(self, conn: &mut PgConnection) -> QueryResult<(Vec<U>, i64)>
-        where for<'a> Self: LoadQuery<'a, PgConnection, (U, i64)>
-    {
-        let results = self.load::<(U, i64)>(conn)?;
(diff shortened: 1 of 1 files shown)


sougata-progress:
> We can't use the `load_and_count_records` anymore?  You can explain later over teams if you want.

load_and_count_records cannot be applied on boxed types. So not using it. It has to be called on a Paginated query.

components/builder-api/src/server/resources/profile.rs
```diff
@@ -222,13 +215,54 @@ async fn set_license(req: HttpRequest,
...
                      -> HttpResponse {
-    let conn = match state.db.get_conn().map_err(Error::DbError) {
+    let mut conn = match state.db.get_conn().map_err(Error::DbError) {
         Ok(conn_ref) => conn_ref,
         Err(err) => return err.into(),
     };
 
     match authorize_session(&req, None, None) {
         Ok(_session) => {
+            let base_url = &state.config.api.license_server_url;
```
mwrock, line 225: I'm not sure I understand what this added block that hits the license server is for. It downloads the license but doesn't seem to do anything and then calls `fetch_license_expiration` which would download the license again.

PR #1913, opened 2025-06-30
Removed delete license key API

sougata-progress:


components/builder-api/src/server/resources/profile.rs
```diff
@@ -253,27 +252,6 @@ async fn set_license(req: HttpRequest,
...
-        Ok(conn_ref) => conn_ref,
-        Err(err) => return err.into(),
-    };
-
-    let account_id = match authorize_session(&req, None, None) {
-        Ok(session) => session.get_id() as i64,
-        Err(err) => return err.into(),
-    };
-
-    match LicenseKey::delete_by_account_id(account_id, &mut conn).map_err(Error::DieselError) {
```
mwrock, line 268: we should also delete `delete_by_account_id` if nothing is using it.

mwrock (approved):

PR #1919, opened 2025-07-16
Fixes verify pipeline and a bit of hygeine

jasonheath:
- Works around a bug in hab 2.0 for .expeditor/scripts/verify/builder-api-functional.sh by
  - now explicitly installing hab-studio, hab-sup, and hab-launcher as a workaround
  - exporting HAB_AUTH_TOKEN as appropriate for chef-oss pipelines so that it's available
- Does a s/core/chef/ in .expeditor/builder_seed.toml
- Removes #! and adds comment to .expeditor/templates/studiorc
  - this cleans up the concatenation of .expeditor/templates/studiorc and helps makes that a bit more explicit
- Removes some comments from there verify piepline that were "sectioning" as well as one that is ~base knowledge for hab team but wasn't mirrored in other pipelines so why have this extra bit.
- makes a white space change in the verify pipeline to eliminate extra indentation in the yaml comb structure.
  - I just had to fix another yaml file's comb structure after dependabot added something and it's service began choking until things where cleaned up.

### DIFF
--- a/.expeditor/builder_seed.toml
+++ b/.expeditor/builder_seed.toml
@@ -2,7 +2,7 @@ format_version = 1
 file_descriptor = "Packages needed to run an instance of Builder"
 
 [[x86_64-linux]]
-channel = "acceptance"
+channel = "stable"
 packages = [
   # Supervisor and prerequisites
   "chef/hab-launcher",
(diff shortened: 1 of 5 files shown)


.expeditor/builder_seed.toml
```diff
@@ -2,12 +2,12 @@ format_version = 1
 file_descriptor = "Packages needed to run an instance of Builder"
 
 [[x86_64-linux]]
-channel = "stable"
+channel = "acceptance"
```
mwrock, line 5: this should remain `stable` since it is for generating the bootstrap bundle and run after a habitat release wherein we would be promoting to stable.

mwrock (approved):
There is one item that needs changing but this otherwise looks good.

PR #1992, opened 2026-05-28
Run token-generator unit tests in verify pipeline

agadgil-progress:
## Summary
- Extends the shared cargo test runner so it can execute Rust packages from `tools/` as well as `components/`.
- Adds `token-generator` to the verify pipeline’s unit-test matrix so its tests run in CI instead of only locally.
- Updates `ai-track-docs/build-test.md` to document the shared CI/local command path.

## Review focus
- `test/run_cargo_test.sh`: confirm the new path selection logic is minimal and won’t affect existing component test runs.
- `.expeditor/verify.pipeline.yml`: confirm the new `token-generator` step matches the existing unit-test pattern.
- `ai-track-docs/build-test.md`: confirm the documented command matches the actual CI entry point.

## Risks
- **Low**: small shell-script routing change could affect existing component test execution if path detection is wrong.
- **Low**: CI runtime increases slightly because of one added unit-test step.
- **Low**: docs could drift if the shared script changes again later.

## Verification steps
```bash
cd /Users/agadgil/Chef/builder
./test/run_cargo_test.sh token-generator
cargo test -p token-generator --quiet
```

Expected result:
- shared script runs successfully from `tools/token-generator`
- direct crate tests remain green

## Rollback
- Revert commit `0cea129577cc8866d3f75d8779ca9fbec63fa423`

### DIFF
--- a/ai-track-docs/build-test.md
+++ b/ai-track-docs/build-test.md
@@ -32,6 +32,8 @@ The repository verify pipeline also includes a dedicated unit-test step for `tok
 ./test/run_cargo_test.sh token-generator
 ```
 
+That CI step reuses the same shared script developers can run locally, which helps keep local and CI behavior aligned.
+
 ## Recommended low-risk module workflow
 
 For small, isolated changes, `tools/token-generator` is a good starting point because it is a standalone CLI with local unit tests.


agadgil-progress:
close

